PR #56, opened 2015-02-04
Update Converter.ts

markvandenbrink:
Running TypeDoc for the following code snippet results in an error in the `extractTypeParameterType` function: 

```typescript
class Example<T> {
    set Something(SomeValue: T) {
    }
}
```

The object node.typeName is `null` for the set accessor. This problem only occurs when the argument of the setter is a generic type variable. Changing `SomeValue: T` to `SomeValue: string` does not result in an error.

This can be fixed by adding an extra condition to the function. After this fix documentation seems to be generated correctly and no further errors occur.


sebastian-lenz:
Thank you very much for checking into this!

PR #69, opened 2015-02-22
fix basic example run script missing ES5 target

markogresak:
Compiler ES5 target was missing, therefore errors occurred when trying to run the script.


sebastian-lenz:
Thanks, I've forgotten the *nix version.

PR #77, opened 2015-03-28
Create property declarations from constructor parameter properties

awerlang:
TypeScript has a shortcut syntax to declare properties, by using access modifiers on constructor parameters.

I was missing this so support for it is added now.

Covers:
- parameter name
- type
- visibility
- @param comment => property description


sebastian-lenz:
Cool! Thanks a lot!

PR #140, opened 2015-09-16
Fixes TypeError when attempting to call copyFrom on an undefined comment.

adamhickey:
Null check to ensure comment exists before calling its copyFrom method. It's not entirely clear to me why the comment is undefined in my specific case.


sebastian-lenz:
Hi Adam, thanks for the contribution! Looks like TypeDoc tries to copy an empty comment from an interface function parameter to its implementation. I've updates the source files of TypeDoc accordingly.

adamhickey:
Thanks. Sorry about checking in the change to a generated file... oops.

sebastian-lenz:
That's cool, any contribution is welcome!

PR #142, opened 2015-09-19
Default export fix

aciccarello:
Fix for sebastian-lenz/typedoc#111

image

Renamed exports are still documented as their local name which might be unexpected.
image

I wasn't sure how to add tests for this change.

### DIFF
--- a/bin/typedoc.js
+++ b/bin/typedoc.js
@@ -3789,7 +3789,13 @@ var td;
             if (!name) {
                 if (!node.symbol)
                     return null;
-                name = node.symbol.name;
+                // Get name (even of default export)
+                if (node.localSymbol) {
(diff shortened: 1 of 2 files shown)


sebastian-lenz:
Okay, lets do it like this. But we should somehow mark default exports.

aciccarello:
@sebastian-lenz I agree. Maybe add a note to the end of the title. Like "Class SingleExportedClass (default)". I'd have to mess around and see how it looks.

sebastian-lenz:
I'm planing to add an export table at the top of the documentations for external modules, seems to be the only reasonably way to show default exports and reexports.

aciccarello:
Ah, I think that's a great idea. If love to see what you're thinking

Issue #98, opened 2015-05-14
Add support for ES6 modules


sebastian-lenz:

robertbaker:
I'm guessing this is my problem?

Using TypeScript 1.5.0-beta from C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\node_modules\typescript\bin

stderr: C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:5079
                    classMember.implementationOf = new td.models.ReferenceType
                                                 ^
TypeError: Cannot set property 'implementationOf' of undefined
    at C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:5079:50
    at Array.forEach (native)
    at ImplementsPlugin.analyzeClass (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:5064:46)
    at C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:5126:35
    at Array.forEach (native)
    at ImplementsPlugin.onResolve (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:5120:49)
    at Converter.EventDispatcher.dispatch (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:102:34)
    at Converter.resolve (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:1996:26)
    at Converter.convert (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:1951:36)
    at Application.convert (C:\Users\Robert\AppData\Roaming\npm\node_modules\typedoc\bin\typedoc.js:418:41)

sebastian-lenz:
Hi Robert! No, this error seems unrelated. Looks like a class does not implement all interface members. I'll add a fix for that.

sebastian-lenz:
Version 0.3.6 is now on NPM containing the fix.

Spittal:
Very interested in this feature.

aciccarello:
This should be working. If there is a specific issue please search for an open issue to upvote or open a new one.

PR #166, opened 2015-12-18
Feature: Add a plugin that allows an ES6 module's name to be set

### LINKED ISSUE #68: Idea for neat external module names
First of all, TypeDoc 0.2.3 works great with TypeScript 1.4! Very happy with that :-)

But, I do have a request. We use a lot of external modules. The names of these modules are currently automatically composed using the relative path of the module surrounded by double quotes. This composed name is then displayed in the generated documentation (title + TOC). The name is also used as file name for the page of the module. In this file name the invalid characters are replaced with an underscore. This results in a bit "ugly" file names for external modules. Module `x` is displayed as `"x"` in the documentation and the file name for the page shows `_x_.html` (see screenshot below). I understand this is done to avoid name conflicts with inline modules with the same name. Seems logical.

May I do the following proposal to streamline the module name and file name:
1. Remove the double quote prefix and suffix out of the reflection name in the `onBeginResolve` function of the class `DynamicModulePlugin`;
2. Split up the name and file name in 2 separate properties (currently `reflection.name` is used for both the display name and the file name). Add a specific property for controlling the file name, so there can be a difference in the display name and the file name;
3. Prefix this file name property for external modules with something like `module_`. The module with name `x` will end up with the file name `module_x.html`.
4. Add a new tag for the external module comment block (for example `@module [name]`) to allow a custom name override instead of the generated name based on the relative path. This allows neat module naming. If the `@module` tag is omitted the path name is used instead (just how it now works, but without the double quote prefix and suffix of course :-)).

Let me know what you think of this idea. I can help you with the implementation if you want/like. I also have another idea for a future release: Automatic inheritance diagram generation (using SVG). [Doxygen](http://www.stack.nl/~dimitri/doxygen/manual/diagrams.html) has such a feature for C++ documentation. But a rather simple implementation should be possible in TypeDoc. Basically the class inheritance is already plotted in the "Hierarchy" block. So all the necessary data sits there... As you can see we embed manually drawn diagrams in the description markdown and it looks great. Automatic inheritance diagrams would be a very nice addition in my opinion!

typedoc-quotes-2

christopherthielen:
Also allows one module to be merged into to another module

Workaround for issue #98
Fixes #68 (parts of)

----

I'm trying to use typedoc for https://github.com/angular-ui/ui-router/tree/feature-1.0 . What I'd like to do is generate docs for `src/*/module.ts`, where each `module.ts` re-exports its children ES6 modules.

Until TypeDoc better understands ES6 import/export, this is a stopgap solution that will allow me to publish UI-Router docs.  

It looks for `@module name` inside a "dynamic module comment" (I think thats the term) and renames the module to the name specified.  If the module specified already exists, it moves all of its children to that module, effectively merging the modules.

in `util/util.ts`:

```
/**
 * This is the utils module. It has foo and bar and stuff.
 * 
 * @module util
 * @preferred
 */
/** necessary second comment? */
export * from "../bar/bar";
export * from "../baz/baz";
```

in `util/foo/bar.ts`:
```
/**
 * @module util
 */
/** */
import {foo} from "../foo";
export bar = "bar";
```

in `util/foo/baz.ts`:
```
/**
 * @module util
 */
/** */
import {bar} from "../bar";
export baz = "baz";
```

This will generate a documented module named `util` with variables `bar` and `baz`.

----

For my project, this generates this:

screen shot 2015-12-18 at 5 52 35 pm

Instead of this:

screen shot 2015-12-18 at 5 53 16 pm

### DIFF
--- a/src/td/converter/plugins/ExternalModuleNamePlugin.ts
+++ b/src/td/converter/plugins/ExternalModuleNamePlugin.ts
@@ -1,0 +1,135 @@
+module td.converter
+{
+    /**
+     * This plugin allows an ES6 module to specify its TypeDoc name.
+     * It also allows multiple ES6 modules to be merged together into a single TypeDoc module.
+     *
+     * @usage
+     * At the top of an ES6 module, add a "dynamic module comment".  Insert "@module typedocModuleName" to
+     * specify that this ES6 module should be merged with module: "typedocModuleName".
+     *
+     * Similar to the [[DynamicModulePlugin]], ensure that there is a comment tag (even blank) for the
+     * first symbol in the file.
+     *
+     * @example
+     * ```
+     *
+     * &#47;**
+     *  * @module newModuleName
+     *  *&#47;
+     * &#47;** for typedoc &#47;
+     * import {foo} from "../foo";
+     * export let bar = "bar";
+     * ```
+     *
+     * Also similar to [[DynamicModulePlugin]], if @preferred is found in a dynamic module comment, the comment
+     * will be used as the module comment, and documentation will be generated from it (note: this plugin does not
+     * attempt to count lengths of merged module comments in order to guess the best one)
+     */
+    export class ModuleAnnotationPlugin extends ConverterPlugin
+    {
+        /** List of module reflections which are models to rename */
+        private moduleRenames: ModuleRename[];
+
+        constructor(converter:Converter) {
+            super(converter);
+            converter.on(Converter.EVENT_BEGIN,              this.onBegin,        this);
+            converter.on(Converter.EVENT_CREATE_DECLARATION, this.onDeclaration,  this);
+            converter.on(Converter.EVENT_RESOLVE_BEGIN,      this.onBeginResolve, this);
+        }
+
+        /**
+         * Triggered when the converter begins converting a project.
+         *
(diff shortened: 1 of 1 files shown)


christopherthielen:
This was good enough to get my project's documentation started, but I realize this may not be exactly what the main project wants/needs.  Perhaps this will spark some discussion.

Spittal:
I would be really into this plugin as it would make my Angular2 application docs not a massive mess in the global scope.

zakhenry:
I've re-implemented this feature (just updating to work with the latest tsdoc) on [my branch](https://github.com/zakhenry/typedoc/tree/ts-1.9) if you want to give it a go, simply update your typedoc version in npm:
```json
"typedoc": "https://github.com/zakhenry/typedoc/releases/download/v0.4.5/typedoc-0.4.5.tgz",
```

Note that I have also updated to TS1.9 in my branch, so I don't think its a good idea to do a PR with both. Let me know if there is interest to separate out into a separate branch for PR

christopherthielen:
I updated my branch (to 1.8) also.  I'll push the updates.

christopherthielen:
I published this as a npm plugin:

```
npm install --save typedoc-plugin-external-module-name
```

PR #245, opened 2016-06-29
Allows comments in tsconfig.json just like tsc

cspotcode:
This uses the same function in Typescript's compiler to parse tsconfig.json, meaning it matches tsc's behavior allowing comments in tsconfig.json.

Do you want to preserve the ability for config files to be written in .js instead of .json?  If so, this PR will need to be updated to use the old method for loading .js.

### DIFF
--- a/src/lib/utils/options/readers/tsconfig.ts
+++ b/src/lib/utils/options/readers/tsconfig.ts
@@ -56,8 +56,7 @@ export class TSConfigReader extends OptionsComponent
             return;
         }
 
-        var result = ts.readConfigFile(fileName, (fileName) => FS.readFileSync(fileName, 'utf8'));
-        var data = result.config;
+        var data = ts.readConfigFile(fileName, (fileName) => ts.sys.readFile(fileName)).config;
         if (typeof data !== "object") {
             event.addError('The tsconfig file %s does not return an object.', fileName);
             return;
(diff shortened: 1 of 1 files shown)


src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -55,7 +56,8 @@ export class TSConfigReader extends OptionsComponent
             return;
         }
 
-        var data = require(fileName);
+        var result = ts.readConfigFile(fileName, (fileName) => FS.readFileSync(fileName, 'utf8'));
```
blakeembrey: I think you can use `ts.sys.readFile` here.

blakeembrey: Just make it `var data = ts.readConfigFile(fileName, ts.sys.readFile).config`.

blakeembrey:
LGTM 👍 Left just one comment.

src/lib/utils/options/readers/tsconfig.ts
cspotcode: Should we use an arrow function so that `readFile` has the proper `this` binding?  `sys` could theoretically be an instance of a class that expects proper `this` values.

cspotcode:
Thanks, I updated the PR to use ts.sys.readFile and also improved the error messages.

blakeembrey:
Thanks! The only possible issue is that `null` could still get through - since `null` is an object and not `undefined`. Happy to merge it though.

cspotcode:
Oh, good catch, I updated the PR.

PR #252, opened 2016-07-08
Fallback to default theme resources

blakeembrey:


felixfbecker:
Does anything prevent this from getting merged?

blakeembrey:
I don't think so, but knowing more people have tested it would help. I'm not familiar with the codebase, and I also couldn't find how this feature could have been broken.

src/lib/output/renderer.ts
```diff
@@ -199,14 +199,7 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
                 }
             }
 
-            var filename = Path.join(path, 'theme.js');
```
Dema: Could we bring this back. As I understand this breaks typedoc-markdown-theme. It has some code in theme.js and now it's being ignored.

blakeembrey: Can you share how that's possible? The code here ran exactly the same regardless of this check.

Issue #256, opened 2016-07-18
Invite new contributors/maintainers?


aciccarello:
@sebastian-lenz has put together a great project here but unfortunately does not seem to have the time to continue working on this. @blakeembrey has helped move typedoc into the TypeStrong organization and done a great job answering questions and keeping the project alive with a few releases. However, he has many other projects that he is involved in and hasn't been able to learn the ins and outs of the typedoc source.

Is anyone able to take a bigger role in this project and would that be welcomed? Right now it feels like there is no one able to take charge of where things are headed?

aciccarello:
@blakeembrey, I would like to be more involved in this project. My OS experience and available time is still limited so I do not think I would be a good maintainer but I would like to contribute. I'll do my best to follow up on and triage GH issues and PRs. I'd also like to contribute to a project code/style review when time allows.

blakeembrey:
Any help on issues right now is certainly appreciated, especially for new feature implementations 😄

emilianosantucci:
Currently `typedoc` is the only valid project to build docs in TypeScript.
Anyway in my opinion, when they started to think on it, them make an error: not consider to build it upon [jsdoc](https://github.com/jsdoc3/jsdoc). 
While two projects differ on target language, it have many common points and features.
Typedoc could take advantage on parsing rules of jsdoc and on community behind it (currently 72 contributors, not bad).

I would like to support this project but now I'm on startup a big project, I hope give you help in near future :)

jiayihu:
I think setting a gitter chat (or even better Slack) could be very helpful for many reasons. I wish there were somebody to ask info when I started using Typedoc's internal API for a script of mine.

I don't have much free time, as probably anyone else, but I'd like to give some contributions to minor issues/changes. I believe it's currently the best project for doc generation in Typescript.

aciccarello:
@jiayihu I'd be up for participating in a Slack channel. There is a big need for documentation updates though. There are features that are not well documented and the existing documentation has not been updated in a long time. I'd prefer we focus on that than having people repeatedly ask the same questions.

I also think that the readme should have some information encouraging people to make PRs since there isn't currently anyone who can devote a ton of time to the project right now.

WasabiFan:
Side-note: If you're going to create somewhere to chat, use Gitter (not Slack). I use both and they're both great, but Slack is designed specifically for private teams while Gitter is the opposite (allowing both private and public). Unless you want to pay for a limited service, don't use Slack.

[...]

PR #266, opened 2016-08-04
Output source files and lines to JSON

tmeasday:
This commit ensures that the json dump includes information about source locations.

### DIFF
--- a/src/lib/models/reflections/container.ts
+++ b/src/lib/models/reflections/container.ts
@@ -67,6 +67,18 @@ export class ContainerReflection extends Reflection
             result['groups'] = groups;
         }
 
+        if (this.sources) {
+          var sources:any[] = [];
+          this.sources.forEach((source) => {
+              sources.push({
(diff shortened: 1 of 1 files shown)


blakeembrey:
FWIW, this coding style is kind of terrible but since it matches the existing style I'll merge it as is. The preferred way to write this would be using `Array#map`.

aciccarello:
@blakeembrey, the entire codebase probably should be reviewed and a tslint configuration added. It would be a lot of work but possibly a good way to someone who wants to take on #256 to get a good handle on the codebase.

blakeembrey:
Absolutely, I agree - there's a lot I'd love to refactor - it's just not in a state where I can actively maintain it. The codebase is takes advantage of too much fanciness for me to focus on how it actually works (decorators, indirect dependencies, etc). I can get a gist, but without rewriting I'm not comfortable maintaining it myself right now 😄

Issue #270, opened 2016-08-05
Domain typedoc.io not reachable, has status pendingDelete


mrbigbunbury:
All links to typedoc.io are leading nowhere, because the domain is on status pendingDelete.

See `whois typedoc.io`:

```
Domain : typedoc.io
Status : pendingDelete
Expiry : 2016-08-04
Delete : 2016-11-03 00:30
```

Edward-Lombe:
Where is the API documented elsewhere I wonder...

aciccarello:
@Edward-Lombe you can build the documentation from the project source if you need. But I do think the website should be brought back.

BurtHarris:
I too discovered the DNS entry for `typedoc.io` is expired.   Similarly, the domain `typestrong.org` looks unattended.   

Custom domains cost money, is there some reason these aren't hosted on https://pages.github.com/ ?

johan-gorter:
Another free option is http://typescript.js.org

BurtHarris:
@sebastian-lenz and @blakeembrey seem to be the key contributors to this project, perhaps we can attract their attention by mentioning their names.

blakeembrey:
Thanks. Someone had reached out on Twitter. I never had control over the websites, so if people are interested we can re-create the site under @TypeStrong.

BurtHarris:
I think it would be a good idea to host it under TypeStrong.   I don’t know what was there before, so I’m curious.

blakeembrey:
I believe this is the existing content: https://github.com/TypeDoc.

aciccarello:
Yeah, I think you are right. Should those get pulled into one repo and updated?

blakeembrey:
Sure @aciccarello! Would you be interested in helping out with that?

blakeembrey:
If someone is interested in setting that up (I can create an repo and add you to it), then I can purchase the domain and we can put things back how it was before.

aciccarello:
@blakeembrey, I would like to but am in the process of moving and won't have any time soon 😢.

ShadowManu:
Hi. I''ve just recently came across the typedoc project but my team may be interested in integrating typedoc in our projects. I could contribute a piece of my daily time next week to fix this and also see how it works for us. I'll confirm tomorrow.

ShadowManu:
Btw, typestrong.org site looks having domain issues too.

d-asensio:
[...]

PR #282, opened 2016-08-30
fixed homepage url in readme

eampiart:
The homepage URL was not alive anymore so this update makes it the same as it is in the repo description, which is alive.


blakeembrey:
Thanks @eampiart. Related to https://github.com/TypeStrong/typedoc/issues/270.

PR #284, opened 2016-08-30
Update README.md

tisawyer:
Added information for Typedoc Webpack plugin.


aciccarello:
LGTM. Only nit pick is that only the NPM links are listed for the other projects. I'd vote for no GitHub links to keep the readme short. You can get to the source from the NPM page.

tisawyer:
Ok sure, I've made that update now

PR #290, opened 2016-09-14
Upgrade to typescript 2.0.3

### LINKED ISSUE #234: TypeScript 2.0 Support
If I try to use `typedoc` with nigthly build or `typescript@next`, it throw this:

`Error: Unknown option `strictNullChecks`.`

DatenMetzgerX:
Upgrade typescript to 2.0.2

Does not add any support for any new features. Just referencing new typescript version.

Fixes #234

### DIFF
--- a/src/lib/converter/nodes/module.ts
+++ b/src/lib/converter/nodes/module.ts
@@ -31,7 +31,7 @@ export class ModuleConverter extends ConverterNodeComponent<ts.ModuleDeclaration
         context.withScope(reflection, () => {
             var opt = context.getCompilerOptions();
             if (parent instanceof ProjectReflection && !context.isDeclaration &&
-                (!opt.module || opt.module == ts.ModuleKind.None)) {
+                (!opt.module || (opt.module as any) === ts.ModuleKind.None)) {
                 reflection.setFlag(ReflectionFlag.Exported);
             }
 
--- a/src/lib/converter/nodes/module.ts
+++ b/src/lib/converter/nodes/module.ts
@@ -31,7 +31,7 @@ export class ModuleConverter extends ConverterNodeComponent<ts.ModuleDeclaration
         context.withScope(reflection, () => {
             var opt = context.getCompilerOptions();
             if (parent instanceof ProjectReflection && !context.isDeclaration &&
-                (!opt.module || (opt.module as any) === ts.ModuleKind.None)) {
+                (!module || module.valueOf() === ts.ModuleKind.None.valueOf())) {
                 reflection.setFlag(ReflectionFlag.Exported);
             }
 
--- a/src/lib/utils/component.ts
+++ b/src/lib/utils/component.ts
@@ -59,7 +59,7 @@ export function Component(options:IComponentOptions):ClassDecorator {
 
         var name = options.name;
         if (name) {
-            proto._componentName = name;
+            proto.componentName = name;
         }
 
         var internal = !!options.internal;
@@ -179,6 +179,10 @@ export abstract class AbstractComponent<O extends IComponentHost> extends EventD
         return this._componentName;
     }
 
+    set componentName(name: string) {
+        this._componentName = name;
+    }
+
 
     /**
      * Return the application / root component instance.
@@ -250,7 +254,7 @@ export abstract class ChildableComponent<O extends IComponentHost, C extends ICo
     }
 
 
-    addComponent<T extends C>(name:string, componentClass:T|IComponentClass<T>):T {
+    addComponent<T extends C & IComponent>(name:string, componentClass:T|IComponentClass<T>):T {
         if (!this._componentChildren) {
             this._componentChildren = {};
         }
--- a/src/lib/utils/component.ts
+++ b/src/lib/utils/component.ts
@@ -130,7 +130,7 @@ export abstract class AbstractComponent<O extends IComponentHost> extends EventD
     /**
      * The name of this component as set by the @Component decorator.
      */
-    private _componentName:string;
+    public componentName:string;
 
     /**
      * A list of options defined by this component.
@@ -173,16 +173,6 @@ export abstract class AbstractComponent<O extends IComponentHost> extends EventD
     getOptionDeclarations():IOptionDeclaration[] {
         return this._componentOptions ? this._componentOptions.slice() : [];
     }
-
-
-    get componentName():string {
-        return this._componentName;
-    }
-
-    set componentName(name: string) {
-        this._componentName = name;
-    }
-
 
     /**
      * Return the application / root component instance.
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "progress": "^1.1.8",
     "shelljs": "^0.7.0",
     "typedoc-default-themes": "^0.4.0",
-    "typescript": "1.8.10"
+    "typescript": "2.0.2"
   },
   "devDependencies": {
     "grunt": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "progress": "^1.1.8",
     "shelljs": "^0.7.0",
     "typedoc-default-themes": "^0.4.0",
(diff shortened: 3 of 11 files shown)


aciccarello:
Does this work with both TS 1.8 and 2.0?

DatenMetzgerX:
No it's not, as this is not supported by the used architecture of typedoc.
I suggest to release a new major version, indicating the breaking change

Am 14.09.2016 20:21 schrieb "Anthony Ciccarello" <notifications@github.com>:

felixfbecker:
I would be nice if this could be merged into master or a v1 branch, targeted as the next major release of TypeDoc and released as prerelease so we can use it now with TS2

satiekue:
I'm using [Path Mapping based Module resolution](https://github.com/Microsoft/TypeScript/issues/5039) feature of TypeScript 2.0 and when I'm using it, it's failing on resolving the paths. (see [example](https://github.com/satiekue/sample))

Note: I'm referencing the [typedoc](https://github.com/TypeStrong/typedoc/pull/290/commits/7b0140ed69ece32a94a440f584be573a0b21506a) version which is in PR right now for 2.0.2 support.

Error: Unknown option `baseUrl`.
Error: Unknown option `paths.*.0`.
Error: Unknown option `paths.*.1`.

DatenMetzgerX:
as mentioned... it does not add any typescript 2.0 functionality...

blakeembrey:
I just tried doing a similar update myself (but to 2.0.3), but couldn't actually get anything to work. There's an incorrect TypeScript `.d.ts` file used, and which I remove it to use the official version there's 100 more compiler errors. If someone has done this update, let me know.

aciccarello:
TypeScript 2.0 should support compiling projects written in 1.8 correct? My understanding is that the default configuration mainly adds stricter checks like control flow so new errors found in documented projects should be minimal.

blakeembrey:
Correct, but there are some incompatibilities. I updated around 30 projects today, around half had errors that weren't caught in 1.8. My issue is ended up being around updating this project though, since there's a custom `.d.ts` file in use that isn't compatible with 2.0, and when I removed it I broke a lot more broke. I'll try again later, was just surprised this PR managed to get away with a lot less changes. I'll give it another go soon though, unless someone else wants to put the legwork in to go for full 2.0 support.

DatenMetzgerX:
The string replaces task messes up the type script definition file for the `performance` module. Had to fix it manually...

jasongrout:
This PR seems to work for me. I do need to compile the bin/typedoc module to use it, though. Can you update this PR to have the compiled typedoc file so we can try using it by installing from github?

DatenMetzgerX:
@jasongrout I cannot include these files in this PR as they are not intended to be checked in and the consequence is a rejected PR ;)

However, I created my own branch that includes the "binaries" as well

```
npm install --save-dev DatenMetzgerX/typedoc#typescript-2-build
```

jasongrout:
Thanks! I just realized that it was the lib/ directory that was missing, and that was in `.gitignore`. I also created a branch, and updated to 2.0.3, manually fixing the issue you saw with the performanc module:  npm install --save-dev jasongrout/typedoc#ts2

lelandmiller:
Hey, took a look at this and it is working for my projects. Has there been any thought on merging this in the past week?

jasongrout:
The string replace grunt task doesn't work quite right, but we fixed it by
hand.

Jason

mohammedzamakhan:
@DatenMetzgerX When can we expect a merge 😄

aciccarello:
I think there are a few questions that need to be answered.

1. Should TypeDoc be expected to work with both TS 1.8 and 2.x?
2. Should TS 2.x support be kept in a separate branch?
3. Should support for features like `readonly` be added before a TS 2.0 version is released?

I think we should have a branch for a TS 2.0 compatible version. Then we could iterate on supporting the new features while figuring out how question 1 should be dealt with.

DatenMetzgerX:
@mohammedzamakhan 
I'm no contributor of this project and therefore not capable to merge the changes myself...

jasongrout:
Personally, I think:

> Should TypeDoc be expected to work with both TS 1.8 and 2.x?

No, not in the same version. It's not designed to do that (even between 1.x releases).

> Should TS 2.x support be kept in a separate branch?

No, I think a working 2.x should be merged and released as a new version of typedoc.

> Should support for features like readonly be added before a TS 2.0 version is released?

No, I don't think that should hold things up (obviously if someone does it, that would be great!).

src/lib/converter/nodes/module.ts
```diff
@@ -31,7 +31,7 @@ export class ModuleConverter extends ConverterNodeComponent<ts.ModuleDeclaration
         context.withScope(reflection, () => {
             var opt = context.getCompilerOptions();
             if (parent instanceof ProjectReflection && !context.isDeclaration &&
-                (!opt.module || opt.module == ts.ModuleKind.None)) {
+                (!opt.module || (opt.module as any) === ts.ModuleKind.None)) {
```
blakeembrey: Please don't cast to `any`.

src/lib/utils/component.ts
```diff
@@ -179,6 +179,10 @@ export abstract class AbstractComponent<O extends IComponentHost> extends EventD
         return this._componentName;
     }
 
+    set componentName(name: string) {
+        this._componentName = name;
+    }
```
blakeembrey: Why not just turn this into a property instead of a getter and setter that does the same thing?

src/lib/converter/nodes/module.ts
DatenMetzgerX: It seems somehow to be needed.... 

`src/lib/converter/nodes/module.ts(34,33): error TS2365: Operator '===' cannot be applied to types 'ModuleKind.CommonJS | ModuleKind.AMD | ModuleKind.UMD | ModuleKind.System | ModuleKind.ES6' and 'ModuleKind.None'.
`

DatenMetzgerX:
I fixed both of your comments (however, enum comparison is still hacky) and upgraded to 2.0.3 as this is the latest stable release.

mohammedzamakhan:
Yayyyyyy!! waiting to use in my project. Let us know when the next version is pushed. 

Thanks @blakeembrey

blakeembrey:
It is `0.5.0`. See https://github.com/TypeStrong/typedoc/releases.

jasonswearingen:
@blakeembrey  um, correct me if I'm wrong, but as per the original push message from @DatenMetzgerX  

> Does not add any support for any new features. Just referencing new typescript version.

This doesn't actually address the #234 issue, which is to generate docs written in TS2 syntax.

blakeembrey:
It should compile 2.0 projects, but I understood that to be that it won't render new features (E.g. that is what constitutes support in `typedoc`). I'd love if someone else wants to help out, I don't have the time myself to maintain this project so I'm relying on community support and PRs. Enough people here seemed to think it was enough from the discussion.

In terms of adding the features, it should be easy enough to get in if someone wants to. Doing things like rendering `readonly`, `null`, `undefined`, etc. Making sure tests for these exist.

mohammedzamakhan:
I am still getting error in my project 

```
Error: Unknown option `baseUrl`.
Error: Unknown option `lib.0`.
Error: Unknown option `lib.1`.
Error: Unknown option `typeRoots.0`.
```

blakeembrey:
Then it looks like there's more work to be done. You can always help out by submitting PRs patching issues you run into 😄

DatenMetzgerX:
Thx... As mentioned... no support for new typescript versions and all these options have been added with typescript 2.0. But you are free to create a pull request that adds support for these new options

BurtHarris:
Oh darn, 0.5.0 doesn't work for my project either.   Like above, complaints about new options in the tsconfig.json file.   While I don't understand much about how this tool works, it seems odd to me to have it complaint about new options it doesn't understand, because it might not *need* to understand about all the options..

PR #293, opened 2016-09-20
Adds --excludePrivate option

### LINKED ISSUE #158: Prevent private members of classes from being documented
I want to exclude documentation about private members so I can make my document simple.
Is there a way to achieve this?

I think it would be nice if there was an option for this purpose, `--excludePrivate` for example.

jkelvie:
Closes #158

### DIFF
--- a/src/lib/converter/nodes/class.ts
+++ b/src/lib/converter/nodes/class.ts
@@ -44,13 +44,10 @@ export class ClassConverter extends ConverterNodeComponent<ts.ClassDeclaration>
         context.withScope(reflection, node.typeParameters, () => {
             if (node.members) {
                 node.members.forEach((member) => {
-                    let privateMember = (member.flags & 16) > 0;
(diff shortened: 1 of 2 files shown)


src/lib/converter/nodes/class.ts
```diff
@@ -36,7 +44,15 @@ export class ClassConverter extends ConverterNodeComponent<ts.ClassDeclaration>
         context.withScope(reflection, node.typeParameters, () => {
             if (node.members) {
                 node.members.forEach((member) => {
-                    this.owner.convertNode(context, member);
+                    let privateMember = (member.flags & 16) > 0;
```
blakeembrey: `const include = member.flags & 16 ? this.excludePrivate : false`?

jkelvie: Hi Blake, made some updates inline with your suggestions for clarity. Take a look...

PR #300, opened 2016-09-30
Exclude constructor private properties

jkelvie:
I noticed the --excludePrivate flag was still adding private constructor properties to the docs.

I think they should be included in the constructor documentation (well, they have to be). But not included as properties on the class.


blakeembrey:
If you can rebase this, I'd be happy to merge. Sorry for the delay.

jkelvie:
Hi Blake, just gave the rebase a try - rebasing is not my expertise, but looks like it came out okay

PR #308, opened 2016-10-16
Ignore new TS 2.0 unsupported options

emilianosantucci:
With this patch, `typedoc` will build docs also in TS 2.x projects without errors.

**Warning**: actually this is only a workaround to allow `typedoc` to get out docs but it need to improve in internal logic to handle new TypeScript features.


christopherthielen:
Note that your PR automatically updates as you push commits to the branch.  You'll probably have to redo the original  PR

emilianosantucci:
@christopherthielen, thanks for suggestion. Looking into linked typedoc plugins, I've also notice the usage of typedoc source code but it was not exported on NPM package so I add it.
If the last commit can be confusing about PR target, I can redo and post a new PR after that.

blakeembrey:
Please don't check in built files, they are ignored for a reason. I'll review once you remove them.

emilianosantucci:
@blakeembrey I have fix it.

package.json
```diff
@@ -56,6 +56,7 @@
     "bin",
     "lib",
     "tasks",
+    "src",
```
blakeembrey: Please remove this.

emilianosantucci: Done

unsafecode:
Do we have an estimation for the release of this fix? Thanks

jonaskello:
+1 Could we please have a release with this merged so we can use Typedoc in TS 2.0?

christopherthielen:
https://github.com/TypeStrong/typedoc/releases/tag/v0.5.1

This was released last week.

jonaskello:
@christopherthielen 👍  Thanks, I didn't notice that :-)

Issue #281, opened 2016-08-26
Cannot find name ... errors when using programmatically with declarations


felixfbecker:
```
c:/Users/felix/git/Projekte/registerex/node_modules/typedoc/lib/converter/types/intrinsic.d.ts(5,105): error TS2305: Module 'ts' has no exported member 'IntrinsicType'.
c:/Users/felix/git/Projekte/registerex/node_modules/typedoc/lib/converter/types/intrinsic.d.ts(6,45): error TS2305: Module 'ts' has no exported member 'IntrinsicType'.
c:/Users/felix/git/Projekte/registerex/node_modules/typedoc/lib/converter/types/intrinsic.d.ts(7,44): error TS2305: Module 'ts' has no exported member 'IntrinsicType'.
c:/Users/felix/git/Projekte/registerex/node_modules/typedoc/lib/output/events.d.ts(20,15): error TS2304: Cannot find name 'HandlebarsTemplateDelegate'.
c:/Users/felix/git/Projekte/registerex/node_modules/typedoc/lib/output/utils/resources/templates.d.ts(4,20): error TS2304: Cannot find name 'HandlebarsTemplateDelegate'.
```
Seems like there are some dependencies typedoc can't live without - should the declarations be published on `typings/registry` instead?

joaogarin:
Having simillar issue, where you able to fix this?

felixfbecker:
Just ignoring it for now. @blakeembrey could you comment on this?

joaogarin:
My issue is opened in https://github.com/TypeStrong/typedoc/issues/268

Also have been ignoring it as the app runs fine, but because I  am using with angular2 there is actually a big list of things I need to include, so its quite a list of warnings;P

blakeembrey:
Most likely both are from global definitions. At least the Handlebars one is. It may be possible to switch to using Typings, or to `@types` when 2.0 is stable.

felixfbecker:
@blakeembrey I tried installing the definitions with `typings`, but typedoc is still using tsd: https://github.com/TypeStrong/typedoc/blob/master/tsd.json. Would be nice if we could add a typings.json until `@types` is ready.

blakeembrey:
I believe there's a PR for `@types`, finishing that up ASAP.

christopherthielen:
These errors might go away if you 
```
npm install --save-dev @types/handlebars
```

blakeembrey:
[...]

PR #310, opened 2016-10-16
Use npm `@types/*` for 3rd party type definitions

christopherthielen:
This PR removes the typings files from `typedoc/src/typings` in favor of using npm `@types/*` packages.

It now (automatically) uses the official `typescript.d.ts` definitions from `node_modules/typescript/lib/typescript.d.ts`.  

TypeDoc uses quite a few internal TypeScript APIs.  Those APIs were exposed in the (old?) `typescript.d.ts` from DefinitelyTyped.  
To accommodate TypeDoc's existing use of internal TypeScript APIs, I created a `ts-internal.ts` file.  The file augments the official type definitions with the internal APIs that we are using.
Obviously it would be best not to use the internal APIs at all.  This first pass can be used to *identify which internal APIs we are using*, so we can work towards replacing those uses with official API.


src/lib/converter/plugins/GitHubPlugin.ts
```diff
@@ -7,6 +7,16 @@ import {BasePath} from "../utils/base-path";
 import {Converter} from "../converter";
 import {Context} from "../context";
 
+// This should be removed when @typings/shelljs typings are updated to the shelljs version being used
+declare module "shelljs" {
+    // `stdout` was added in:
+    // https://github.com/shelljs/shelljs/commit/8a7f7ceec4d3a77a9309d935755675ac368b1eda#diff-c3bfabb5e6987aa21bc75ffd95a162d6
+    // As of 2016-10-16, DefinitelyTyped's defs are for shelljs v0.3.0, but we're on 0.7.0
```
aciccarello: Has anyone created a PR to update this yet?

src/lib/converter/ts-internal.ts
```diff
@@ -0,0 +1,130 @@
+import * as ts from "typescript";
+
+/**
+ * Expose the internal TypeScript APIs that are used by TypeDoc
+ */
```
aciccarello: +1 to identifying the internal APIs so we can move off of them. Way to also link to the internal TS definitions.

src/lib/converter/types/intrinsic.ts
```diff
@@ -4,6 +4,20 @@ import {Type, IntrinsicType} from "../../models/index";
 import {Component, ConverterTypeComponent, ITypeTypeConverter} from "../components";
 import {Context} from "../context";
 
+// Copy typescript's @internal enum set from:
+// https://github.com/Microsoft/TypeScript/blob/v2.0.5/src/compiler/types.ts#L2297-L2298
```
aciccarello: Any reason you defined this here instead of `ts-internal.ts`?

tsconfig.json
```diff
@@ -13,32 +13,19 @@
         "experimentalDecorators": true
     },
     "exclude": [
-        "./node_modules/typescript/lib/typescript.d.ts",
-        "./typescript/lib/typescript.d.ts"
     ],
     "filesGlob": [
```
aciccarello: Should `files`/`filesGlob` be migrated to [`include`](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#glob-support-in-tsconfigjson)?

src/lib/converter/plugins/GitHubPlugin.ts
christopherthielen: I did not find any updates or PRs for shelljs

src/lib/converter/types/intrinsic.ts
christopherthielen: The code in `ts-internal.ts` exposes the existing internal/hidden APIs by augmenting the "typescript" module and exported interfaces.

In this case, the `TypeFlags` enum itself is exported from `ts` namespace.  However, the `Intrinsic` enum value is marked as `@hidden`.  I could not find a way to augment the  existing `TypeFlags` enum (which is exposed as a public API) with a new enum value (which is @internal and hidden).   Since I couldn't cleanly expose something from the `typescript` module using an augmentation, I decided to move it outside that file.  

The other option, I suppose, is to export it from the `ts-internal.ts` file as a separate symbol (not inside the `typescript` namespace).  I don't really have a strong opinion on this.

aciccarello: That makes sense. Unless it is needed in multiple places, defining it here is good with me.

tsconfig.json
blakeembrey: Kill `filesGlob`, nothing uses it anyway.

src/lib/converter/types/intrinsic.ts
blakeembrey: What about using `any` to access this? My only fear is forgetting to update it if anything changes in the future.

christopherthielen: I'm ok with the `any` approach.

src/lib/converter/ts-internal.ts
blakeembrey: Is there any concern about the out-datedness that could occur here? Is there a way to build it with the internal APIs exposed instead and use that?

blakeembrey:
Thanks a ton! It's really awesome. My only concern is with the possible out-datedness that can occur when we pull the TypeScript definition in manually, but it's an amazing start. Thanks for the PR 😄

aciccarello:
> My only concern is with the possible out-datedness that can occur when we pull the TypeScript definition in manually, but it's an amazing start.

@christopherthielen Did you only bring in the definitions that were used? It'd be great to try to reduce this list with future PRs (anyone interested :wink:).

christopherthielen:
@aciccarello yes, the only internal APIs I added typings for are the APIs that TypeDoc is currently using.

src/lib/converter/ts-internal.ts
christopherthielen: > any concern about the out-datedness that could occur

That is certainly a concern.  My hope is that these internal usages can be replaced by other APIs.

christopherthielen:
> My only concern is with the possible out-datedness that can occur when we pull the TypeScript definition in manually

When I created this PR, I thought the `typescript.d.ts` that was committed came from DefinitelyTyped.  However, I not see that it is recent (2.0.3) and *does include the private APIs*.  Maybe @DatenMetzgerX can explain where the definitions came from?

I did a little experimenting and found out that it's trivial to generate the full `typescript.d.ts` including the `@internal` APIs:

```
git clone https://github.com/Microsoft/TypeScript.git
cd TypeScript
npm i
perl -pi.bak -e 's/stripInternal: true/stripInternal: false/g' Jakefile.js
npm run build:compiler
cp built/local/typescript.d.ts ../typedoc/typescript.d.ts
```

I'll see if I can automate the generation of a subset of that file, including only the internal apis being used.

Potentially we could just use the full `typescript.d.ts` with ALL the internal APIs exposed, but I'd much prefer to limit the visibility to those APIs.

blakeembrey:
Understandable. FWIW, I think `ntypescript` exposes all internal APIs. You could look at how that is generated. /cc @basarat

DatenMetzgerX:
@christopherthielen

It's generated as you have described. There is an `UPDATING.md` that describes the updating process in detail

christopherthielen:
@DatenMetzgerX oh, that's great to know and understand, thanks!   I completely missed those instructions.

I guess the question related to this PR is then: which would ya'll prefer:
     
- 1) continue exposing the entire internal typescript API
- 2) only expose the internal APIs actually used

Option 1 is the easiest, and is what the project currently uses.  We could move to `@types/` for only the third party libs and use `typescript.d.ts` with all internal apis exposed.

Option 2 could be automated and might be a step towards a goal of eliminating (or reducing) use of internal APIs altogether.  I have this process partially automated.  I suppose I'll finish automating it and add it to the PR, then ya'll can take a look and make a decision.

felixfbecker:
The only question I have: will this cause "duplicate identifier" errors?

christopherthielen:
@felixfbecker this is only for *building* typedoc, so it shouldn't affect your project's doc generation.

felixfbecker:
@christopherthielen I don't think so, when using typings programmatically in TypeScript it will depend on the typings.

blakeembrey:
@felixfbecker That's a real PITA. I remember having this conversation with the TypeScript team before because we can fix this on our end by providing a custom `typeRoots`, but I don't think `typeRoots` is part of the compilation output which means our configuration means nothing when we publish. /cc @mhegazy @DanielRosenwasser what's the correct approach we should take here and does/will the TypeScript definition output be updated to include the `typeRoot` used with the references?

christopherthielen:
@felixfbecker @blakeembrey  I'm not sure I understand what the problem is.  Can one of you clarify?  

Is the problem related to using `@types/` in general?  
Is it about using the internal TypeScript APIs? 
Or about exposing those internal APIs in `ts-internal.ts` as part of the TypeDoc build?

blakeembrey:
@christopherthielen Shouldn't affect the existing discussion, but in case you're interested it's related to https://github.com/TypeStrong/typedoc/issues/281. Basically, we should move to a system where we can publish the typings included so people can use it programmatically. However, if we `declare module` it'll either cause conflicts with the official TypeScript definition in a consumers project, I believe, or not be used. Either way, probably not what we want. The workaround for this in TS 2.0 is to write external module definitions (a la `node_modules/@types`). However, we can't do this because it's confusing and not supported - so a workaround is `typeRoots` in 2.0. However, this feature won't work either because the `.d.ts` file doesn't emit the `typeRoots` for dependencies so even though we'd use it, and publish it, anyone consuming the project will fail anyway.

For now, don't worry about it, it's just a related conversation that was somewhat relevant and it's a good example for a point I logged a while back in May (?) for TS 2.0 module support.

christopherthielen:
@blakeembrey based on this information, I think we should do the following:

1) Stop referencing internal API functions from the "typescript" module.

Instead of adding them back into the "typescript" module (like `ts-internal` currently does) we should simply export them from `ts-internal.ts` and import from there where used in the TypeDoc code.

```js
import { getJsDocComments } from "../ts-internal";
```

2) Augmenting an exported TypeScript interface should be fine.  Duplicated interface merges are handled nicely by TypeScript. 

For example, this works:
```js
declare module "typescript" {
  interface Symbol {
    id?: number;
    parent?: ts.Symbol;
  }
}

declare module "typescript" {
  interface Symbol {
    parent?: ts.Symbol;
  }
}
```

3) Re-declare and export the (completely hidden) @internal enum and interfaces from `ts-internal`

```js
export interface CommandLineOption {
    name: string;
    type: string;
    shortName: string;
    description: DiagnosticsEnumValue;
    paramType: DiagnosticsEnumValue;
  }
```

Of course, we'll have to maintain them ourselves, but it seems like this approach would address both issues.

christopherthielen:
For what it's worth, I'm also affected by #281.  My kludge currently looks like so: https://github.com/christopherthielen/typedoc-plugin-internal-external/blob/master/tsconfig.json#L12

felixfbecker:
@christopherthielen What about the handlebars typings dependency? Currently it gives a "cannot find module..." when using typedoc programmatically. I assume with this PR that will go away, but what if I already have typings for handlebars installed in my project, e.g. some file that does `declare module 'handlebars'`? That was my question: would this PR then cause a "duplicate identifier ..." error?

christopherthielen:
@felixfbecker because the typings in this PR are listed as `devDependencies`, I don't think a project which depends on `typedoc` would be affected.

blakeembrey:
@christopherthielen That's a good point. Given that, can we make them dependencies? It'll be much more valuable if someone can actually use this module from NPM without having to manually install the dependent types.

mhegazy:
> @felixfbecker That's a real PITA. I remember having this conversation with the TypeScript team before because we can fix this on our end by providing a custom typeRoots, but I don't think typeRoots is part of the compilation output which means our configuration means nothing when we publish. /cc @mhegazy @DanielRosenwasser what's the correct approach we should take here and does/will the TypeScript definition output be updated to include the typeRoot used with the references?

I would say the core here is not custom typeRoots, but the use of private APIs. These APIs are marked private as there are no support promises here. so they could be broken at will or removed all together.

As for typeRoots; this is mainly for users who want to override declarations of a certain package, but I do not think it should be used to publish a dependency. republishing declarations of other packages in general have proved painful for everybody. so my recommendation is to get the original package author to include them (in this case Typescript :)), or have a new package that "expands" on the original package, e.g. publish `blah-internal` that is not `blah` and then use that instead. 

I left a few comments on the PR. Feel free to ignore them, i just found it easier to leave them on the code rather than listing them all on one comment.

felixfbecker:
@mhegazy What do you think about the fact that this module depends on some packages that do _not_ have included typings, like handlebars? The idea is to make `@types/handlebars` a dependency so that a user can use typedoc programmatically without getting "cannot find module..." errors.

christopherthielen:
@mhegazy thanks for the feedback on the internal API usage!

src/lib/converter/ts-internal.ts
```diff
@@ -0,0 +1,130 @@
...
+    id?: number;
+    // https://github.com/Microsoft/TypeScript/blob/v2.0.5/src/compiler/types.ts#L2168
+    parent?: ts.Symbol;
+  }
+
+  interface Node {
+    // https://github.com/Microsoft/TypeScript/blob/v2.0.5/src/compiler/types.ts#L469
+    symbol?: ts.Symbol;
+    // https://github.com/Microsoft/TypeScript/blob/v2.0.5/src/compiler/types.ts#L472
+    localSymbol?: ts.Symbol;
```
aciccarello: I think I used this to get the local name of a default export

christopherthielen:
I'll make another attempt at this PR in the next few days taking into account all this new information.

mhegazy:
> @mhegazy What do you think about the fact that this module depends on some packages that do not have included typings, like handlebars? The idea is to make @types/handlebars a dependency so that a user can use typedoc programmatically without getting "cannot find module..." errors.

Let's say typedoc has a dependency with `@types/handlebars` (after this PR), now if you depend on typedoc and you have another declaratio for `handlebars` (say in `typeings/handlebars/index.d.ts`) you will get a duplicate declaration error. 

The issue here is really how `handelbars` is authored, more on this later. but let's say you can not change `@types\handelbars`. so the recommendation to override it is to add you `typings` folder in the `typeRoots`. this way you are using your local version of the definition. e.g.:

```json
"compilerOptions": {
    "typeRoots": [ "./typings", "./node_modules/@types" ]
}
```
this way the compiler will look at `./typings/handelbars/index.d.ts` before looking at `/node_modules/@types/handelbars/index.d.t.s`. 

But the main problem here is https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/handlebars/index.d.ts, it should be authored as:

```ts
...
export = Handlebars;
```

instead of:

```ts
...
declare module "handlebars" {
    export = Handlebars;
}
```

package.json
```diff
@@ -42,6 +42,13 @@
     "typescript": "2.0.3"
   },
   "devDependencies": {
+    "@types/fs-extra": "0.0.33",
+    "@types/handlebars": "^4.0.31",
+    "@types/highlight.js": "^9.1.8",
+    "@types/lodash": "^4.14.37",
+    "@types/marked": "0.0.28",
+    "@types/minimatch": "^2.0.29",
+    "@types/shelljs": "^0.3.32",
```
blakeembrey: If we can move all these to `dependencies`, I'd be happy to merge today and leave the rest of the awesome comments from @mhegazy for a future PR. I'll open an issue for them too, once this PR lands.

christopherthielen: done. 

I'm not 100% sure this is the right thing to do though.  I think we only need to add as `dependencies` those @types/* packages which are exposed via a public API in the typedoc.d.ts files.

felixfbecker: @christopherthielen No, user's code won't compile if not all modules referenced in the declaration files are available.

christopherthielen: @felixfbecker I guess that's my point, though.  Only those typings that are referenced in the declaration files need to be in `dependencies`.

blakeembrey: That's understandable. The issue with it is that you might accidentally expose one or the other at some point and forget to move them around. I think putting them in the same place as the real dependency is the best approach to avoid unfortunate side-effects.

blakeembrey:
@mhegazy Thanks for all your comments.

> I would say the core here is not custom typeRoots, but the use of private APIs. These APIs are marked private as there are no support promises here. so they could be broken at will or removed all together.

I agree, I was trying to allude to the larger problem as it was part of a discussion we'd had in the past. As a module author, if there's an incorrect definition, I'd like to override `typeRoots` for development and probably publish. I don't want to have to wait days or months just to release a working version of software. This is unrelated to this module, but the bigger issue of `typeRoots` as a temporary solution. For reference, I've already run into this situation in production but because I use Typings and not TypeScript `@types` yet I had no trouble working around it (Typings can install from GitHub URLs, I just installed my fork and the PR is still waiting).

Issue #311, opened 2016-10-17
Support for compilerOptions.types


jupl:
Given this `tsconfig.json`:

``` json
{
  "compilerOptions": {
    "jsx": "react",
    "moduleResolution": "node",
    "noFallthroughCasesInSwitch": true,
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "sourceMap": true,
    "strictNullChecks": true,
    "target": "es2015",
    "types": [
      "jest",
      "node"
    ]
  }
}
```

TypeDoc fails to run with the following error:

```
Error: Unknown option `types.0`.
Error: Unknown option `types.1`.
```

aciccarello:
This is a remaining issue from #234. A PR for this would be great. Also related is #305 (supporting typeroots)

Radagaisus:
A quick patch is to add `typeRoots` and `types` to the [ignored compiler options](https://github.com/TypeStrong/typedoc/blob/22ccb660ae6eb5b3e136dc277ecc1d62fb04d016/src/lib/utils/options/sources/typescript.ts#L16-L20). Though, for better support of future TypeScript versions, and an easier upgrade path, it’s best to switch to a whitelist.

aciccarello:
Does anyone know if this was fixed in #359?

jupl:
@aciccarello The specific error does not show as of the latest version in NPM but errors are being thrown when trying to compile docs such that it appears that type `types` option is being completely ignored by Typedoc.

jiayihu:
I can confirm that `types` option is ignored and I can't compile docs

Gerrit0:
This appears to have been fixed with #449

Issue #305, opened 2016-10-11
support for typeRoots


m3l7:
typedoc crashes if typeRoot option is set in tsconfig.json. Is there a way to avoid this?
here is my tsconfig.json:

```
  {
      "compilerOptions": {
          "target": "es5",
          "module": "commonjs",
          "sourceMap": true,
          "outDir": "src",
          "typeRoots": ["./typings/modules", "./typings/globals"]
      },
      "include":[
          "src/**/*"
      ]
  }
```

aciccarello:
Related to #234. Someone is working on basic TS 2.0 support. You can try the version in #290 and see if that fixes this issue.

m3l7:
I've updated from master and this issue seems to be solved.

PR #312, opened 2016-10-18
Ignore --types and --typeRoots TypeScript compiler configuration options

Radagaisus:
Added `types` and `typeRoots` to the list of strings ignored by `TypeScriptSource`. This means TypeDoc will ignore these compiler settings. This is a quick patch for #305 and #311.


Radagaisus:
Same slash related to #308.

blakeembrey:
Do you mind if I wait for #308 to be tidied up and use that instead?

PR #314, opened 2016-10-20
Use private TypeScript APIs from internal module

christopherthielen:
Do not expose the private TypeScript APIs by augmenting the "typescript" module.  Modifying the declarations of the "typescript" module causes problems when attempting to use TypeDoc programmatically (for building plugins, etc) because the 

No longer augment the "typescript" module's typings to expose private APIs being used by TypeDoc (like this):
```js
// no longer do this
declare module "typescript" {
  function somePrivateTypeScriptApi();
}
```

Instead, create a ES6 module in TypeDoc which has passthroughs to the private TypeDoc API.
```js
import * as ts from "typescript";
const tsany = ts as any;

export function somePrivateTypeScriptApi() {
  return tsany.somePrivateTypeScriptApi.apply(this, arguments);
}
```

Elsewhere, import the passthrough APIs from the internal ES6 module.

```js
import * as _ts from "../ts-internal";

var result = _ts.somePrivateTypeScriptApi();
```


aciccarello:
I just want to note (if I remember correctly) that this PR was made in response to comments in #310

Issue #228, opened 2016-06-07
TypeError: Cannot read property 'flags' of undefined


PascalSenn:
I got this error during the execution of typedoc:

```

 return !!(type.flags & 80896);
                      ^
TypeError: Cannot read property 'flags' of undefined
    at ReferenceConverter.supportsNode ([....]node_modules\typedoc\lib\converter\types\reference.js:26:23)
    at Converter.convertType ([....]node_modules\typedoc\lib\converter\converter.js:127:31)
    at [....]node_modules\typedoc\lib\converter\nodes\class.js:47:63
    at Context.withScope ([....]node_modules\typedoc\lib\converter\context.js:99:9)
    at ClassConverter.convert ([....]node_modules\typedoc\lib\converter\nodes\class.js:35:17)
    at Converter.convertNode ([....]node_modules\typedoc\lib\converter\converter.js:117:53)
    at [....]node_modules\typedoc\lib\converter\nodes\block.js:71:33
    at Array.forEach (native)
    at BlockConverter.convertStatements ([....]node_modules\typedoc\lib\converter\nodes\block.js:69:29)
    at [....]node_modules\typedoc\lib\converter\nodes\block.js:55:27

```

Does someone has an idea what the problem could be?

Typedoc v. 0.4.1
Tpyescript v.1.18.10

ajayambre:
Facing same issue when using gulp-type

lholznagel:
Please take a look at #223

kylenicola:
I ran into this issue when I wasn't compiling something outside of the directory.  My tsconfig had

files ['all.ts' (from ./src), '../typings/index.d.ts']

but I was only running

typedoc --out ./doc ./src

The solution was to run

typedoc --out ./doc ./src ./typings/index.d.ts

0815fox:
I saw the same thing once. The best you can do is, to use a tsconfig file for compiling and for typedoc. Seems, that typedoc invokes something that fails without catching the error. When this error occured, i ran tsc and it failed, too. Best is, to get compilation pass before generating typedoc.

samvv:
I'm having the same issue when invoking TypeDoc programmatically. Anyone knows how to fix this? `tsconfig.json` is in my case no option because I need to dynamically render the documentation.

gms1:
I saw the same issue when typescript@next was installed; using typescript 1.8.10,  typedoc complained about 'Cannot find name/module/...'.
In both cases the project transpiled successfully by tsc!
[...]

PR #333, opened 2016-11-13
Fix for Issue #228

fencim:
SupportsNode method at at references.ts is causing error because getTypeAtLocation is returning undefined since the given node does not contain node.
Solution is to retrieve the symbol of the given node from its parent node or parent's parent.


aciccarello:
@fencim do you know if we still need a try/catch around `this.checker.getTypeAtLocation(node)`?

fencim:
I think try catch is not necessary if we add validation for the argument node. Since exception might occur if is null or undefined.

kamranayub:
If you can merge in latest master, the build should pass.

PR #335, opened 2016-11-14
Type definitions only for development

DaSchTour:
@types are only needed for development. Installing them as dependencies installs them on every project that uses typedocs which may lead to duplicate identifiers if @types isn't used in the project or if @types expose functions that are not really available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,6 @@
     "node": ">=0.10.0"
   },
   "dependencies": {
-    "@types/fs-extra": "0.0.33",
-    "@types/handlebars": "^4.0.31",
-    "@types/highlight.js": "^9.1.8",
-    "@types/lodash": "^4.14.37",
-    "@types/marked": "0.0.28",
-    "@types/minimatch": "^2.0.29",
-    "@types/shelljs": "^0.3.32",
     "fs-extra": "^0.30.0",
     "handlebars": "4.0.5",
     "highlight.js": "^9.0.0",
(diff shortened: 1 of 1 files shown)


blakeembrey:
No, this was already correct. If we're using them as a dependency, the type must also be a dependency otherwise you need to re-install them all again. If there's an issue with the definition causing duplicates, that needs to be fixed in DefinitelyTyped first.

DaSchTour:
That can't be fixed with DefinitleyTyped because I'm not using @types but still typings. And that's why there are duplicate definitions. Typings are not needed for runtime!

blakeembrey:
@DaSchTour I didn't say they are needed for runtime. But they are needed for sub-dependencies that are exposed transitively. I can feel your pain, I still use Typings myself too because `@types` is still not complete, but saying the solution is to remove all dependencies is even more wrong. Have you tried changing `types` in `tsconfig.json`?

DaSchTour:
@blakeembrey I still don't get why @types are added to dependencies while they are not needed at runtime. Typings are for development only. NPM-Package should be like a binary. Just executable. So there is no need to add type definitions as there also shouldn't be any typescript left for runtime.

blakeembrey:
I'm not sure you understand the difference here then. `devDependencies` would mean that **you** have to install the @types. Again, I said **nothing** about runtime. It's needed for TypeScript compilation.

DaSchTour:
But I don't want and don't need to compile typedoc package so I don't want the typings.

blakeembrey:
Again, this **is not** about compiling the typedoc package. This is for other people using it programmatically.

Issue #340, opened 2016-11-17
Support `baseUrl` and `paths` as TS compiler options


lennartjansson:
I need these tsconfig options to be able to run typedoc on my project without compiler errors, since I'm relying on path mapping using the `paths` option.

If you think this is a relatively easy task, I may be able to write the PR myself. In this case, advice/guidance/pointers to the correct area of the code would be appreciated. Thanks!

aciccarello:
Hello @lennartjansson, it would be great if you put together a PR. The current maintainer took over the project from someone else so no one is an expert on the code base. Various people have been helping improve the TypeScript 2.0 support starting with #234 so your help would be greatly appreciated.

I do know that those compiler options were set to be temporarily ignored in #308 so that the compiler would run again. So I would start by removing them from that array and figuring out how to get the values to be used correctly.

lennartjansson:
Cool, will take a swing at it.

> So I would start by removing them from that array and figuring out how to get the values to be used correctly.

I did try this -- `baseUrl` seems to parse with no further issues, but the `paths` option led to the error messages
```
Error: Unknown option `paths.*.0`.
Error: Unknown option `paths.*.1`.
```
etc., which suggests it's trying to parse object keys as separate options... typedoc does not consume any compiler options with object or array types already, right? The error message looks similar to the one from https://github.com/TypeStrong/typedoc/issues/234#issuecomment-242051948, for example.

aciccarello:
@lennartjansson I haven't really looked into the relevant typedoc code but from browsing the [compiler options docs](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Compiler%20Options.md) it doesn't look like there were any array types prior to 2.0

aciccarello:
Closed via #342

PR #342, opened 2016-11-22
Support `baseUrl` and `paths` compiler options

lennartjansson:
Fixes issue https://github.com/TypeStrong/typedoc/issues/340.

The `paths` option, which is expected to be a JSON object mapping strings to arrays of strings, comes with an internal TS option declaration with property `map: 'object'`, so we special-case the option-setting logic when we find that property on the declaration. This sticks the whole `paths` object unchanged into `options.compilerOptions`, which is forwarded on to `ts.createProgram` in the converter.

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -97,14 +97,16 @@ export class OptionDeclaration
                 value = value || "";
                 break;
             case ParameterType.Map:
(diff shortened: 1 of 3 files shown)


lennartjansson:
Looks like all the recently opened PRs are failing with the same compile error in CI. Is there something I need to fix here before this is ready to merge?

kamranayub:
@blakeembrey This one can be merged safely looks like

PR #353, opened 2016-12-02
Import lodash into converter.ts

aciccarello:
This missing import was breaking the PR builds. Lodash is already imported in other files.

### DIFF
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -1,6 +1,7 @@
 import * as ts from "typescript";
 import * as _ts from "../ts-internal";
 import * as Path from "path";
+import * as _ from "lodash";
 
 import {Application} from "../application";
 import {ParameterType} from "../utils/options/declaration";


kamranayub:
Can this be merged ASAP (doesn't even need a package update) since doing any PRs against this repo requires this change in order to build the project properly... (and we have to currently pull this in manually to our fork but leave it off during commits to avoid messing with this PR).

blakeembrey:
Sorry @kamranayub. Is anyone else willing to help maintain this module yet? I had unsubscribed from notifications after the last bout of issues barrelled in with nowhere to go, I just couldn't keep up.

kamranayub:
@blakeembrey I've taken cursory looks at the codebase but don't think I could officially maintain it, we use it in our project so we rely on it. This is the only solid TS documentation choice so I wish *someone* would take it up, I don't mind being a contributor but I don't have the time to be a full maintainer. 😞

blakeembrey:
That's understandable. I would have loved to make the time for this too, but as we've both found it's a non-trivial amount of work to get it into the states we want 😄 I've subscribed again and I'll try to keep up with new issues.

PR #354, opened 2016-12-05
Add lodash import necessary for compilation.

lddubeau:
First attempt at compiling the code and I get a compliation error due to a missing import. 

This fixes the error.


WasabiFan:
I think this is equivalent: #353

lddubeau:
Right. I ran into the problem shortly before that PR was posted. I did a search at that time but did not find any mention of anybody having a problem with lodash. I did not have time to submit a PR until now but did not recheck issues/PRs before submitting. Oops...

Issue #301, opened 2016-10-05
Support more tsconfig.json scenarios


brphelps:
Right now, the tsconfig.json reader is somewhat limited -- it's only looking for "files" sections and doesn't take into account other significant parameters like "exclude" and "compilerOptions.rootDir". 

Since these are critical to being able to process a tsconfig.json the same way a TypeScript compiler would, support should be added.

aciccarello:
I took a look at the status of Typescript 2.0 support. As a workaround, several options were temporarily ignored via #308 to allow builds to continue but people using the new features may not be able to compile. See #234 for the original discussion.

|feature|issues|PRs|status|
|--------|------|----|------|
|baseUrl|#340|#342|Awaiting #353|
|paths|#340|#342|Awaiting #353|
|include|#326| |Needs PR|
|lib|#315| |Needs PR|
|types|#311| |Needs PR|
|typeRoots|#305| |Needs PR|

Gerrit0:
As far as I know - 0.20 respects all compiler options and behaves the same as tsc does.

PR #359, opened 2016-12-12
Have the typescript package perform as much processing of the configuration file as possible.

### LINKED ISSUE #326: Typedoc doesn't work with include globs
We use TS 2.0 globs to include source files. But it seems typedoc doesn't care of the `include` option in tsconfig - so that the generated doc is empty.
Below is an example of our `tsconfig.json`:
```
  "files": [
    "typings/index.d.ts"
  ],
  "include": [
    "lib/**/*.ts*",
    "test/unit/**/*.ts*",
    "scripts/**/*.ts"
  ],
  "rootDirs": [
    "./lib/client",
    "."
  ]
```

It also ignores the source directory I pass to the command (I could pass non-existing directory - and get no error). My command is:
`typedoc --out tsdocs ./lib --ignoreCompilerErrors --theme minimal`

Is this a bug or not supported feature? Or am I doing something wrong?

Typedoc version: 0.5.1

lddubeau:
This is proposed to fix #326. It *may* also have an impact on some of the other issues in #301.

This is not a complete PR yet but I'd like some help before moving forward. I quickly discovered while working on this that there is currently no test in the test suite that covers the code I changed. (I know because I had some pretty big errors when I copy-pasted into typedoc's codebase the code from the prototype file I created to figure how I'd load a `tsconfig.json` outside `tsc`. The big errors I had did not cause any test failure!) I'm not sure though what kinds of tests are desired. Should I create an  `Application`, and check whether it is seeing the right configuration options? Is there a better way to do this?

(The first commit in this PR is fixing the same problem as in #353. I need it there, otherwise the code does not compile. I'll remove it once #353 is merged.)

### DIFF
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -1,6 +1,7 @@
 import * as ts from "typescript";
 import * as _ts from "../ts-internal";
 import * as Path from "path";
+import * as _ from "lodash";
 
 import {Application} from "../application";
 import {ParameterType} from "../utils/options/declaration";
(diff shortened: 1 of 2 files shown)


src/lib/converter/converter.ts
```diff
@@ -1,6 +1,7 @@
 import * as ts from "typescript";
 import * as _ts from "../ts-internal";
 import * as Path from "path";
+import * as _ from "lodash";
```
aciccarello: Just want to note that this is also a part of #353.

PR #363, opened 2016-12-21
Fix npm run test by running npm run build

seescode:
I noticed that npm run test was failing on master.  After I ran npm run build, npm run test started working fine.


kamranayub:
@blakeembrey this can safely be merged without a package update

seescode:
I made the changes except the inlineSources stuff.  I'm not familiar enough with how sourcemap internals work.

Let me know if there's anything else you see. Thanks

kamranayub:
Well I know the source maps would only be used for debugging Type Doc
itself, and I've never had to use the inline option, it's always just
worked fine without it. I think we will keep having crossplat issues unless
we ignore the generated JS and Map files (possible) or we remove that flag.

blakeembrey:
We need `inlineSources` for debugging with source maps - I'm not publishing the original JS files.

@kamranayub Could you elaborate on the problem for me? I must not be seeing it. What's the issue with the source maps and I'll try to review, but were those specs used for diffing the output or is that somewhere else in the test script?

kamranayub:
If you take a look at the diff for the source map, you can see it's taking
the line endings of the current platform. So if someone on Windows
recompiles, they'll get a diff if the last person that committed worked on
Linux.

Like I said, at least with my projects I've never had to enable inline
sources and I use VS Code, Code supports regular source maps and it's cross
platform so it works for everyone.

Those rendered files are generated when the specs are built and they again
are generated with whatever the current platforms line endings are, so I
think it's best to keep them out of Git and have the build/test command
generate them ad-hoc.

kamranayub:
And yes, they are used for the diff testing

blakeembrey:
> Like I said, at least with my projects I've never had to enable inline sources and I use VS Code

This has nothing to do with being "cross platform". You need the source so that you can map from the generated output back to the original file - there's no way around it.

> So if someone on Windows recompiles, they'll get a diff if the last person that committed worked on Linux.

How? The source code is exactly the same. If the diffs are on JavaScript files, there's a simple fix. Use `tsconfig.json` to specify the line endings so TypeScript generates it correctly.

kamranayub:
The line endings trick in tsconfig would fix it then I think.

I don't know what the benefit of inlining source does, by default it's
false in tsconfig and I've never had to turn it on to use the Node
debugger. The source is next to the map, correct? Maybe I'm missing
something but I know for a fact I've never had to turn on that flag. I just
tested it on Tuesday since I had to do some debugging on a sample project
I'm writing.

kamranayub:
OK--I see the issue, the `index.ts` file is not next to the generated JS/Map file. However, inline sources is only useful if the source is not in the workspace, and it is, so debugging should still work fine without inline source turned on.

The problem is that the source **is not the same** on each platform. Commit Unix style, Checkout Windows Style. So if I clone the repo, all my files have CRLF:

image

So, when I run TSC and it outputs a source map and the source is inlined, you get `\r\n` for line endings:

image

If you, Blake, clone the repo on Linux you've checked out Linux style, therefore when you run the compiler, the source map will have `\r` causing a difference.

That's why right now in the repo, the source map(s) have `\r` and this gentleman is working on Windows so he is getting a diff with the source map.

Does that help explain?

kamranayub:
I just tested in my branch. Debugging works (in VS Code) fine if I have the following launch config:

```
{
            "type": "node",
            "request": "launch",
            "name": "Launch Program",
            "program": "${workspaceRoot}\\bin\\typedoc",
            "args": [
                "--out", "tmp"
            ],
            "cwd": "${workspaceRoot}",
            "outFiles": [],
            "outDir": "${workspaceRoot}/lib",
            "sourceMaps": true
        }
```

And I can turn off inline source flag without issues.

image

So, maybe whatever sebastian had originally used to debug didn't support debugging without inline sources but we know VS Code does and VS Code is cross-platform, so I don't see a reason to keep using inline sources.

blakeembrey:
@kamranayub I mentioned this above, but I changed it when I started maintaining the project. You're probably testing locally instead of from NPM. I don't publish the original TypeScript files to NPM, see https://github.com/TypeStrong/typedoc/blob/master/package.json#L62-L70.

blakeembrey:
> Does that help explain?

I believe so. Do you know where map files are being diffed? Maybe it'd be easier to skip anything using original sources altogether.

kamranayub:
I think it's only `index.js.map` TBH, just the one annoying file :)

blakeembrey:
Oh, because it's checked in 😦 Yeah, that's a legacy location because TypeDoc had it located there originally. I'd be happy to move it inside `src/` to match the rest of the project properly and so it's not checked in.

kamranayub:
@blakeembrey in my new PR I removed the need for inline sources anyway, so maybe it doesn't matter.

Also, I think the tests are weird. When you run the `specs` grunt task, it runs the `build-specs` task which generates all the `spec.json` files. However, this seems wrong because this means that tests will pass every time as its comparing the newly generated output with the output from TypeDoc and they'll be the exact same (because it *just* generated it). I think the specs.json should only be regenerated when you have stable tests, so build-specs should only be called by a human whose reviewed the output. That way, when you refactor/add features, you are checking for regressions in the output of the spec JSON.

blakeembrey:
@kamranayub I don't see where you "removed the need for inline sources", there's no fix change you've applied that fixes removing it, you've just deleted it.

As for the tests, happy to have them fixed. I think it's an easy fix and it's currently just a usage issue. Should split "build and test" apart.

kamranayub:
@blakeembrey you're right, I just misunderstood--apologies. For the tests, maybe a separate PR and stuff to fix that. Otherwise this can probably be merged unless you want to move the index.ts file back into the src folder to avoid the line endings issue.

blakeembrey:
I believe this is resolved with https://github.com/TypeStrong/typedoc/pull/366, but let me know if it's still in the wrong. Cheers!

PR #365, opened 2016-12-25
[#356] Add support for TypeScript 2.1.4

kamranayub:
This PR adds support for TS 2.1.4. Replaces #356 

## TODO

- [x] Check and test output for excaliburjs/Excalibur API
- [x] Update ts-internal comment references

## Notable Changes

- Added VS Code debugging support for tests
- ModifierFlags were separated from NodeFlags so I had to update all those references
- Constants are treated differently in 2.1, as [documented here](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#better-inference-for-literal-types)

Right now, TypeDoc will output an "unknown" type for constant declarations because they are no longer marked as Intrinsic types, instead they are Literal types. I think to properly support constants, there might need to be a new `ReflectionKind.Constant` so that constants can be grouped separately from variables.


tsconfig.json
```diff
@@ -7,139 +7,14 @@
         "preserveConstEnums": true,
         "declaration": true,
         "sourceMap": true,
-        "inlineSources": true,
```
blakeembrey: Please don't remove this. I've explained it a few times now, this is needed for debugging when you install from NPM.

tsconfig.json
```diff
@@ -7,139 +7,14 @@
...
-        "src/lib/utils/options/readers/typedoc.ts",
-        "src/lib/utils/options/sources/component.ts",
-        "src/lib/utils/options/sources/index.ts",
-        "src/lib/utils/options/sources/typescript.ts",
-        "src/lib/utils/plugins.ts",
-        "src/index.ts"
+    "include": [
+        "src/**/*.ts"
+    ],
+    "exclude": [
```
blakeembrey: Can you remove the empty exclude?

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
```
blakeembrey: Please don't check the `.vscode` files into this project.

blakeembrey:
This PR is good, but it would be easier broken into separate PRs for the breaking functionality. I'll merge once those changes are fixed, thanks @kamranayub for your help!

tsconfig.json
kamranayub: Ah yes, sorry, I can do that.

tsconfig.json
kamranayub: Exclude has to be present for TS to compile it correctly, because the default exclude is the current working directory.

```
error TS18003: No inputs were found in config file 'E:/Development/Contrib/typedoc/tsconfig.json'. Specified 'include' paths were '["src/**/*.ts"]' and 'exclude' paths were '["./"]'.
```

.vscode/launch.json
kamranayub: They are helpful for contributors, otherwise contributors have to manually figure out how to set up the debugger (this can be especially difficult for TS-based projects to know what the correct parameters are). It would be preferable to leave them in, there's nothing specific to my workspace that's in the files.

kamranayub:
This is what constants look like in 2.1:

image

It's acceptable for now I think until proper Constant support is added.

tsconfig.json
blakeembrey: That sounds like a bug for someone else then, that's not a normal compiler issue but something must be mutating our config incorrectly.

.vscode/launch.json
blakeembrey: The thing specific to your workspace is VS code. Should we check in configurations for other IDEs too?

tsconfig.json
blakeembrey: Looks like it's been reported https://github.com/TypeStrong/grunt-ts/issues/381.

kamranayub:
@blakeembrey when you have time (no rush) I had responded to your feedback 👍

.vscode/launch.json
kamranayub: I like to optimize my projects for a specific xplat IDE, which is VS Code for me--it provides a happy path for contributors. However, you are the maintainer so would you be OK adding the configuration to the CONTRIBUTING readme? That provides a copy/paste way for contributors who want to use VS Code at least.

tsconfig.json
kamranayub: What would your preference be? Should I leave it in? It doesn't seem to break anything.

nycdotnet:
Hi - I updated grunt-ts to support `include` now, so the empty exclude should no longer be needed.  Would you mind checking out grunt-ts@6.0.0-beta.9 ?

PR #370, opened 2017-01-01
Add `--listInvalidSymbolLinks` option to list broken symbol links after generation

kamranayub:
image

Just added a `--listInvalidSymbolLinks` switch to aid in fixing broken links due to refactoring/changes in large codebases. It does not break the generation, although that could be done too.

Let me know if you want any of the messaging changed. Note that the MarkedLinks plugin has no context for the exact symbol/signature it is parsing, it is only passed the text to parse so that's the most I could expose. I am not familiar enough to figure out how to pass in exact line/file reference info :)


blakeembrey:
Looks nice!

PR #373, opened 2017-01-04
Make CLI compiler options take precedence over tsconfig

### LINKED ISSUE #371: command line arguments should override tsconfig.json options
From the `tsc` docs:

>Compiler options specified on the command line override those specified in the tsconfig.json file.
-- https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

I think typedoc should have the same behavior as the typescript compiler in terms of options precedence. For example, `--target ES6` should override the target specified in the tsconfig.json.

jdanyow:
From the `tsc` docs:

>Compiler options specified on the command line override those specified in the tsconfig.json file.
-- https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

I think typedoc should have the same behavior as the typescript compiler in terms of options precedence. For example, `--target ES6` should override the target specified in the tsconfig.json.

fixes #371

### DIFF
--- a/src/lib/utils/options/readers/tsconfig.ts
+++ b/src/lib/utils/options/readers/tsconfig.ts
@@ -82,10 +82,6 @@ export class TSConfigReader extends OptionsComponent
             delete compilerOptions[key];
         }
 
-        _.merge(event.data, compilerOptions);
-
-        if ("typedocOptions" in data) {
-            _.merge(event.data, data.raw.typedocOptions);
-        }
+        _.defaults(event.data, data.raw.typedocOptions, compilerOptions);
     }
 }


src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -82,7 +82,7 @@ export class TSConfigReader extends OptionsComponent
             delete compilerOptions[key];
         }
 
-        _.merge(event.data, compilerOptions);
+        _.defaultsDeep(event.data, compilerOptions);
```
blakeembrey: Shouldn't this just be `_.defaults`?

jdanyow: I chose `defaultsDeep` because I thought it might be more future proof. Let me know if you would like the PR amended.

blakeembrey: I'm not sure why it'd be future proof? Wouldn't it mean odd usages of compiler options will break because of merging instead of overriding now? E.g. `paths` would deep merge now when you might be overriding?

jdanyow:
Good call- would be weird for `lib` too. I've switched to `defaults`.

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -82,7 +82,7 @@ export class TSConfigReader extends OptionsComponent
             delete compilerOptions[key];
         }
 
-        _.merge(event.data, compilerOptions);
+        _.defaults(event.data, compilerOptions);
 
         if ("typedocOptions" in data) {
             _.merge(event.data, data.typedocOptions);
```
blakeembrey: What do you think about this one too? Seems like this would also be better with `_.defaults` for the same reason. In which case, this whole thing can probably change to:

```ts
_.defaults(event.data, data.raw.typedocOptions, compilerOptions)
```

jdanyow:
Even better! Thanks for all the feedback. 

p.s. I confirmed that `_.defaults` accepts `undefined` args in case `data.raw.typedocOptions` is undefined.

src/lib/utils/options/readers/tsconfig.ts
jdanyow: Even better! Thanks for all the feedback. 

p.s. I confirmed that `_.defaults` accepts `undefined` args in case `data.raw.typedocOptions` is undefined.

PR #383, opened 2017-01-14
Converter for the "this" type

makana:
Created a converter for the type "this" that is often used in chainable
methods.


blakeembrey:
@makana Were you still interested or needing this? Can you share more information on how it works?

makana:
@blakeembrey 

Yeah, I'm still interested in having this.

This converter is just for chainable methods or similar that have a the ["this" type](https://github.com/Microsoft/TypeScript/pull/4910) as return type.
```ts
export class Test
{
  public chainableMethod(): this
  {
    // do stuff
    return this;
  }
}
```

The converter uses the intrinsic type model to output the word "this".
I don't think we need to work out a new type model here, since there would be no real benefit (link to something -> doesn't make sense, we are already in the class, ...).

Without this converter, having methods with the "this" type led to some null or undefined errors (I have to take a look again to specify them) because it wasn't represented by a type model instance at all.


Since my pull request was some days ago, I don't know if the this type is now supported without my changes. I will try later.

Regards,
makana

blakeembrey:
Thanks. If you're willing to update the PR and post a screenshot of what the change looks like, I'd be happy to merge 👍

makana:
Hi @blakeembrey ,

I updated the PR. Unfortunately the checks failed. I'm sure the test passes, but exceeded with a timeout...

Here a screenshot of how the output looks like without the converter:
without converter

Note the missing return type.

here with converter:
with converter

blakeembrey:
Thanks @makana!

Issue #350, opened 2016-11-28
@types in dependencies causing build failures


crazytoucan:
Unfortunately, we just hit the `dependencies` vs `devDependencies` issue as discussed in https://github.com/TypeStrong/typedoc/pull/310#pullrequestreview-4978799. Specifically, our project:

- Uses a different version of Lodash / its own `@types/lodash`.
- Has a dev dependency on `typedoc`, which installs a separate version of `@types/lodash`

Both `@types/lodash` installations are being picked up by the typescript compiler, causing duplicate definition warnings. Per discussion in that thread, can we move those types back out to `devDependencies` to prevent the unnecessary downstream installation?

If any of the `@types/` modules are actually needed by the generated `typedoc.d.ts` (seems like `@types/handlebars` might fall into this category?), I propose we inline those typings (re-expose them manually) so that we don't cause any build inconsistencies for other folks that might be using a different version of handlebars.

Will you all accept a PR for this? Any concerns about the approach?

DaSchTour:
I already made a PR for this #335 but it seams the project is close to death
Have to remove typedoc from my projects because this creates a lot of trouble 😡

blakeembrey:
Answered in https://github.com/TypeStrong/typedoc/pull/335#issuecomment-268070899. The definition needs to be fixed in DefinitelyTyped - which it may already have been, because this was brought up to them a few weeks ago.

blakeembrey:
If this seriously needs to be done for TypeScript to function, it's also a flaw in TypeScript that should be fixed. There's no reasonable expectation here for module authors to require duplication of dependencies to work.

DaSchTour:
There is, because Typdoc add typings that are not needed for the project. Modules show up that are not available in the project and create conflicts with typings installed from other sources like typings. I don't see any case were typings are needed at runtime!

PR #384, opened 2017-01-15
Moved @types to devDependencies.

illuxio:
@types are not necessary at runtime. Therefore, @types should be a devDependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,6 @@
     "node": ">= 4"
   },
   "dependencies": {
-    "@types/fs-extra": "0.0.33",
-    "@types/handlebars": "^4.0.31",
-    "@types/highlight.js": "^9.1.8",
-    "@types/lodash": "^4.14.37",
-    "@types/marked": "0.0.28",
-    "@types/minimatch": "^2.0.29",
-    "@types/shelljs": "^0.3.32",
     "fs-extra": "^0.30.0",
     "handlebars": "4.0.5",
     "highlight.js": "^9.0.0",
@@ -49,6 +42,13 @@
     "typescript": "2.1.5"
   },
   "devDependencies": {
+    "@types/fs-extra": "0.0.33",
+    "@types/handlebars": "^4.0.31",
+    "@types/highlight.js": "^9.1.8",
+    "@types/lodash": "^4.14.37",
+    "@types/marked": "0.0.28",
+    "@types/minimatch": "^2.0.29",
+    "@types/shelljs": "^0.3.32",
     "grunt": "^1.0.1",
     "grunt-cli": "^1.2.0",
     "grunt-contrib-clean": "^1.0.0",


blakeembrey:
Not true. Please read the existing discussions.

aciccarello:
@illuxio See #335, #350, and [#310 (review)](https://github.com/TypeStrong/typedoc/pull/310#pullrequestreview-4978799) for prior discussions.

illuxio:
@blakeembrey After reading the prior discussions I'm speechless. Enforcing build dependencies that a user may need, would be for me like forcing the user to install another OS because he/she may need it. Other project for example based on compiling c have a readme that states which additional resources you may need and yes you have to take care of.

blakeembrey:
@illuxio That is a pretty dumb argument and has zero relation to an OS. They're regular dependencies. I can't help it if you don't happen to use that code path. How would you like it if every NPM dependency you installed you have to read through their `package.json` and install the dependencies yourself?

blakeembrey:
@illuxio What exactly is your problem? You've put absolutely zero effort into explaining why this PR is remotely valid and really have just wasted my time by insulting instead of being constructive. If you can't understand something, you should learn about it yourself or try it out before assuming you're correct. If you think it's best that projects ship with zero dependencies and you should have to re-install everything every time, each to their own, but that's not how the node ecosystem works.

illuxio:
Actually I would like it, because when I use a package as library I have to read a documentation anyway! 

But in order to find a solution. How about putting @types in the optional dependency section?

blakeembrey:
@illuxio No, they are a dependency. Please explain what problem you're having so I can help you find an appropriate solution first, that'll be more productive here.

> With this you force users also to use maybe old typing that may not fit after a package update

That shouldn't matter though, since you aren't forced to use them - they are only for `typedoc`.

blakeembrey:
It seems to me that, as I noted in previous discussions, your issue would be with DefinitelyTyped being incorrect. For example, I found that https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d568404f00bf57c1f00c9cdb71545d77d1480471/handlebars/index.d.ts#L293 is still written in global format. To be correct, this should really be written in external module format. If that were done, none of our dependencies would be incorrectly global. The fact that the default behaviour of TypeScript is to pick up these globals is an unfortunate behaviour that I dislike and have comment to TypeScript about. But these issues are related to TypeScript's implementation of modules, and forcing every module author to re-install every sub-dependency would put us back in the days of TSD which is not productive for anyone.

illuxio:
One of my main problems is the "Duplicate identifier" error given by typescript, because two packages include different version of their @types as dependencies. For example @types/lodash@3.10.0 and @types/lodash@4.14.50. However, I am only using lodash@4.14.x and know that the @types/lodash@3.10.0 is irrelevant to me and makes only trouble. How would you solve this issue?

blakeembrey:
@illuxio Thanks, that makes sense now. So, all `@types` dependencies should be written as external modules (if they're external modules) and that's the issue here - some legacy DefinitelyTyped stuff. It's really unfortunate that TypeScript decided to use `@types` for everything instead of separate namespaces like I did with Typings, but we just have to live with it now - the unfortunate side-effect is published `@types` might be global **or** external. External ones are what we really want to be using, because they would not cause any conflicts for you.

The latest `lodash` appears that it's written correctly - see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/lodash/index.d.ts. Do you know where `lodash@3.10.0` comes from?

blakeembrey:
From my manual exploration, it seems from our dependencies `highlight.js`, `marked` and `handlebars` are still written incorrectly and would need to be updated in DefinitelyTyped. I'm not sure why Lodash would be causing an issue at this time for you, unless it's coming from a different package?

illuxio:
As I mentioned it was an example. At the current state lodash is no issue and back then when I had this issue I removed the causing package from our package.json because we weren't using it anymore. But this was still in my mind.

But thanks for your time. I will think about what you said.

blakeembrey:
I understand the concern, believe me, but it's detrimental to enforce on module authors as it makes the development experience so much worse. The real issue, to me, is the fact that these conflicts can simply occur at any time thanks to the TypeScript compiler and dependency set up. If there were two namespace (global and modules), then this entire discussion could have been averted. Don't worry about it too much though, I've spent countless hours describing the issues that exist to others too, I just strongly disagree the solution would be to make development suck any more than it needs to.

hdeshev:
Another guy that got bitten by the "@types/* as dependencies" issue here.

I realize you want to make people's life easier when they want to use typedoc as a library, but it seems that is hurting the majority of users that just call the binary to get their API reference right. Would it be better if there was a documentation section that tells "library" users which @types dependencies they need to reference?

Would you consider accepting a PR that adds a script installing the extra dependencies, so "library" users could run `./node_modules/.bin/typedoc-dev-dependencies` and have all necessary dev dependencies set up for them automatically?

aciccarello:
@hdeshev thanks for offering another solution but I don't think this is TypeDoc's problem. Isn't the real solution is fixing the `@types` definitions for everyone?

hdeshev:
@aciccarello, I think we both agree that:

1. This is a problem, and it affects TypeDoc users.
2. This isn't TypeDoc's problem.

That doesn't help people like me very much though. I still had to fork TypeDoc in order to fix NativeScript's documentation build. That is the only thing I could do, since, even if I'd forked the broken lodash typings, I would still have had to fork TypeDoc in order to make it pull in my fixed version of those typings.

I think that there is value in solving the problem in a way that will:

1. Reduce the amount of packages most people download.
2. Eliminate the chance of future type clashes and other problems inflicted by bad typings for most end users like me.

All that at the tradeoff of having to document the typing dependencies needed by people who use TypeDoc as a library (TypeDoc development shouldn't be affected, as those dev dependencies will make sure the typings are available.). So, to rephrase my question from the previous comment: is this an acceptable tradeoff?

aciccarello:
@hdeshev Right, so the main problems that we are trying to solve are the "type clashes and other problems inflicted by bad typings for most end users". My gut reaction is that **users should submit PRs to DefinitelyTyped to fix the broken typings** as that is the source of the issue. Generally, this isn't too hard and helps the larger community.

Manually installing the `@types` dependencies of TypeDoc may seem trivial but I agree that fundamentally you shouldn't have to do that to use it as a library. It gets more and more complicated with multiple levels of dependencies and different versions of libraries being used.

 **But still that does leave a period of time while type definitions are being fixed when TypeDoc is unusable therefore I think we should have a workaround.**

Ideally I would hope that we could make the workaround opt-in. So if you are having trouble with TypeDoc while a type definition is being fixed you could change some configuration which prevents the conflict issues. This would temporarily prevent the error but indicates that there needs to be a more complete solution. That's what I'm thinking at the moment, but I'd be open to discussing it further.

illuxio:
@blakeembrey: Hi it's me again,

today I had to face the following issue:

```
node_modules/@types/lodash/index.d.ts(12762,29): error TS2304: Cannot find name 'object'.
node_modules/@types/lodash/index.d.ts(19507,15): error TS2428: All declarations of 'WeakMap' must have identical type parameters.
node_modules/@types/lodash/index.d.ts(19507,33): error TS2304: Cannot find name 'object'.
```

Currently we have to using typescript 2.1 and therefore use a proper typedoc version (currently 0.5.5). However, due to the recent update of @types/lodash (which require ts 2.2) our build is failing. But the typing is just a dependency of typedoc and we do not need it. :(

Any suggestions?

blakeembrey:
@illuxio I'm really sorry, I hate this auto-updating and breaking users approach to `@types`. Unfortunately it's just what TypeScript has given us. Can you open an issue on TypeScript or DefinitelyTyped about this, they need to see that what they're doing - with the lack of real semver - is breaking users. On the TypeDoc side, would it help if we pin the versions of `@types` we're using?

illuxio:
@blakeembrey: There are some issues occurring that states my problem e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14384 . However the solution is sub-optimal due to the lost of version control with the tags. 

Form typedoc side, pinning sound reasonable due to the relation between typedoc and typescript. I would like this solution.

aciccarello:
If we pin the `@types` versions, is there any way for us to test that we won't cause errors for users when we do update?

dehru:
Hello, I'm having this same error with the WeakMap as mentioned above.

dehru:
Would it help for me to open a new issue instead of us all commenting on this issue which is unrelated?

blakeembrey:
Anyone can submit a PR to pin the versions for now. It'd be useful if someone, other than me, wants to pester TypeScript about this issue 😄  I think I've complained enough to them about the lack of real semver in the `@types` implementation.

dehru:
I'm a typescript newbie.  Would I pin down the version of @types/lodash or typescript?

aciccarello:
It would be the `@types` you'd pin down

Issue #170, opened 2016-01-01
How does exclude work ?


ocombe:
Hello,
I use the command line: `typedoc --options typedoc.json --exclude *.spec.ts src/app/`

But it generates documentation for my file "src/app/app.spec.ts".
Shouldn't it be excluded ?

enanox:
Hi, I'm on the same boat as @ocombe. 
Is the `--exclude` pattern a RegExp, or a pattern similar those used in Grunt/Gulp, such as `directory/**/*.ext`?

EMarginy:
Hi, got the same problem with exclude trying this: ```typedoc --media modules/ --exclude modules/**/tests/*```

dsebastien:
Same issue here. Also how to configure multiple exclusions?

caseyhoward:
I'm having the exact same issue. I want to ignore spec files. "<pattern>" is not helpful at all without an example of what a "pattern" is.

caseyhoward:
@ocombe I think the following should work:
```
typedoc --options typedoc.json --exclude **/*.spec.ts src/app/
```

typedoc uses [Minimatch](https://github.com/isaacs/minimatch) for globs. I believe *.spec.ts will only look for files in the current directory.

hverlin:
Hi, you can use for example: ``"exclude": "**/*+(e2e|spec|index).ts"`` if you want to have multiple exclusions.

The syntax seems to be ``+(pattern-list)``.

blakeembrey:
Seen here (https://github.com/TypeStrong/typedoc/blob/53bb762ad208f2be266ece767f6204159ce71a0d/src/lib/application.ts#L253), the pattern is a standard minimatch (glob) pattern in Node.js. The existing comments by @caseyhoward should work for you (make sure you quote so it's not automatically expanded by your shell).

daotoad:
Changing the docs to say that patterns are "standard" minimatch patterns would be very helpful.

blakeembrey:
You are welcome to submit a PR improving the docs.

YouriT:
I know this has been closed but I think that what would be interesting to know in the docs is what is given to minimatch for testing.

If you pass the whole path to the file, the pattern will be different than if you are passing what comes after `path/to/typescript/project/`. From my understanding, you make a full resolution and give it to the minimatch pattern but correct me if I'm wrong.

Zeioth:
Then there is no way to exclude an entire directory?

YouriT:
[...]

PR #387, opened 2017-01-18
TypeDoc now also honors the --exclude option on a list of files

dploeger:
This pull request lets typedoc honor the --exclude argument also for typedoc calls, that include a list of files. Previously --exclude only worked if you specify a directory.

As one could possibly argue, running typedoc with a list of files already excludes the files we don't want to use, but typedoc seems to also check tsconfig.json for input files and overwrites the list of the provided files.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -272,7 +272,7 @@ export class Application extends ChildableComponent<Application, AbstractCompone
             file = Path.resolve(file);
             if (FS.statSync(file).isDirectory()) {
                 add(file);
-            } else {
+            } else if (exclude && !exclude.match(file)) {
                 files.push(file);
             }
         });
--- a/test/typedoc.js
+++ b/test/typedoc.js
@@ -16,5 +16,13 @@ describe('TypeDoc', function() {
             Assert.notEqual(expanded.indexOf(Path.join(inputFiles, 'class.ts')), -1);
             Assert.equal(expanded.indexOf(inputFiles), -1);
         });
+        it('honors the exclude argument even on a fixed file list', function() {
+            var inputFiles = Path.join(__dirname, 'converter', 'class');
+            application.options.setValue('exclude', '**/class.ts');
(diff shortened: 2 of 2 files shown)


kondei:
I want this to be merged

aciccarello:
@kondei Have you tested this PR yourself?

dploeger:
Travis is actually green on this.

test/typedoc.js
```diff
@@ -16,5 +16,13 @@ describe('TypeDoc', function() {
...
             Assert.equal(expanded.indexOf(inputFiles), -1);
         });
+        it('honors the exclude argument even on a fixed file list', function() {
+            var inputFiles = Path.join(__dirname, 'converter', 'class');
+            application.options.setValue('exclude', '**/class.ts');
+            var expanded = application.expandInputFiles([inputFiles]);
+
+            Assert.equal(expanded.indexOf(Path.join(inputFiles, 'class.ts')), -1);
+            Assert.equal(expanded.indexOf(inputFiles), -1);
+        });
```
aciccarello: Could we add a few more test cases? There are quite a few GitHub issues regarding exclude so better testing will help us sort out what problems actually exist. For example, could you add a test with multiple excludes (see [#170 comment](https://github.com/TypeStrong/typedoc/issues/170#issuecomment-228983184))?

src/lib/application.ts
```diff
@@ -272,7 +272,7 @@ export class Application extends ChildableComponent<Application, AbstractCompone
             file = Path.resolve(file);
             if (FS.statSync(file).isDirectory()) {
                 add(file);
-            } else {
+            } else if (exclude && !exclude.match(file)) {
```
aciccarello: Just wanted to note that #395 suggested there was a problem here. Thanks for the fix!

dploeger:
Done. I've scanned through the exclude-issues, and could also only think about the multiple excludes test.

dploeger:
Thanks for the merge. When will this be available in an official release?

PR #389, opened 2017-01-19
Handles constructor return types

JegoYann:
Hello !
I need to get documentation about an interface constructor that returns a different type.
https://www.typescriptlang.org/docs/handbook/interfaces.html

```
interface ClockConstructor {
    new (hour: number, minute: number): ClockInterface;
}
interface ClockInterface {
    tick();
}
```

Currently, this typescript usage is not handle by Typedoc. Typedoc overrides the return type by the parent type.

Do not hesitate to comment if you need me to update something.
Great lib by the way !


src/lib/converter/nodes/constructor.ts
```diff
@@ -42,7 +42,10 @@ export class ConstructorConverter extends ConverterNodeComponent<ts.ConstructorD
             if (!hasBody || !method.signatures) {
                 var name = 'new ' + parent.name;
                 var signature = createSignature(context, node, name, ReflectionKind.ConstructorSignature);
-                signature.type = new ReferenceType(parent.name, ReferenceType.SYMBOL_ID_RESOLVED, parent);
+                 // If no return type defined, then use the parent one.
+                if (!node.type) {
```
blakeembrey: What about checking `signature.type` instead of `node.type`? It seems that there is an `else` condition in https://github.com/TypeStrong/typedoc/blob/22ccb660ae6eb5b3e136dc277ecc1d62fb04d016/src/lib/converter/factories/signature.ts#L65 but the `.type` is being completely erased anyway?

JegoYann: When there is no return type ( ex:`new();` ) then `signature.type == any` and is override with the `parent.name` by `new ReferenceType()`. Still it is possible to define an `any` return type even if not recommended (ex: `new() : any;`).
So checking `signature.type` is not concluant. `node.type` is not defined if no return type written and match the use case.
In my case, the execution is going through the [first extractSignatureType return](https://github.com/TypeStrong/typedoc/blob/22ccb660ae6eb5b3e136dc277ecc1d62fb04d016/src/lib/converter/factories/signature.ts#L58). No possibility to handle the use case here.

aciccarello:
Do we know why one of the tests times out in Node 6?

>   1) Converter access converts fixtures:
>
>   Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

JegoYann:
O_O No idea, I just merge master by converting `var` into `const`.

blakeembrey:
@aciccarello Legacy. It seems to happen randomly sometimes. Just restarted, we'll see if it passes this time.

Issue #378, opened 2017-01-12
General architecture changes / getting involved


dylans:
Not sure if this is the best place to raise this, but here it goes:

We've been working on documentation for [Dojo 2](https://github.com/dojo/meta/) and are interested in making some hopefully useful changes to improve TypeDoc. We'd rather reach out to collaborate than start yet another project.

Overall, some of our goals include:

# Decouple the parsing from rendering approach

The main reason for this is that we would probably prefer a different approach to how we render our API docs, and a fairly substantial part of TypeDoc is focused on how things will be rendered. This would also make it easier to include partial API docs, e.g. say you want to embed a small API subset into a tutorial.

We think that decoupling doc parsing and rendering into separate packages would lead to better architecture decisions by having one package that parses docs into an intermediate format, and then the separate rendering package that reads content in this format. We realize they are in separate modules today, but the dependency chain for the rendering side of things is pretty large. It would also likely remove a few patterns we see today such as copying backbone/events directly into the project, and including all of lodash rather than just the sub-packages that are needed.

Another goal with this would be to keep things smaller and easier for someone to follow, rather than a large, somewhat monolithic package for all things related to generating API documentation from TypeScript (I say somewhat, because things are separated into modules, but the assumption with the current architecture is that you want to use the current approach and opinions around rendering).

# Override code with comments

We have an implementation of an object composition system, [dojo/compose](https://github.com/dojo/compose), and TypeDoc currently makes supporting this nearly impossible because TypeDoc doesn't know what to do with objects that are created using our composition system. So we’d like to propose a way to extend TypeDoc to introduce other object type systems that can be re-used. Our initial thoughts here are to offer an optional ability to look to function and variable comments (like jsdoc) for type information, and override the inferred type if they are present.

# Model based design, annotations, and decorators

It does feel like there's quite a bit of complexity and code to model the model. It feels like there's a pretty strong drive to leverage Domain Driven Design (DDD) and annotations/decorators are commonplace. This approach is not widely used in JavaScript, which we feel may limit third party contributions due to the complexity it is adding currently.

Also Decorators are used pretty heavily, and the TS implementation and the version in ES.next are starting to diverge. It's probably ok, but it may be worth a rethink with this approach.

# Some minor nitpicks

* Remove remaining uses of `var`
* Use `util/fs` consistently
* Directory exports with `index.js`, we would probably change this approach

# POC

[...]

PR #396, opened 2017-01-29
Add typeArguments to ReferenceType.toString()

jiayihu:
First of all good job for the hard work on this project! ❤️ 
I'm using the `Converter` API to generate custom documentation for Angular components and I noticed that `typeArguments` are missing when calling `.toString()` on a `ReferenceType`.

### DIFF
--- a/src/lib/models/types/reference.ts
+++ b/src/lib/models/types/reference.ts
@@ -113,12 +113,18 @@ export class ReferenceType extends Type
 
     /**
      * Return a string representation of this type.
+     * @example EventEmitter<any>
      */
     toString() {
-        if (this.reflection) {
-            return this.reflection.name + (this.isArray ? '[]' : '');
-        } else {
-            return this.name + (this.isArray ? '[]' : '');
+        var name = this.reflection ? this.reflection.name : this.name;
+        var arraySuffix = this.isArray ? '[]' : '';
+        var typeArgs = '';
+        if (this.typeArguments) {
+            typeArgs += '<';
+            typeArgs += this.typeArguments.map(arg => arg.toString()).join(', ');
+            typeArgs += '>'
         }
+        
+        return name + typeArgs + arraySuffix;
     }
 }


src/lib/models/types/reference.ts
```diff
@@ -17,7 +17,7 @@ export class ReferenceType extends Type
      * If the symbol cannot be found cause it's not part of the documentation this
      * can be used to represent the type.
      */
-    name:string;
+    name:string;
```
aciccarello: Looks like a space was accidentally added here.

src/lib/models/types/reference.ts
```diff
@@ -115,10 +115,15 @@ export class ReferenceType extends Type
      * Return a string representation of this type.
      */
```
aciccarello: It might be helpful to add an example output string to the method docs

src/lib/models/types/reference.ts
jiayihu: I actually thought that my editor removed a trailing space :D Will fix it!

src/lib/models/types/reference.ts
jiayihu: You're right. I was also looking for some test for the class but I didn't it. Is there any? Thanks for your review (:

aciccarello: You're welcome. I don't think there is currently a test set up. The travis build seems to have run fine.

jiayihu:
I updated after your suggestions. I added also a commit because many test files were updated by, I suppose, a `postinstall` script with the new Typedoc API domain.

aciccarello:
Great! Having the example output helps clarify what is going on.

I'm surprised that the domain wasn't changed before. Ideally that would have been done in a separate PR.

jiayihu:
No problem, I removed the last commit (:

jiayihu:
Any update? Are you @aciccarello allowed to merge PRs? You're helping a lot in issues too, would be sad to leave the project inactive.

jiayihu:
@blakeembrey thanks for merging (: do you have any thoughts about adding collaborators to the project to keep it alive if you don't have much time?

aciccarello:
@jiayihu We've had a few conversations on #256 and #378. I do think that we could do more in inviting people to be involved. Maybe we could add some information to the readme.

blakeembrey:
If someone is willing to take on the work, I'm certainly happy to add them as a maintainer and collaborator. I added @aciccarello for that reason 😄  He also linked to the related issues which should shed more light - I do hope to give this project to someone who can take it to the right place eventually, since that person is not currently me.

PR #398, opened 2017-01-30
add exported flag to declaration when default export - Fixes #393

### LINKED ISSUE #393: excludeNotExported with default export declaration separate from variable creation not working
In a module where the default export is defined separate from the creation of the exported value, TypeDoc can't seem to map the exported value back to the internal variable.

Example module:

```ts
// keys.ts 

/**
 * A list of special keys
 */
const keys = {
    foo: 'foo',
    bar: true
};

export default keys;
```

This generates a page that looks like this:

screen shot 2017-01-26 at 3 08 00 pm

nicknisi:
Add `ReflectionFlag.Export` flag to a node's flag variable when the export
converter is run and then check for that flag in `createDeclaration`.

This fixes a scenario described in #393 where `export default` is used in a separate
statement from the variable creation, such as:

```ts
const foo = 'bar';
export default foo;
```

In the above example, the export converter cannot find the symbolmapping
for the `foo` declaration because the createDeclaration code returns
null when --excludeNotExported flag is set.

Looking at the tests, it doesn't look like any of them run with `excludeNoExport` set, and setting it makes several fail. Setting `excludeNoExport` and running the export-default test does pass with this fix, though. I could try looking into that here, or perhaps as part of the proposed work in #378?

### DIFF
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -62,6 +62,11 @@ export function createDeclaration(context:Context, node:ts.Node, kind:Reflection
         isExported = isExported || !!(modifiers & ts.ModifierFlags.Export);
     }
 
+    // if the node itself has its exported flag set, override isExported to true
+    if (node.flags & ReflectionFlag.Exported) {
+        isExported = true;
+    }
+
     if (!isExported && context.converter.excludeNotExported) {
         return null;
     }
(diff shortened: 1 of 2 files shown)


nicknisi:
Looking at this a bit more after I've had some time to dig into the TypeScript and TypeDoc code a bit more, and I think this might not be the best approach. I probably shouldn't be changing a Node's flags or adding a Reflection flag to it. I'll research a better approach.

nicknisi:
Closing this PR. Will open a new one with a better solution to the problem.

PR #410, opened 2017-02-08
Add directoryExists implementation to CompilerHost

jason0x43:
This PR adds an implementation of the optional `directoryExists` property to the CompilerHost class. This property must be present for TypeScript's automatic type discovery to work (like finding things in @types).


jason0x43:
Combining this with something else...

PR #411, opened 2017-02-08
Selftest updates

jason0x43:
This PR includes 2 commits. I'm not sure of the history behind the current way the repo is organized, so these may not be desirable changes, but they seemed reasonable to me.

1. The first commit removes the `noLib` option from the examples and self tests so that the standard lib for the current compiler will be used, and removes the repo's local copies of `lib.d.ts` and `lib.core.d.ts`. Some of typedoc's dependencies were using newer features (like the `Partial` type) that the included copies of the default lib lacked. This change seemed like a better way to go that maintaining multiple local copies of the default lib.
2. The second commit updates the copies of the generated test files in the repo. The default theme has been updated since they were last generated, so elements related to the default template are different (e.g., modernizr was removed). I'm assuming those are present for simple comparative purposes (e.g., run the tests and verify that the output hasn't changed).


jason0x43:
Closing this to rebase and combine it with something else...

PR #412, opened 2017-02-08
Use default lib, update CompilerHost

jason0x43:
There are 3 commits in this PR.

1. The first commit removes the `noLib` option from the examples and self tests so that the standard lib for the current compiler will be used, and removes the repo's local copies of `lib.d.ts` and `lib.core.d.ts`. Initially I had implemented commit 3 (below) first. Doing this enabled automatic type discovery in the TS compiler, and some of typedoc's dependencies were using newer features (like the `Partial` type) that the included copies of the default lib lacked. This change seemed like a better way to go that maintaining multiple local copies of the default lib.
2. The second commit updates the copies of the generated test files in the repo. The default theme has been updated since they were last generated, so elements related to the default template are different (e.g., modernizr was removed). I'm assuming those are present for simple comparative purposes (e.g., run the tests and verify that the output hasn't changed).
3. The third commit adds an implementation of the optional `directoryExists` property to the CompilerHost class. This property must be present for TypeScript's automatic type discovery to work (like finding things in @types).

My initial goal was just to add the `directoryExists` property to CompilerHost. I added the other two commits after running into problems running the self tests. While I can see an argument for preventing the compiler from automatically discovering third party types (in `@types`) during tests for better isolation, it also seems like typedoc should be able to run in a typical fashion (with default TS behavior) within its own project.


jason0x43:
Interesting. So, the selftests were frequently failing for me with errors like those in https://travis-ci.org/TypeStrong/typedoc/jobs/199618572, which are due to missing type definitions (those from the old version of the default lib were being used). The current failures appear to be due to timing issues (the VM took too long to run a test).

Closing and reopening to restart the tests...

jason0x43:
Oh, yeah, the type errors in https://travis-ci.org/TypeStrong/typedoc/jobs/199618572 were happening because of the addition of `directoryExists`. With that method on the CompilerHost, TS started looking up types automatically, and some of the @types installed for typedoc rely on newer TS features, so things started breaking. Using the actual TS default lib in tests takes care of the issue.

The current failures are happening because removing the `--noLib` option from the tests causes rendering to take a bit longer, so some of the tests are timing out.

jason0x43:
Missed a timeout. Restarting...

blakeembrey:
I believe this PR is ok, though given the large number of changes, can someone else review also?

jason0x43:
I can split up the PR if that's helpful. I could also remove the commit that updates the test cases (probably the majority of changed files). I wasn't 100% sure why those were there since they're generated; I just updated them since they kept showing up as changed, and the changes appeared to be due to template updates.

blakeembrey:
It should be ok, I looked over it and it seemed good. I'm a little confused where the GitHub metric is coming up with 25000 deletions, I only see a single large ~4000 line file deleted. The rest appear to be relevant updates.

jason0x43:
Mostly it was the other lib file (`lib.d.ts`), which was a bit over 17k lines.

PR #415, opened 2017-02-08
Support 'strictNullChecks' option

jason0x43:
This PR adds support for the `strictNullChecks` option (by allowing it to pass through to TS).


jason0x43:
Restarting build...

PR #417, opened 2017-02-09
TypeScript 2.2 OptionDeclaration#type is now Map

### LINKED ISSUE #414: TypeScript 2.2 OptionDeclaration#type is now Map
TypeScript 2.2 use `Map` instead of `Object` in OptionDeclaration#type (and maybe other places)

I created this patch

https://github.com/shlomiassaf/typedoc/commit/d6557609eaeabb2522bc79c74e6a58c511a867b5

I can't PR it since I did it live on GitHub...
The `map` property in `OptionDeclaration` is public so it is probably used in other locations and it needs to be `Object | Map` or if taking the easy way `any`

i'm new to this project so I can't really decide on those, it looks like some sort of refactoring is needed.

aciccarello:
Support Map and Object for backward compatibility

This is just a partial fix for TS 2.2

Closes #414

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -107,9 +107,16 @@ export class OptionDeclaration
             case ParameterType.Map:
                 if (this.map !== 'object') {
                     var key = value ? (value + "").toLowerCase() : '';
-                    if (key in this.map) {
+                    let hasKey = false;
+                    if (this.map instanceof Map) {
+                        if (hasKey = this.map.has(key)) {
+                            value = this.map.get(key);
+                        }
+                    } else if (hasKey = key in this.map) {
                         value = this.map[key];
-                    } else if (errorCallback) {
+                    }
+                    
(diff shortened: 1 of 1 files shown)


src/lib/utils/options/declaration.ts
```diff
@@ -107,9 +107,16 @@ export class OptionDeclaration
             case ParameterType.Map:
                 if (this.map !== 'object') {
                     var key = value ? (value + "").toLowerCase() : '';
-                    if (key in this.map) {
+                    let hasKey = false;
+                    if (this.map instanceof Map) {
+                        if (hasKey = this.map.has(key)) {
+                            value = this.map.get(key);
+                        }
+                    } else if (hasKey = key in this.map) {
```
jiayihu: Shouldn't be `this.map.hasOwnProperty(key)` to avoid checking in `prototype`?

blakeembrey:
It looks like I merged a similar feature with https://github.com/TypeStrong/typedoc/pull/448, sorry!

Issue #421, opened 2017-02-15
0.5.6 version pulling in extra files


gregvanl:
When running typedoc with a single file, the 0.5.6 version is now pulling in all .d.ts files in the project. This was not the behavior in 0.5.5 and might be due to the recent change around the --exclude switch https://github.com/TypeStrong/typedoc/commit/6ddb7f59766aa0929a94f170dd9165921925036b.

aciccarello:
Yes, there was a bug in #387 which makes the exclude option required. As was noted on 6ddb7f5, you can workaround this by adding a dummy `--exclude xx`. Thanks for submitting an issue for this. PRs are welcome.

jason0x43:
When you say 0.5.6 is pulling in all the .d.ts files, do you mean that docs are being generated for .d.ts files?

gregvanl:
Yes, our .json file which included generated types for one TypeScript .d.ts file went from 713K to 13M when all the node_modules TypeScript .d.ts files where pulled in. No change in the project, just an upgrade from TypeDoc 0.5.5 to 0.5.6.

jason0x43:
Hmmm...at a guess I'd say the issue is probably due to the addition of `directoryExists` to TypeDoc's CompilerHost class. This allows the TypeScript compiler's automatic type discovery to work, which will find types in `node_modules/@types` by default. If you're using `--includeDeclarations`, that could cause all the discovered types to be documented.

Try using a combination of `--externalPattern` and `--excludeExternals` to prevent docs being generated for those files while still allowing them to be used by the TS compiler during the generation process. I've been doing something like `--externalPattern '**/+(node_modules|tests|typings)/**/*.ts' --excludeExternals`.

PR #423, opened 2017-02-16
Only apply exclusion rules when available

jason0x43:
A recent change to the exclusion logic causes all files to be excluded by default (files  are only added to the input list if an exclude rule is provided _and_ it doesn't match). This PR causes files to be added to the input list if there is no exclude rule _or_ if the provided exclude rule doesn't match a file.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -272,7 +272,7 @@ export class Application extends ChildableComponent<Application, AbstractCompone
             file = Path.resolve(file);
             if (FS.statSync(file).isDirectory()) {
                 add(file);
-            } else if (exclude && !exclude.match(file)) {
+            } else if (!exclude || !exclude.match(file)) {
                 files.push(file);
             }
         });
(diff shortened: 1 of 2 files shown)


aciccarello:
Thanks! Could you also add a unit test to make sure that this error doesn't happen again?

jason0x43:
Will do

aciccarello:
This should fix #421

jason0x43:
I updated the unit tests to check that the exclusion rules are applied properly to both directory and file entries. The reason the tests were passing the after the initial update is that exclusion was handled different if entries in the input list were directories vs files, and the tests were only trying directories.

jason0x43:
I'm not entirely sure that this will deal with #421. At least, the issue I was experiencing is that 0.5.6 wasn't pulling in _any_ files (vs #421 saying that extra files were being pulled in) because all individual input files were being excluded if no exclude rule was provided.

Issue #377, opened 2017-01-12
TypeError: Cannot read property 'length' of undefined


sharwell:
I'm getting this error when compiling [antlr4ts](https://github.com/tunnelvisionlabs/antlr4ts) with typedoc.

```text
Using TypeScript 2.1.4 from J:\dev\antlr4ts\node_modules\typescript\lib
J:\dev\antlr4ts\node_modules\typedoc\dist\lib\converter\plugins\ImplementsPlugin.js:35
            for (var index = 0, count = classReflection.children.length; index < count; index++) {
                                                                ^

TypeError: Cannot read property 'length' of undefined
    at J:\dev\antlr4ts\node_modules\typedoc\dist\lib\converter\plugins\ImplementsPlugin.js:35:65
    at Array.forEach (native)
    at ImplementsPlugin.analyzeClass (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\converter\plugins\ImplementsPlugin.js:30:38)
    at J:\dev\antlr4ts\node_modules\typedoc\dist\lib\converter\plugins\ImplementsPlugin.js:85:27
    at Array.forEach (native)
    at ImplementsPlugin.onResolve (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\converter\plugins\ImplementsPlugin.js:79:41)
    at triggerEvents (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\utils\events.js:125:43)
    at triggerApi (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\utils\events.js:106:13)
    at eventsApi (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\utils\events.js:20:18)
    at Converter.EventDispatcher.trigger (J:\dev\antlr4ts\node_modules\typedoc\dist\lib\utils\events.js:261:13)
```

PR #425, opened 2017-02-17
fix(implementsPlugin): check if reflection has children

jiayihu:
Fixes https://github.com/TypeStrong/typedoc/issues/377


jiayihu:
Is it normal that it includes also last PR commits? Although on `Files changed` panel seems to be okay.

PR #426, opened 2017-02-17
fix(implementsPlugin): check if reflection has children

jiayihu:
Fixes https://github.com/TypeStrong/typedoc/issues/377

### DIFF
--- a/src/lib/converter/plugins/ImplementsPlugin.ts
+++ b/src/lib/converter/plugins/ImplementsPlugin.ts
@@ -38,6 +38,11 @@ export class ImplementsPlugin extends ConverterComponent
             }
 
             var classMember:DeclarationReflection;
+
+            if (!classReflection.children) {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


src/lib/converter/plugins/ImplementsPlugin.ts
```diff
@@ -38,6 +38,9 @@ export class ImplementsPlugin extends ConverterComponent
             }
 
             var classMember:DeclarationReflection;
+
+            if (!classReflection.children) return;
```
aciccarello: This should do the trick. The style does seem inconsistent though. I think the return should either be on a new line (like the `if (!classMember)`), or wrap the for loop.

jiayihu:
@aciccarello fixed style.

aciccarello:
Thanks for adjusting that. I know it's small but as someone who doesn't know all the code the consistency helps.

jiayihu:
No problem, I always use linters in my projects so I perfectly understand 👍

PR #429, opened 2017-02-20
refactor var usage to const/let

nicknisi:
Remove all var usage in the project and tests, and replace it with const/let usage. This PR touches a lot of files but the only changes in each file are to convert `var`to `const` or `let`, and to move the declaration of variables based on the new scope rules when necessary. This is part of the refactor updates mentioned in #378.


nicknisi:
I updated the commit and reverted the changes to the tests. I forgot that `let` and `const` don't work in all cases in Node 4. :-)

nicknisi:
Also, it might be worth looking into converting the tests to TS in another PR.

bin/typedoc
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env node
 
-var td = require('../dist/lib/cli.js');
+const td = require('../dist/lib/cli.js');
```
blakeembrey: Should we leave this one as `var` since it's JavaScript?

nicknisi: I went ahead and put the `var` back here. It should be fine, as that usage of `const` is ok in Node 4, but just to be consistent, it seemed better to put it back.

PR #430, opened 2017-02-21
Add TSLint to the project

nicknisi:
Add a tslint.json file to the project, and clean up the files in src/ to follow it.

This PR touches a lot of files in order to standardize things such as `===` usage single quotes. Additionally, I added the `noUnusedVariables` to tsconfig.json and removed all unused imports and interfaces across the project.

This is part of the refactor changes detailed in #378.


test/converter/access/specs.json
```diff
@@ -20,6 +20,7 @@
           "kind": 128,
           "kindString": "Class",
           "flags": {
+            "0": true,
```
jiayihu: Why there are added zeros?

jiayihu:
Wow great work! There are strange zeros appearing in `specs.json`, which may be the cause of travis failure.

Anyway I think we should discuss more about tslint rules. For example I personally prefer:

1. Trailing commas in objects, arrays and function calls
2. Keeping the type `property: string = 'defaultValue'` instead of `property = 'defaultValue'` because it assures I won't be tempted to change it to `property = 12` in future.

src/lib/converter/converter.ts
```diff
@@ -31,102 +30,49 @@ export interface IConverterResult
 
 
 /**
- * Event callback definition for generic converter events.
```
jiayihu: Was this part removed because of `unused vars`?

nicknisi: Yeah, these were removed because `tsc` now complains that they're not used anywhere.

test/converter/access/specs.json
nicknisi: I'll look into this and why the CI build failed.

nicknisi:
@jiayihu I tried to match the styles already present in the project when creating the config file to make it as easy as possible, with a few exceptions where I added in best practices with `===` and trailing commas, although I could go either way on the latter. I also had to pick one or the other with regard to quote mark, so I went with single quotes since I'm lazy and it's one less key to press.

In this PR I wanted to match as close as possible with what's there, but in addition to the two changes you mentioned, I would also like to possibly see the following:

* `"curly": true` - enforce curly braces for if/else/etc.
* `"interface-name": [ true, "never-prefix" ]` - Remove the need for `I` prefixes on interfaces, as that is no longer the recommended style in TypeScript.
* `"one-line": - Ensure that `{`, ` } catch`, and `} finally` are on the same line
* `"typedef-whitespace": [ true, ... ]`: I think it'd be nice to have whitespace between variable names and type definitions (for example: `const foo: string;`)

These changes, if agreed upon, could come in as a separate PR to further improve the code style.

aciccarello:
Great work! Haven't had a chance to review everything but I'm good with the tslint config. As for the config mentioned in the comments, I'd support what @nicknisi suggested. I'm also good with trailing commas in objects & arrays but I'd prefer to allow implicit `let property = 'defaultValue'` type definitions since it keeps things clean. Thanks @jiayihu for looking at everything.

nicknisi:
I found the issue that was causing the `0`s to show up in the specs and fixed it, and now the CI builds are passing!

src/lib/converter/plugins/CommentPlugin.ts
```diff
@@ -144,7 +144,7 @@ export class CommentPlugin extends ConverterComponent
         const comment = reflection.parent.comment;
         if (comment) {
             let tag = comment.getTag('typeparam', reflection.name);
-            if (!tag) tag = comment.getTag('param', '<' + reflection.name + '>');
+            if (!tag) tag = comment.getTag('param', `<${reflection.name}>`);
```
aciccarello: Not sure the template literal is any better here

src/lib/converter/plugins/GitHubPlugin.ts
```diff
@@ -1,20 +1,21 @@
...
+import {Converter} from '../converter';
+import {Context} from '../context';
 
 // This should be removed when @typings/shelljs typings are updated to the shelljs version being used
-declare module "shelljs" {
+declare module 'shelljs' {
     // `stdout` was added in:
     // https://github.com/shelljs/shelljs/commit/8a7f7ceec4d3a77a9309d935755675ac368b1eda#diff-c3bfabb5e6987aa21bc75ffd95a162d6
     // As of 2016-10-16, DefinitelyTyped's defs are for shelljs v0.3.0, but we're on 0.7.0
+    // tslint:disable-next-line:interface-name
```
aciccarello: Any idea why this interface was unprefixed?

src/lib/converter/plugins/CommentPlugin.ts
jiayihu: For consistency I personally use template literal anywhere there is interpolation

src/lib/converter/plugins/GitHubPlugin.ts
nicknisi: I'm not sure why this was unprefixed. ¯\_(ツ)_/¯

nicknisi:
@aciccarello I agree on the spacing between types, and there is a TSLint rule for it. I'll submit a separate PR.

src/lib/converter/plugins/GitHubPlugin.ts
blakeembrey: Possibly my fault when I started updating TypeDoc and needed to improve that definition. I'd just check the blame, but it can be made consistent. I'd love to have another pass through removing the prefixes and correcting whitespace and whatnot.

nicknisi:
Per @blakeembrey's comment, I went ahead with another passthrough to fix the whitespace rules and to remove the prefix `I` on interface names. Along with that, I renamed `IOptionDeclaration` to `DeclarationOption` so that the name doesn't conflict with the class `OptionDeclaration` in the same module, as they appear to be two different things.

jiayihu:
> to remove the prefix I on interface names

Are we all agree to this? I personally prefer `I` prefix to avoid confusion with classes.

nicknisi:
I personally don't like the `I` prefix, and not having them matches up with the TypeScript coding guidelines: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names

aciccarello:
I can go either way on interface prefixes. Following TS guidelines seems good to me. I haven't done anything with plugins.

Does anyone know if they use the interface names? If so I'd assume it would be a breaking change to rename the interfaces?

nicknisi:
I _think_ that because no interfaces are exported from `src/index.ts` that they aren't really part of the public API, but I could be wrong.

blakeembrey:
@aciccarello I'll be sure to release this as a minor just in case 👍

mootari:
All converter event descriptions still state
> The listener should implement [[IConverterCallback]].
> The listener should implement [[IConverterNodeCallback]].

I assume those references were left in by accident?

PR #433, opened 2017-02-23
Move '@types' dependencies to devDependencies

andy-hanson:
These are implementation details, so they don't need to be part of the dependencies.
Currently, since these are in `dependencies`, npm will install them to a user's `node_modules/@types` when it installs `typedoc`, and they will automatically be included in the user's compilation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,6 @@
     "node": ">= 4"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


blakeembrey:
Please search existing issues/PRs before creating a new one. See https://github.com/TypeStrong/typedoc/pull/384.

Issue #408, opened 2017-02-07
Installing typedoc via npm installs @types/node


dieseldjango:
Typedoc has a dependency on @types/fs-extra, which has a dependency on @types/node. Unfortunately, installing @types/node into node_modules in a react-native typescript project creates global type definition conflicts. This is preventing me from installing typedoc as a local package dependency.

blakeembrey:
Yeah, it's shitty but nothing this repo can or should fix. Just correct your `tsconfig.json` file and search existing issues first.

PR #434, opened 2017-02-24
Make @types packages dev dependencies.

hdeshev:
I hit a [typing clash](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14324) with @types/lodash and TypeScript 2.2.1 today.

While that is definitely a problem in the lodash typings, I realized that I wasn't supposed to even have `@types/lodash` in my project modules. I found out `typedoc` installs those, which is completely unnecessary.

This PR moves all `@types` packages to `devDependencies`. Probably related to #432.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,6 @@
     "node": ">= 4"
   },
   "dependencies": {
-    "@types/fs-extra": "0.0.33",
-    "@types/handlebars": "^4.0.31",
-    "@types/highlight.js": "^9.1.8",
-    "@types/lodash": "^4.14.37",
-    "@types/marked": "0.0.28",
-    "@types/minimatch": "^2.0.29",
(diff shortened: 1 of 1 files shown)


aciccarello:
@hdeshev thanks for the contribution but I don't believe this is the fix that is required.

As commented in [the issue you linked](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14324#issuecomment-276792877)

> I believe the problem lies with @types/lodash (for its mis-matching WeakMap interface). It's just that previous versions of TypeScript did not catch the error

We already have about half a dozen issues/PRs related to this issue which you can view for reference.

- [#310 (comment)](https://github.com/TypeStrong/typedoc/pull/310#issuecomment-254633782)
- #408 
- #384
- #350 
- #335

PR #436, opened 2017-02-25
feat(readers/tsconfig.ts): support "extends"

### LINKED ISSUE #380: Error with nested tsconfig extends
Hi!
Recent typescript supports configuration inheritance via "extends" in `tsconfig.json`. Everything's working until extended config tries to extend another itself.
Example:
```
//tsconfig.json
{
 "extends": "./tsconfig.a.json"
}

//tsconfig.a.json
{
 "extends": "./tsconfig.b.json",
}

//tsconfig.b.json
{
 "compilerOptions": {
 }
}
```

An error occures: `TypeError: Cannot convert undefined or null to object` at `typedoc\dist\lib\utils\options\readers\tsconfig.js:60` when trying to `delete compilerOptions[key];`.
Seems like `data.raw` does not contain `compilerOptions` at all.

```
ts: 2.1.5
typedoc: 0.5.5
```

StrahilKazlachev:
close #380


src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -26,6 +26,30 @@ export class TSConfigReader extends OptionsComponent
     private static OPTIONS_KEY:string = 'tsconfig';
 
 
+    private readConfigFile(event:DiscoverEvent, fileName:string) {
+        if (!FS.existsSync(fileName)) {
```
StrahilKazlachev: @blakeembrey Not sure if [this check](https://github.com/Microsoft/TypeScript/blob/cf72ceb0933e0297ed9546cad3ae9def9916a114/src/compiler/commandLineParser.ts#L920) is needed here, what do you think?

blakeembrey: Probably not if it's already handled. However, is it possible to skip most of this code and have `ts` read the entire file with `extends` instead? I thought there was a method for it.

StrahilKazlachev: As I get it we need the *raw* compiler options. `parseJsonConfigFileContent()` does follow the `extends` chain and aggregate the data but not in the *raw form*. If we use the `options: CompilerOptions` property from the result of `parseJsonConfigFileContent()` we will have to remove all internal options and map the enum represented ones(`ModuleKind`, `ScriptTarget`, `NewLineKind`, ...) to the correct string values.

blakeembrey: Do you know where we use the raw compiler options? My primary concern is that we may implement extends "wrong" if the compiler options ever change. If I've learnt anything from using the TypeScript compiler, it's that things can change at any time so implementing your own is usually bad.

StrahilKazlachev: I'm not very familiar with the project so here is how I've understood it so far:
1. [Here](https://github.com/TypeStrong/typedoc/blob/181a86d07a36170527ede0b6fbb870ade29d6086/src/lib/utils/options/readers/tsconfig.ts#L79) a typedoc intended only options may be present in the tsconfig.json, those options get merged with the raw `compilerOptions` from the tsconfig.json.
2. The conversion of the raw options to `CompilerOptions` and separation - TS/TypeDoc , is done [here](https://github.com/TypeStrong/typedoc/blob/181a86d07a36170527ede0b6fbb870ade29d6086/src/lib/utils/options/options.ts#L143) by [OptionDeclaration](https://github.com/TypeStrong/typedoc/blob/181a86d07a36170527ede0b6fbb870ade29d6086/src/lib/utils/options/declaration.ts#L76).

Why the conversion is done by typedoc and not TS I do not understand, I'm probably missing something or it is legacy. That's why I went with aggregating the raw options and not using the parsed ones returned from the `parseJsonConfigFileContent()` method used [here](https://github.com/TypeStrong/typedoc/blob/181a86d07a36170527ede0b6fbb870ade29d6086/src/lib/utils/options/readers/tsconfig.ts#L64) - it returns [ParsedCommandLine](https://github.com/Microsoft/TypeScript/blob/cf72ceb0933e0297ed9546cad3ae9def9916a114/src/compiler/types.ts#L3422) which has the parsed options - `options:CompilerOptions`.
If you have any suggestions please do tell.

blakeembrey: I feel like it would be legacy and we should be able to tidy it up instead of trying to build our own extends. However, if there's too much to correct, let me know.

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -51,31 +51,45 @@ export class TSConfigReader extends OptionsComponent {
...
+        let {options, fileNames} = ts.parseJsonConfigFileContent(
             data,
             ts.sys,
             Path.resolve(Path.dirname(fileName)),
             {},
             Path.resolve(fileName));
 
-        event.inputFiles = data.fileNames;
+        event.inputFiles = fileNames;
+        options = _.clone(options);
```
blakeembrey: Should we care about cloning the options here? The config parse function should be generating a plain object that wouldn't be used anywhere else anyway.

src/lib/utils/options/options.ts
```diff
@@ -72,6 +74,10 @@ export class Options extends ChildableComponent<Application, OptionsComponent> {
         this.trigger(event);
         this.setValues(event.data, '', event.addError.bind(event));
 
+        if (event.compilerOptions) {
+            _.assign(this.compilerOptions, event.compilerOptions);
```
blakeembrey: Should we use `event.data` for everything instead of having two ways to pass this data?

src/lib/utils/options/readers/tsconfig.ts
blakeembrey: Can you also point out where `extends` is now handled? I'm not sure I'm reading this snippet correctly, but it seems to move things around but I can't tell what/where something new was introduced.

StrahilKazlachev: `extends` is handled by `ts.readConfigFile()`.  The rest of the changes here are for preserving the precedence of the options(cli, ..., tsconfig.compilerOptions).

src/lib/utils/options/options.ts
StrahilKazlachev: It can be used, but currently `event.data` is more like `event.rawOptions` so whatever additional data is added it should not be in conflict with any of the options and should be deleted before parsing them.
If the following is OK I'll make the changes:
1. in `readers/tsconfig.ts` - add `parsedCompilerOptions: CompilerOptions` to the `event.data`
2. in `options/options.ts ` - delete `parsedCompilerOptions` from `event.data` before parsing

StrahilKazlachev:
@blakeembrey ping

PR #437, opened 2017-02-28
Convert unit tests to TypeScript

nicknisi:
The following changes are included in this Pull Request:

* Convert test/*.js to test/*.ts
* Add test files to tsconfig.json, compiling them to dist/test and src/ to dist/src/
* Updates path in bin/typedoc
* Add @types/mocha as a devDependency
* Fix some improper method signatures in src/lib/utils/events.ts using function overloading
* rename a `test` method in the generics example to `testFunction` and check in the associated built files
* add grunt-contrib-copy as a devDependency to copy the static test files to dist/test
* Add dist/src/ to "files" in package.json so that test files aren't published.


blakeembrey:
What do you think about 1. moving `test/` into `src/` to avoid structure changes or 2. using something like `ts-node` to avoid compilation altogether. I think it'd be better to go with 1 and just run tests from `dist/test` as you've done, but avoid the need to structure like `dist/src/lib`.

nicknisi:
I agree that the pathing isn't great. I'd like to propose something we've done on other projects:

1. During development / test -  build src/ and test/ into a _build/ directory.
1. For publishing - build src/ based off a tsconfig-release.json that extends the original and modifies it to build a dist/ directory with the pathing the way it currently is in this project.

This way, the actual release version pathing won't change, but we can still have separate src/ and test/ directories. What do you think?

blakeembrey:
Do we care about the separate `src/` and `test/` directories though? I know I don't. Personally, it's probably a little cleaner to put it all in `src/` since it's all TypeScript and requires compilation (nicer to have all the compilation assets be 1:1). I think I'd prefer that over the separate top-level directories, even if we do end up doing the `tsconfig-release.json` as well (which we definitely can, we'd just specify includes to ignore `src/test`).

nicknisi:
ok, I'll move test into src and give it a try. Thanks for the feedback!

nicknisi:
@blakeembrey I updated the PR and moved the `test/` directory into the `src/` directory, and updated all the pathing. Also, I had to get more specific in the `"files"` configuration in `package.json` because of a [known issue](https://github.com/npm/npm/wiki/Files-and-Ignores#known-issues) with npm where .npmignore doesn't take precedent over a directory listed in `"files"`.

tsconfig.json
```diff
@@ -14,7 +14,10 @@
     "experimentalDecorators": true
   },
   "include": [
-    "src/**/*.ts"
+    "src/lib/**/*.ts",
+    "src/test/*.ts"
```
blakeembrey: Shouldn't this stay `src/**/*.ts`?

package.json
```diff
@@ -63,12 +65,15 @@
   },
   "files": [
     "bin",
-    "dist",
+    "dist/lib",
+	"dist/index.d.ts",
+	"dist/index.js",
+	"dist/index.js.map",
```
blakeembrey: Couldn't we just keep it as `dist/` instead? I think that'd be better to avoid anyone running into an issue in the future. At the very least, if what you're trying to do is ignore `dist/test/`, we _could_ do the `tsconfig.release.json` idea.

gruntfile.js
```diff
@@ -64,9 +64,20 @@ module.exports = function(grunt)
                 }
             }
         },
+        copy:  {
+            staticTestFiles: {
+                expand: true,
+                cwd: 'src',
+                src: [
+                    'test/converter/**/*',
+                    'test/renderer/**/*'
```
blakeembrey: Do we need to fix this path now?

gruntfile.js
```diff
@@ -34,7 +34,7 @@ module.exports = function(grunt)
                 configuration: 'tslint.json'
             },
             files: {
-                src: [ 'src/**/*.ts' ]
+                src: [ 'src/lib/**/*.ts', 'src/test/*.ts' ]
```
blakeembrey: Can this stay as `src/**/*.ts`?

gruntfile.js
nicknisi: The `cwd` on this is set to 'src', so the paths should be correct.

nicknisi:
@blakeembrey Sorry, yeah those paths were definitely incorrect. I fixed them. Also, I fixed the `"files"` issue by explicitly excluding `dist/test`. I tested this setup with `npm pack` to make sure that `dist/test` was not included.

PR #439, opened 2017-03-03
Decouple converter for stand-alone use

bryanforbes:
This is part of the work for #378. Right now, `lib/converter/converter.ts` uses a "component" architecture via decorators to register handlers to convert from the TypeScript compiler's representation of the code to typedoc's reflections. This architecture is very difficult to follow and know for sure how everything works together and directly couples the converter to the `Application` and `Options` classes. This pull request simplifies the converter by removing the concept of "components", eschews the use of decorators, and changes the constructor to accept an options object.


blakeembrey:
I would love to merge this if we can get an additional set of eyes. Aside from that, my only question is - do you want to delete that commented out code instead?

bryanforbes:
Not to put a damper on this, but I made some discoveries in the TypeScript APIs today and I'm looking into a way to simplify the converter more than what I've done here. I should have a proof of concept tomorrow.

dupski:
Was there ever any update on this @bryanforbes ? Would be great to get it merged, TypeDoc is awesome, but is real hard to hack on currently, partially due to this stuff :)

PR #441, opened 2017-03-07
Expose src for direct api usage

asakusuma:
Allow consumers to directly use TypeDoc's internal TypeScript objects (beyond the ones exposed in `index.d.ts`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "tslint": "^4.4.2"
   },
   "files": [
+    "src",
     "bin",
     "dist",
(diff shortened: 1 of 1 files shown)


nicknisi:
Shouldn't this already be possible since the `.d.ts` files for each module are already published within `dist`?

asakusuma:
Ah, you're right, I was doing something else wrong that made me assume the compiler was resolving the `.js` file instead of the `.d.ts` one. Thanks, closing.

Issue #344, opened 2016-11-23
TS 2.2dev doesn't output anything... :(


electricessence:
I've tried with different versions and recently pulled directly from GitHub as Daniel R on Twitter posted an article about.

Firstly, with 2.0.2, and 2.0.8 I get a Node runtime error something about ```...declarations.forEach``` and it won't continue.

2.2dev seems to work just fine.  But no API shows up.  This used to work fine before.  Is there maybe a source parameter I'm missing?

aciccarello:
Hello @electricessence, [support for Typescript 2.x](https://github.com/TypeStrong/typedoc/issues/234) is not at 100% yet. Members of the community have been submitting PRs to support newer features but more work needs to be done to support it completely.

Any further information to help the community figure out a fix would be greatly appreciated!

electricessence:
@aciccarello The article I read presented a means for using 2.x so I thought it would be a good idea to double-check/verify if that's an actual thing.

aciccarello:
@electricessence I'm going to assume that this was fixed in #448. Feel free to reopen if you are still encountering this problem.

PR #448, opened 2017-03-14
Update to TypeScript 2.2.1

devpaul:
* updated node engine to 4.2.0+ to match TypeScript's
* optionDeclarations.type can now be a `Map` added support to `declaration`
* tsconfig to include es2015 collections lib

Will resolve #432, #414, and #417; maybe #344

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -101,8 +101,11 @@ export class OptionDeclaration {
                     } else if (key in map) {
                         value = map[key];
                     } else if (errorCallback) {
-                        const message = this.mapError ? [this.mapError] : ['Invalid value for option "%s".', this.name];
-                        errorCallback(... message);
+                        if (this.mapError) {
+                            errorCallback(this.mapError);
+                        } else {
+                            errorCallback('Invalid value for option "%s".', this.name);
+                        }
                     }
                 }
                 break;
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -46,7 +46,7 @@ export class OptionDeclaration {
 
     scope: ParameterScope;
 
-    map: Object;
+    protected map: Object | Map<string, string> | 'object';
 
     mapError: string;
 
@@ -92,11 +92,11 @@ export class OptionDeclaration {
                 }
                 break;
             case ParameterType.Map:
-                if (this.map !== 'object') {
-                    const key = value ? (value + '').toLowerCase() : '';
-                    if (key in this.map) {
-                        value = this.map[key];
-                    } else if (errorCallback) {
+                const key = value ? (value + '').toLowerCase() : '';
+                try {
+                    this.getMapValue(key);
+                } catch (error) {
+                    if (errorCallback) {
                         if (this.mapError) {
                             errorCallback(this.mapError);
                         } else {
@@ -108,5 +108,21 @@ export class OptionDeclaration {
         }
 
         return value;
+    }
+
+    protected getMapValue(key: string): any {
+        const map = this.map;
+
+        if (map === 'object') {
+            return undefined;
+        } else if (map instanceof Map) {
+            if (map.has(key)) {
+                return map.get(key);
+            }
+        } else if (key in map) {
+            return map[key];
+        }
+
+        throw new Error('Invalid value for option');
     }
 }
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -46,7 +46,7 @@ export class OptionDeclaration {
 
     scope: ParameterScope;
 
-    protected map: Object | Map<string, string> | 'object';
+    protected map: Object | Map<string, any> | 'object';
 
     mapError: string;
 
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -1,6 +1,12 @@
 {
   "compilerOptions": {
     "module": "commonjs",
+    "lib": [
+      "DOM",
+      "ES5",
+      "ScriptHost",
```
blakeembrey: What is `ScriptHost` used for?

src/lib/utils/options/declaration.ts
```diff
@@ -46,7 +46,7 @@ export class OptionDeclaration {
 
     scope: ParameterScope;
 
-    map: Object;
+    protected map: Object | Map<string, string> | 'object';
```
blakeembrey: Can we avoid the `protected` change?

src/lib/utils/options/declaration.ts
```diff
@@ -92,11 +92,11 @@ export class OptionDeclaration {
...
             case ParameterType.Map:
-                if (this.map !== 'object') {
-                    const key = value ? (value + '').toLowerCase() : '';
-                    if (key in this.map) {
-                        value = this.map[key];
-                    } else if (errorCallback) {
+                const key = value ? (value + '').toLowerCase() : '';
+                try {
+                    this.getMapValue(key);
+                } catch (error) {
```
blakeembrey: Can we please avoid using `try..catch` statements here. It seems like we should be able to keep most of the code the same as it was instead of changing all this.

blakeembrey: I don't actually see where you'd updating `value =` anymore.

tsconfig.json
devpaul: It's a standard include w/ an ES5 target. It's possible we don't need it.

https://github.com/Microsoft/TypeScript/blob/master/lib/lib.scripthost.d.ts

src/lib/utils/options/declaration.ts
devpaul: Sure. I can make this change.

src/lib/utils/options/declaration.ts
devpaul: Limiting the scope made it easier to understand where `map` is being used. Is there a reason to keep this value public?

tsconfig.json
blakeembrey: No it's not. That looks like IE-specific extensions and definitely isn't ES spec.

src/lib/utils/options/declaration.ts
blakeembrey: Yes - if someone is using it, you just broke them.

tsconfig.json
devpaul: Yes. It is mostly IE-specific extensions. It's not part of the ES spec. It is included when `--target: ES5` (and ES6) is used. See https://www.typescriptlang.org/docs/handbook/compiler-options.html under the `--lib` row.

> Note: If --lib is not specified a default library is injected. The default library injected is: 
► For --target ES5: DOM,ES5,ScriptHost

Anyway, I removed `ScriptHost` locally and it compiled. Ambient declarations from definitely typed don't typically reference it, so we should be OK removing it.

src/lib/utils/options/declaration.ts
devpaul: `map` is really an internal implementation detail of what OptionDeclaration does and shouldn't be public.  If `map` remains `public` then anyone who relies on `map` will silently fail; because `Map` is also an `Object` TypeScript will not throw an error about incompatible types and `map` will look like an empty object. Changing `map` to `protected` ensures anyone who wants to continue to use `map` will need to extend this class and be aware of the change. Overall, we are better off making this breaking change and preventing silent failures for users.

tsconfig.json
blakeembrey: What references is irrelevant here - we shouldn't be including any libraries we don't have. We shouldn't even include `dom` since we're not in the DOM. All that does is give a false sense of security and errors at runtime instead.

nicknisi: I think that if `lib` is not specified in tsconfig.json, then `dom` is part of the default libs brought in per https://www.typescriptlang.org/docs/handbook/compiler-options.html and DOM is necessary in this project because highlight.js is a dependency.

devpaul: Removing `DOM` fails the build:

```
Running "ts:typedoc" (ts) task
Compiling...
Using tsc v2.2.1
node_modules/@types/highlight.js/index.d.ts(19,40): error TS2304: Cannot find name 'Node'.
```

blakeembrey: Yeah, that's a known issue with highlight.js - I believe we've mentioned it somewhere - maybe wherever we moved to `@types`. Fixing that is sadly contingent on fixing the definition in DefinitelyTyped or enable TypeScript to understand `browser` resolution/do conditional definition paths (since there's no `Node` constructor in node).

src/lib/utils/options/declaration.ts
```diff
@@ -92,16 +92,17 @@ export class OptionDeclaration {
...
+                    } else if (key in map) {
+                        value = map[key];
                     } else if (errorCallback) {
-                        if (this.mapError) {
-                            errorCallback(this.mapError);
-                        } else {
-                            errorCallback('Invalid value for option "%s".', this.name);
-                        }
+                        const message = this.mapError ? [this.mapError] : ['Invalid value for option "%s".', this.name];
+                        errorCallback(... message);
```
blakeembrey: Can we keep this code the same as it was before, I don't think there's any need to be clever here.

PR #449, opened 2017-03-15
remove exclusion of TS 2 options, add lib support

### LINKED ISSUE #315: TypeDoc doesn't support the lib option
I use the new `lib` option in `tsconfig.json` to explicitly set the [built-in type definitions](http://www.typescriptlang.org/docs/release-notes/typescript-2.0.html#including-built-in-type-declarations-with---lib) I want to use in my project:

``` json
{
    "$schema": "http://json.schemastore.org/tsconfig",
    "compilerOptions": {
        "target": "es5",
        "lib": [ "es2015.promise", "es2015.core", "es2015.collection", "es5", "dom" ],
        "module": "commonjs",
        "rootDir": "src",
        "outDir": "lib",
        "noEmitOnError": true,
        "sourceMap": true,
        "declaration": true,
    },
    "exclude": [
        "node_modules",
        "lib"
    ]
}
```

Unfortunately typedoc doesn't support this:

``` bash
$ typedoc --out doc src
Error: Unknown option `lib.0`.
Error: Unknown option `lib.1`.
Error: Unknown option `lib.2`.
Error: Unknown option `lib.3`.
Error: Unknown option `lib.4`.
```

For now I reverted back to using external type definitions for the needed ES2015 features so I'm able to use TypeDoc but it would be nice if TypeDoc could support this useful new feature.

nicknisi:
* Remove restriction on what compilerOptions can be set
* remove filtering of TS 2.2 compilerOptions
* add support for 'list' option types
* Switch to processed compilerOptions
* add support for `compilerOptions.lib`

This PR removes the restrictions on setting options that aren't explicitly setup as options to allow new compilerOptions to pass through TypeDoc to TypeScript. It also sets up support for 'list' parameter types, and switches to use the processed compilerOptions, rather than `raw` so that the `lib` properties are properly parsed.

This should fix #315 and I believe #311.

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -95,12 +95,13 @@ export class OptionDeclaration {
                 const map = this.map;
                 if (map !== 'object') {
                     const key = value ? (value + '').toLowerCase() : '';
+                    const values = Object.keys(map).map(key => map[key]);
 
                     if (map instanceof Map && map.has(key)) {
                         value = map.get(key);
                     } else if (key in map) {
                         value = map[key];
-                    } else if (errorCallback) {
+                    } else if (!~values.indexOf(value) && errorCallback) {
                         if (this.mapError) {
                             errorCallback(this.mapError);
                         } else {
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -101,7 +101,7 @@ export class OptionDeclaration {
                         value = map.get(key);
                     } else if (key in map) {
                         value = map[key];
-                    } else if (!~values.indexOf(value) && errorCallback) {
+                    } else if (values.indexOf(value) === -1 && errorCallback) {
                         if (this.mapError) {
                             errorCallback(this.mapError);
                         } else {
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -97,8 +97,8 @@ export class OptionDeclaration {
                     const key = value ? (value + '').toLowerCase() : '';
                     const values = Object.keys(map).map(key => map[key]);
 
-                    if (map instanceof Map && map.has(key)) {
-                        value = map.get(key);
+                    if (map instanceof Map) {
+                        value = map.has(key) ? map.get(key) : key;
                     } else if (key in map) {
                         value = map[key];
                     } else if (values.indexOf(value) === -1 && errorCallback) {
--- a/src/lib/utils/options/options.ts
+++ b/src/lib/utils/options/options.ts
@@ -152,10 +152,10 @@ export class Options extends ChildableComponent<Application, OptionsComponent> {
             const value = obj[key];
             const declaration = this.getDeclaration(key);
             const shouldValueBeAnObject = declaration && declaration['map'] === 'object';
-            if (typeof value === 'object' && !shouldValueBeAnObject) {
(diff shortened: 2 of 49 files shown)


src/lib/utils/options/declaration.ts
```diff
@@ -95,12 +95,13 @@ export class OptionDeclaration {
...
                 if (map !== 'object') {
                     const key = value ? (value + '').toLowerCase() : '';
+                    const values = Object.keys(map).map(key => map[key]);
 
                     if (map instanceof Map && map.has(key)) {
                         value = map.get(key);
                     } else if (key in map) {
                         value = map[key];
-                    } else if (errorCallback) {
+                    } else if (!~values.indexOf(value) && errorCallback) {
```
blakeembrey: Can we use the non-clever format of `indexOf`, cheers! 😄

nicknisi: Removed! Sorry about that!

src/lib/utils/options/options.ts
```diff
@@ -152,10 +152,10 @@ export class Options extends ChildableComponent<Application, OptionsComponent> {
             const value = obj[key];
             const declaration = this.getDeclaration(key);
             const shouldValueBeAnObject = declaration && declaration['map'] === 'object';
-            if (typeof value === 'object' && !shouldValueBeAnObject) {
+            if (!Array.isArray(value) && typeof value === 'object' && !shouldValueBeAnObject) {
                 this.setValues(value, prefix + key + '.', errorCallback);
             } else {
-                this.setValue(prefix + key, value, errorCallback);
+                this.setValue(prefix + key, value);
```
blakeembrey: Was there a reason behind this signature change?

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -51,31 +51,30 @@ export class TSConfigReader extends OptionsComponent {
             return;
         }
 
-        let data = ts.readConfigFile(fileName, ts.sys.readFile).config;
-        if (data === undefined) {
+        let { config } = ts.readConfigFile(fileName, ts.sys.readFile);
```
blakeembrey: Could we use `const` if there's no other mutation?

blakeembrey:
Sorry for the delay, I just had a couple of last comments. Once those are reviewed, I'm happy to merge!

nicknisi:
@blakeembrey Thanks! I rebased off the latest master and pushed up two commits to address your comments. Regarding the signature change, I actually modified my approach now that `OptionDeclaration#map` is a `Map`. The big thing I changed with this code (and the previous signature change) was that I am trying to let unknown-to-TypeDoc options simply pass through without throwing an error. That way, TypeDoc doesn't need to maintain a whitelist of `compilerOptions`, and should be more future proof.

Also, I checked in a number of changes to the built files (in a separate commit). These changes aren't related to this work, and appear when running the tests, even on master. I can remove this commit if you like.

blakeembrey:
Thanks @nicknisi! ❤️ 

I'm going to make a minor edit since it looks like you accidentally inserted a tab where there's spaces, then I'll merge 👍

blakeembrey:
Weird. I just pulled this in and tested, but I end up reverting all the `.html` changes you recently made. Do you have something locally affecting that?

nicknisi:
Yep, you're totally right... I didn't realize I had the themes repo `npm link`'d in!

PR #465, opened 2017-04-02
Fix separator in getDefaultLibFileName

### LINKED ISSUE #464: Searching for lib files in root directory
I am running the latest (0.5.9) version of typedoc. Node version 7.8.0

Output: 
```
> typedoc --out docs --mode modules src

Using TypeScript 2.2.1 from C:\...\node_modules\typescript\lib
Error: File 'C:/lib.dom.d.ts' not found.
Error: File 'C:/lib.es2016.d.ts' not found.
```

My `tsconfig.json` file is the following: 
```json
{
    "compilerOptions": {
        "target": "es6",
        "moduleResolution": "node",
        "module": "commonjs",
        "lib": ["es2016", "dom"],
        "noImplicitAny": true,
        "removeComments": false,
        "strictNullChecks": true
    }
}
```

Please let me know if any further information is needed, for now I've just swapped `arr.includes` for `!!~arr.indexOf`, though I'd much rather use the readable alternative.

nicknisi:
Remove use of `Path.join` in the `getDefaultLibFileName` method that, on Windows, converts '/' in paths to '\\'. TypeScript uses '/' internally so it has trouble locating the default lib location when this is reversed. Fixes #464 where lib directory cannot be found on Windows.

In this PR, I changed the `getDefaultLibFileName` method in compiler-host.ts to more resemble that which would be returned from `ts.getCompilerHost`, which also included exposing a few more methods in ts-internal.ts.

### DIFF
--- a/src/lib/converter/utils/compiler-host.ts
+++ b/src/lib/converter/utils/compiler-host.ts
@@ -1,9 +1,7 @@
 import * as ts from 'typescript';
 import * as _ts from '../../ts-internal';
-import * as Path from 'path';
 
 import {ConverterComponent} from '../components';
-import {normalizePath} from '../../utils/fs';
 
 /**
  * Return code of ts.sys.readFile when the file encoding is unsupported.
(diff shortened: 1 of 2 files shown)


aciccarello:
Thanks @nicknisi. Is there any way we can avoid being tightly dependent on those internal typescript methods? See #310 for related discussion

PR #473, opened 2017-04-08
Move @types into devDependencies

orta:
Hey there, these `@types/x` dependencies are being shipped to all library consumers of typedoc. I can understand why you would want to include them in dependencies (so they match) but simply including typedoc in my project makes tsc fail.

If you want to make this a systemic thing, in Danger [we check](https://github.com/danger/danger-js/blob/master/dangerfile.ts#L88-L109) to make sure no-one ships `@types` inside the dependencies section of the package in each PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,6 @@
     "node": ">= 4.2.0"
   },
   "dependencies": {
-    "@types/fs-extra": "0.0.33",
-    "@types/handlebars": "^4.0.31",
-    "@types/highlight.js": "^9.1.8",
-    "@types/lodash": "^4.14.37",
-    "@types/marked": "0.0.28",
-    "@types/minimatch": "^2.0.29",
-    "@types/shelljs": "^0.3.32",
     "fs-extra": "^2.0.0",
     "handlebars": "4.0.5",
     "highlight.js": "^9.0.0",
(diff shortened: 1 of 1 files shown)


blakeembrey:
Please see the half dozen previous PRs and issues on this before opening a new one.

orta:
Ah, that's a shame. Sorry about that, might be worth adding a PR template and including that if it's a pretty common mis-conception?

For anyone that does the same thing, I've made a quick fork they can use instead: `yarn add typedoc@orta/typedoc#0.5.10-no-types` 👍

blakeembrey:
Sure, just no one will be able to use that package in TypeScript without going through the dependency list and installing the types manually. If there's something breaking you, we should fix it properly instead of hacking around the issue. There's several issues I've pointed out, but no one has worked on them yet - mostly just moving DefinitelyTyped definitions we use to be proper modules and (more complex) on TypeScript's side properly support browser vs node versions of definitions.

orta:
Ah yeah, I should have added that caveat, my focus was to get the `typedoc` command to work locally.

It doesn't seem like a critical issue, considering typedoc can be installed globally and it's got a solid CLI option set.

blakeembrey:
Definitely agree, but I know there's at least a couple of people using it programmatically. No where near how many would be using the CLI, but I'd just like to see things fixed properly. I know we can all hack around the issues, but fixing it properly would help the project long-term and also help us understand where `@types` stands today and what needs to be fixed with it going forward.

PR #475, opened 2017-04-10
Change the pattern of --exclude option to Regular Expression

JohnKoutsoumpas:
The minimatch glob pattern does not properly support multiple exclusions other than a simple (this|that) which is not enough if you need to exclude multiple directories *and* subdirectories. A good solution to this is to change from a minimatch glob pattern to a Regular Expression pattern which is more powerfull and can almost match anything.

Please consider accepting this as it doesn't affect anything other than this file matching and makes it considerably more powerful.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -9,7 +9,6 @@
 import * as Path from 'path';
 import * as FS from 'fs';
 import * as typescript from 'typescript';
-import {Minimatch, IMinimatch} from 'minimatch';
 
 import {Converter} from './converter/index';
 import {Renderer} from './output/renderer';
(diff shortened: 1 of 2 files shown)


blakeembrey:
It does support it, you can do something like `{this,that/bar/**}` I believe. Either way, I'm 👎  on this PR as it'll break every current user using this feature and minimatch patterns as well-used and understood for CLI tools.

JohnKoutsoumpas:
That's fine. If the globbing does support it then that is great but I couldn't manage to make it work and I've spent hours trying. I'll try it with your suggestion.

JohnKoutsoumpas:
Success! Your suggestion worked well, thanks. I'll just close this now.

PR #478, opened 2017-04-12
Small enhancements

lenovouser:
- Update dependencies
 - Update copyright date


blakeembrey:
Unfortunately I can't accept these changes. Please do a quick search for existing `@types` issues and PRs for background info. I'm also not comfortable changing the copyright information - Sebastian hasn't been involved with the project for a long time now.

lenovouser:
@blakeembrey what do you think about

```diff
Copyright (c) 2015 [Sebastian Lenz](http://typedoc.org).<br>
+Copyright (c) 2016-2017 [TypeDoc Contributors](https://github.com/TypeStrong/typedoc/graphs/contributors).<br>
```
for the copyright date? I can remove it too.

I did some quick research and upgrading `handlebars` is no problem, only minor semver change anyways. `progress` is a major semver change but the API stayed the same (https://github.com/visionmedia/node-progress/commit/d84326ed9ab7720592b6bbc9c108849cd2a79908)

`@types/fs-extra` is the biggest change here, but it actually just aligns the types to the npm version of the actual `fs-extra` dependency you use in the code.

Both `tslint` and `grunt-tslint` should be fine too, I don't see any changes in [tslint@5.0.0](https://github.com/palantir/tslint/releases/tag/5.0.0) that break anything here.

The build, linter and tests all pass or succeed and I tested that it generates documentation correctly and without errors on some of my own projects.

blakeembrey:
Sure, I'm good with that. I'm always not a lawyer, but looks good regardless 😄 

I'm happy with all the dependency updates, just not moving those dependencies into `devDependencies` or otherwise. If there's an issue with them being in dependencies, we should fix whatever that is directly.

lenovouser:
Ah, okay. Well. When you install the package globally or in your dependencies all it's dependencies will get installed too (devDependencies not) as far as I know. That's why you usually put these in the devDependencies because they are just needed for development. Or am I missing something?

blakeembrey:
They are needed for people who use the project programmatically from TypeScript.

Issue #146, opened 2015-09-22
BUG: Type Typo in Generated JSON


EisenbergEffect:
Throughout the generated json we see this a lot: `"type": "instrinct"`. What is that? It seems like a typo. It would be great to see that fixed.

(PS - Aurelia.io is using TypeDoc on all it's libraries now..and we've also built a docs viewer on top of the TSDoc json format. :))

sebastian-lenz:
Yes, this looks like a typo, should be <code>"intrinsic"</code> I guess. Sorry for my bad english.

sebastian-lenz:
I'm unsure whether to fix that one in the current branch as it would break everyones code using the json export. v0.4 is in the making and I think it would be better to correct the typo with this new version.

EisenbergEffect:
@sebastian-lenz Thanks for the update. Do you have any timeline on 0.4 or any feature roadmap on that. I'm just curios.

PR #482, opened 2017-04-23
Support for intersection type

makana:
This PR introduces the support for intersection types (Microsoft/Typescript#3622).

I also added the type name identifiers ('void', 'reference', 'union' etc.) as a property to the type model to access them within themes (see TypeStrong/typedoc-default-themes#45).


src/lib/models/types/intrinsic.ts
```diff
@@ -14,6 +14,11 @@ export class IntrinsicType extends Type {
     name: string;
 
     /**
+     * The type name identifier.
+     */
+    readonly type: string = 'instrinct'; // TODO: Is there any change to correct this typo?
```
blakeembrey: What do you think we should do here? I'm happy to accept the fix in a future PR. Just let me know what you think is needed to fix this!

makana:
It just continues the discussion on #146.

There's the argument not do change anything before 0.4 as it will break everyones code that relies on typedoc.

Now we have 0.5.x and just my two cents is, we should correct this (= every occurrence of "instrinct") to "intrinsic" - either with 0.6 or just now. I think it is not to late. Typedoc is in a relatively early state and everyone should be able to handle the "breaking change" with a simple replace-in-all-files action. Certainly it will be better than living with this pretty ugly typo for the rest of our lives ;)

blakeembrey:
I'm definitely happy to have it fixed and I would just bump the minor and document the occurance. Thanks.

PR #484, opened 2017-04-25
Fixed the intrinsic type name identifier typo from "instrinct" to "intrinsic"

### LINKED ISSUE #146: BUG: Type Typo in Generated JSON

makana:
This PR fixes the type name identifier of the "intrinsic" type which was "instrinct" before.

Changed the type model, test specs were updated automatically.

As this is a breaking change, the minor version was incremented.
**So as of 0.6.0 if you rely on the output of typedoc using the --json option, please check your code for the occurrence of "instrinct"!**

closes #146


blakeembrey:
Please don't change the version number. I can merge any fixes when you undo it.

makana:
Sorry about that. I was unsure about your comment in #482

blakeembrey:
No problem. You generally avoid bumping versions, especially as part of a feature, because people normally have an existing release process and tagging to do. I just meant I would bump the minor version when we release this 😄

PR #491, opened 2017-05-01
Support for multidimensional arrays

makana:
This PR adds the support for multidimensional arrays.

Therefore the isArray flag was removed in favor of a new array type model.

This is necessary because setting the isArray flag of a type that is already an array - which in fact happened for multidimensional arrays - has no effect.
```ts
convertNode(context: Context, node: ts.ArrayTypeNode): Type {
        let result = this.owner.convertType(context, node.elementType);

        if (result) { // What if isArray is already true here?
            result.isArray = true;
        } else {
            result = new IntrinsicType('Array');
        }

        return result;
}
```

Please consider the corresponding PR for themes: TypeStrong/typedoc-default-themes#48

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "minimatch": "^3.0.0",
     "progress": "^2.0.0",
     "shelljs": "^0.7.0",
-    "typedoc-default-themes": "^0.4.2",
+    "typedoc-default-themes": "^0.4.4",
     "typescript": "2.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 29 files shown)


blakeembrey:
This seems like a useful change to me. Does anyone else want to review it?

Should we merge and release the theme changes as a new minor first, then update this PR and release a minor of TypeDoc?

makana:
Any reactions on this one? I'll be definitely happy to get a second and third opinion.

However you handle the two repositories, both changes are not compatible without each other. So the way you mentioned sounds like a proper way to go.

PR #496, opened 2017-05-05
"exclude" option is not taken into account when passed through constructor

jbgraug:
The "exclude" option is not taken into account when passed through constructor.
This is important when using typedoc through the API. For instance when typedoc is used together with gulp-typedoc.


jbgraug:
There is no need for this fix. I was debugging and for some reason 'exclude' was undefined.

Issue #382, opened 2017-01-13
theme.js is broken


ajsantander:
theme.js is not being called anymore, this is a feature that is currently not working:
https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/theme.ts#L47

Some themes require theme.js to be executed for proper functioning, and there are several themes out there than rely on this.

Eg.
https://github.com/kimamula/typedoc-markdown-theme
https://github.com/alg/typedoc-dash-theme

I can verify that theme.js was being used in typedoc v0.3.12 but not anymore shortly after that.

Is this feature still available?

thepalebluedot:
For anyone interested, I found a workaround by using the typedoc API and warpping it in a node script: 
https://github.com/thepalebluedot/typedoc-md-theme

cpriest:
@ajsantander  The link you have posted above is no longer any good, do you still have the workaround?  I'm also running into this issue.

aciccarello:
Looks like this was caused by #252. PRs are welcome.

blakeembrey:
@aciccarello That PR shouldn't have changed the behaviour, it was already commented out and both code paths are identical as a result. It seems this behaviour was changed before (maybe before I was involved) so I can't be sure why/when it was changed. Feel free to link me if there's something I've missed though.

aciccarello:
@blakeembrey I believe you are correct. It appears that it was caused by prior changes that were the issue but that code was [commented out more than a year and a half ago](https://github.com/TypeStrong/typedoc/commit/fa7ebeb9ba3792a7083351a5a717e500d37f85d6#diff-63fbf19d888caafafd72e21f29cb1f4d). I dug up #246  and #237 which were prior to #252. But regardless of when it was removed we should align the features and the documentation.

blakeembrey:
Thanks for the investigation work 😄 Is this the best way to enable the needed features or is there something people would prefer? I'm not really familiar with it, but since it was removed so long ago if there's a better fix maybe we can do that.

aciccarello:
I'm trying to get a handle on the motivation for theme.js. Looking at the related issues:

- #246 found a work around by configuring handlebars
- @cpriest (per gitter conversation) is looking to change the results of `getNavigation()`
- #237 was focused on generating documentation in a single page (not sure how that's different than the minimal theme)

bbugh:
This is still a problem. My `theme.js` is not being loaded. Is there some other way to add JavaScript and/or Handlebars helpers?

[...]

Issue #227, opened 2016-06-07
Resource not found "reflection.hbs"


shaddockh:
In the latest version (```0.4.1```), I'm getting an error trying to locate ```reflection.hbs```.  I'm using a custom theme, but only overriding a couple .hbs files.  My understanding is that any files not detected in the custom theme folder would be loaded in from the default theme.

Backing down to ```0.3.12``` resolves the error.

kamranayub:
Yeah, can repro this and I cloned the current master of typedoc-default-themes. Happens when I try to pass in through `--theme`.

lisss:
Any updates on this? It's still an issue when trying to use `minimal` theme - the reflection.hbs is missing from templates directory for this theme (although present for default)

blakeembrey:
Could someone in this thread please test https://github.com/TypeStrong/typedoc/pull/252? I've not used TypeDoc with a custom theme, nor can I see anywhere that this functionality has possibly been removed. 

The issue @kamranayub alludes to should be unrelated, since yes, the files are no longer checked in but no, they still exist if you do an NPM install. This is because checked in built assets are really hard to maintain, just looking through the commit log to find out how this feature could have disappeared is near impossible.

shaddockh:
I just pulled down your branch and tried it out and it appears to still be a problem.
```

Using TypeScript 1.8.10 from /Users/shaddockh/Development/Atomic/AtomicMaster/AtomicGameEngine/Artifacts/Build/JSDoc/node_modules/typescript/lib
/Users/shaddockh/Development/Atomic/AtomicMaster/AtomicGameEngine/Artifacts/Build/JSDoc/node_modules/typedoc/lib/output/utils/resources/stack.js:98
        throw new Error(Util.format("Cannot find resource `%s`.", name));
        ^

Error: Cannot find resource `reflection.hbs`.
    at TemplateStack.ResourceStack.getResource (/Users/shaddockh/Development/Atomic/AtomicMaster/AtomicGameEngine/Artifacts/Build/JSDoc/node_modules/typedoc/lib/output/utils/resources/stack.j
s:98:15)

    at Renderer.renderDocument (/Users/shaddockh/Development/Atomic/AtomicMaster/AtomicGameEngine/Artifacts/Build/JSDoc/node_modules/typedoc/lib/output/renderer.js:57:73)
[...]

PR #502, opened 2017-05-14
Fix theme.js in custom themes

cpriest:
I believe this (along with the original PR#252) fully corrects the original issue #227 as well as the regression #382.

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -164,7 +164,7 @@ export function parseComment(text: string, comment: Comment = new Comment()): Co
         line = line.replace(/^\s*\*? ?/, '');
         line = line.replace(/\s*$/, '');
 
-        const tag = /^@(\w+)/.exec(line);
+        const tag = /^@(\S+)/.exec(line);
         if (tag) {
             readTagLine(line, tag);
         } else {
--- a/src/lib/output/renderer.ts
+++ b/src/lib/output/renderer.ts
@@ -201,7 +201,17 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
                 }
             }
 
-            this.theme = this.addComponent('theme', new DefaultTheme(this, path));
+            let filename = Path.join(path, 'theme.js');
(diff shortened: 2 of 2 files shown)


blakeembrey:
Can you explain how this fixes the issue? In the linked PR I did, the code doing this sort of behaviour was already commented out and this change is not consistent with that commented out behaviour.

cpriest:
Sure, I don't have much historical knowledge of the project but I believe this part of your commit:

```
+        this.addDirectory('default', Renderer.getDefaultTheme());
+        this.addDirectory('theme', theme.basePath);
```

Fixed the 'fallback' part which may not have ever been working properly.  I essentially took the other part of your commit, re-instituted the code and fixed it so that it would work properly.

cpriest:
Shoot I did not mean for this to come into this same PR.  LMK what you'd like me to do here...

Close this one and submit them separately?  I'm not sure there is a way to pull this commit out of the PR, if you know of a way, let me know and I will.

blakeembrey:
You can use rebase on the branch and fix it up. 

I'd also prefer that you use `const` over `let` for the PR, so feel free to also fix that up. I'd also say that class names should always use `PascalCase`, so that change might be useful. I'm not 100% on using `default` always - maybe we should check `default` before using it or use the old behaviour of `eval`?

cpriest:
I ran the tests on my local machine and they all passed, I think Travis was running slow or something.

cpriest:
@blakeembrey I think that may be why it was commented out as it didn't work with eval when tested against the typedoc-theme-markdown.  I based the fix on export default which is what the markdown theme had.  

If you didn't want to use the default, we could make it the same as for plugins.  The require() should pass back a function which accepts the app.  Thoughts?

I'll fix them up and re-submit.

blakeembrey:
Sure, but we can make them both consistent by including a check like `typeof x=== 'function' ? x : x.default`. I think that'd be the best compromise to enable both use-cases for now.

Yeah, Travis sometimes has issues with the timing. Might need to increase it sometime, usually I just restart it.

PR #505, opened 2017-05-16
Corrects @tag capture regular expression pattern.

cpriest:
This allows for non-whitespace characters in a tag name.  This
specifically, does not correct for inline tags such as {@link http...}
which should probably be handled differently.

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -164,7 +164,7 @@ export function parseComment(text: string, comment: Comment = new Comment()): Co
         line = line.replace(/^\s*\*? ?/, '');
         line = line.replace(/\s*$/, '');
 
-        const tag = /^@(\w+)/.exec(line);
+        const tag = /^@(\S+)/.exec(line);
         if (tag) {
             readTagLine(line, tag);
         } else {


cpriest:
I wanted to know if you would accept a PR which would implement features such as these:

* @td:Flags [flags...]
  * KeepChildOrder - Would not sure children in the navigation area
* @td:Title - Change title of the item (module, class, etc) 
* @td:Type - Change the type of the item (To other, non-standard types)

I have need of these features (with the ultimate goal of creating a full documentation site) through typedoc.  This means, specifically there would be pages generated that would be of "Kind" document, for things like examples, how to do this, how to do that for another project [hainproject/hain](https://github.com/hainproject/hain).  I already have been implementing them in a plugin but if you're interested, I can just roll them into typedoc its-self.

aciccarello:
Thanks @cpriest. Can you give some examples of what will change with this regex?

As far as adding the `@td:*` tags, could you explain your use case more? Why would you want to override those?

cpriest:
Sure, so I'm one of the leads on another project (hain), an electron launcher.  When I got involved I just wrote a plugin for it, later I wrote a typescript definition file for the plugin and have expanded on it since.

The goal of that was twofold, one so that authors had a definitions file to use for their plugins and also to produce documentation directly from the typescript file.  That began my foray into typedoc which has been great.

I ran into trouble though when it came to the "other documentation" that existed.  We currently build [this entire site](http://hainproject.github.io/hain/docs/) with Jekyll, and all API documentation is maintained by hand and is not entirely up to date.

To simplify things, I wanted to replace the entire Jekyll based website build with typedoc.  The only trouble was getting typedoc to generate additional pages, which I solved with a "stub" definitions file who's comments were markdown includes.  

So basically the above suggestions are to expand upon typedocs abilities, allowing it to be more flexible and over time, the general build for an entire front-end.  _I don't propose to make monumental changes to make this a reality, just a few additional features and I think it would be there._

My current thoughts on how to make that happen (with a plugin or theme.js) would be a hacky way to go about it and would probably break in future versions of typedoc, whereas building in the functionality (at least the sort-order feature) would be a few lines of code within typedoc its-self.  (In fact, if the DefaultTheme.js were moved to be a theme.js within the separated typedoc-theme-default, the change could be further isolated within just the theme as that is where the unconditional sort is happening).

You can see the [commits so far](https://github.com/hainproject/hain/blob/typedoc/typedoc/src/hain-guides.d.ts) towards this end.  If you wanted to see the build in a browser as it would be, you'd need to clone the project locally.

cpriest:
The change to the regex simply makes it accept all non-whitespace characters from the start of an @ tag, which I believe is accurate from my looking around typescript and jsdocs websites.  Nowhere do they mention any rules about it one way or the other, but no official tags have any space in them.

PR #510, opened 2017-05-20
Fix themejs 2

cpriest:
Alright, this one doesn't modify the stack (modifying message has no effect, so I opted for a simple application error log message.

 * I found the way you expected `require(filename)` to work and noted it in the code.  FTR these two ways are *entirely inconsistent* with the way plugins are done, which *I thought was the point* of being consistent.
 * I added a check against the themeClass that it has the resources property upon creation as it will have if derived from Theme.  (instanceof Theme does not work, as noted in the code).

### DIFF
--- a/src/lib/output/renderer.ts
+++ b/src/lib/output/renderer.ts
@@ -206,14 +206,12 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
                 this.theme = this.addComponent('theme', new DefaultTheme(this, path));
             } else {
                 try {
-                    const themeClass = typeof require(filename) === 'function' ? require(filename) : require(filename).default;
+                    const themeClass = typeof require(filename) === 'function' ? require(filename)() : require(filename).default;
 
                     this.theme = this.addComponent('theme', new (themeClass)(this, path));
                 } catch (err) {
-                    throw new Error(
-                        `Exception while loading "${filename}". You must export a \`new Theme(renderer, basePath)\` compatible class.\n` +
-                        err
-                    );
+                    err.stack = `Exception while loading "${filename}". You must export a \`new Theme(renderer, basePath)\` compatible class.\n\n` + err.stack;
+                    throw err;
                 }
             }
         }
--- a/src/lib/output/renderer.ts
+++ b/src/lib/output/renderer.ts
@@ -206,11 +206,22 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
                 this.theme = this.addComponent('theme', new DefaultTheme(this, path));
             } else {
                 try {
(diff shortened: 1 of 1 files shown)


blakeembrey:
I kept it working the way it was in the examples you posted. It was consistent between the export styles, which is the place I was saying needs to be consistent (you can't have default exports and module exports acting differently). I don't think this needs to be documented there, what I had asked to be done was to make this a utility function used in the two places so the imports were consistent - I wouldn't want to break either plugins or existing themes when they can both stay working today. I'm sorry if that wasn't clear.

I'd check the existing themes, but I think using instanceof is the correct check if you wanted to add a check. What's the intent of the check? If instanceof isn't working, it's because you have two instances of typedoc being required which maybe incompatible already (checking resources there probably isn't enough). I also would fail hard - don't try to fallback on the default theme as that'll just cause confusion from someone trying to use it.

cpriest:
> It was consistent between the export styles, 

This is where the confusion arose.  Plugins are imported entirely differently than this.  They are expected to pass back a function which is then called by typedoc with a reference to the PluginHost.  I thought since (to my knowledge) there is only a handful of existing themes, unifying the methodology would be a better choice (and right now with so few, if any, themes using theme.js, standardizing on it the same way that plugins are setup would be better).

>  I think using instanceof is the correct check

I would prefer to use that; however I tried it seven ways from sundown, and it would not work.  Even using a debugger and seeing that it is indeed inheriting from DefaultTheme/Theme, instanceof Theme still returns false.  My only explanation of this is that JS must be checking against the prototype being === and technically the theme.js has its own 'copy' of the DefaultTheme based on its separate npm install of typedoc, ergo there are two different DefaultThemes under the hood.  

I thought the solution I put in place was simpler than checking the prototype chain manually for constructor.name being correct.

> What's the intent of the check?

The intent is for the user to get a clear error message.  Without the check, if you just export a function, typedoc will happily call new (..., ...) and it will fail on `this.theme.resources.activate();` which doesn't tell the user anything.

>  I also would fail hard - don't try to fallback on the default theme as that'll just cause confusion from someone trying to use it.

I'll make that change in a moment.

blakeembrey:
Makes sense. I realised the instanceof check might be harder because of global/local usage mixing, so the current way sounds as good as it'll get. We could also check that the export is a function, but it'll crash and I'm good with that today.

I see how what I said was confused now. The import/export style was purely referring to how it was being imported (e.g. Node module or ES6 transpired module) and I didn't intend to mix it up with how it functions. I'm not familiar with how plugins work, so I never made the connection with that was what you were trying to do. I'm ok with the behaviour being different for plugins vs themes, although maybe not ideal, I think I'd make the same decision now anyway - there must be a reason it acts like those ways today right? For the export/import styles, I would like to make the particular require pattern shared between them both eventually.

Issue #488, opened 2017-04-28
References to methods that are both instance and static


jonchardy:
Say I have some class with an instance and static method of the same name:
```
export class Foo {
  ...
  public bar() { }
  public static bar() { }
  ...
}
```
and some reference to it within documentation of another class:
```
/**
 * See [[Foo.bar]].
 */
export class Baz {
  ...
}
```

Is there some way for TypeDoc to reference one of these particular methods? Currently, it seems to link to the first one. In JSDoc, this was done by using {@link Foo#bar} vs {@link Foo#.bar}.

aciccarello:
@jonchardy I don't think TypeDoc supports that currently. I honestly would like to see the links syntax aligned with JSDoc.

jonchardy:
I've been looking into this some more, and it turns out the @link tag works, but it works in the same way as the brackets, meaning you'd have to say {@link Foo.bar} rather than JSDoc's standard {@link Foo#bar}. I agree that the link syntax should work just as JSDoc does, as that's what people would expect when coming to TypeDoc.

I'll see if I can start some work on fixing the way links are parsed and also address the static issue in the process, as that's linked to JSDoc syntax.

aciccarello:
It should be noted that TypeDoc was not designed to have 100% parity with JSDoc, but I would be good with aligning the links tag. Thanks for all your investigation and work!

jonchardy:
PR #514 submitted!

jonchardy:
Ok, some notes on this, as it turns out to be pretty complicated and I won't have time to complete a robust solution...

{@link Foo#bar} links to an instance member
{@link Foo.bar} links to a static member
{@link foo.Bar.baz} links to a static member within some class in a namespace

Parsing these to pass to the findReflectionByName method is not a simple task, as we can't easily distinguish when we are looking for a static method without having more information about the structure of Reflections. We can't naively insert some token when we see '.' since it also gets used for namespaces. There probably need to be bigger changes than just updating the MarkedLinksPlugin.

nrser:
Hey, just taking TypeDoc out for a spin and hit this almost immediately. Did this issue get abandoned? How are people dealing with this?

BTW, I think I like the `[[...]]` syntax... I never really took to JSDoc & friends' `{@link ...}`-style, always felt bulky to me.

jonchardy:
We wrote something custom to handle it that works for our use cases. I don't think anyone is currently working on a general solution that matches the JSDoc spec.

[...]

PR #514, opened 2017-05-23
Allow JSDoc-style @link inline tags

jonchardy:
This PR is meant to allow for standard JSDoc syntax of @link tags to be parsed by TypeDoc. JSDoc tags follow the standards laid out here: http://usejsdoc.org/tags-inline-link.html. The main difference between the current TypeDoc implementation (which remains as is), and JSDoc is the use of '#' rather than '.'.

See also https://github.com/TypeStrong/typedoc/issues/488. This is also resolved by this PR. Converting any instance of '#.' in a tag (valid JSDoc syntax to reference a static function) to '.@static', which can then be used to distinguish members of the same name and select the static version.

The last minor change is including 'static-' in the anchor of static members, which didn't work before. It is now done at the alias level to avoid unnecessary suffixes in names.

The overlying goal here is to allow users coming from JS/JSDoc environments to more quickly get up and running with TS/TypeDoc.

As a side note, this is my first PR for this project, so in addition to any flaws, let me know about style issues.

### DIFF
--- a/src/lib/models/reflections/abstract.ts
+++ b/src/lib/models/reflections/abstract.ts
@@ -417,6 +417,9 @@ export abstract class Reflection {
             if (alias === '') {
                 alias = 'reflection-' + this.id;
             }
+            if (this.flags && this.flags.isStatic) {
+                alias = 'static-' + alias;
+            }
 
             let target = <Reflection> this;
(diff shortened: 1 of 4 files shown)


jonchardy:
Closing this as it needs some adjustments after seeing http://usejsdoc.org/about-namepaths.html#namepaths-in-jsdoc-3

PR #525, opened 2017-06-07
feat: Implement @internal as synonym of @hidden

### LINKED ISSUE #198: Support for @ignore?
I assume that the syntax typedoc uses is based off jsdoc, but I have noticed that it doesn't support all the common features, in this instance the `@ignore` option which should omit the related variable/class/property etc from the documentation.

### LINKED ISSUE #364: The @internal doc flag recognised by tsc should be an alias for @hidden
The [`@internal` flag](https://www.typescriptlang.org/docs/handbook/compiler-options.html) is used by `tsc --stripInternal` to exclude a particular property from a generated declaration file. If you're producing a documented library with public methods for consumption this is basically the same as the `@hidden` flag you'll want to use in your typedoc.

It would be great if I could just use `@internal` and have it work on both my `.d.ts` and typedoc output.

I realise PRs are needed to make progress on typedoc at the moment but I thought I'd put this up as an issue in case anybody has any better ideas!

BurtHarris:
Fixes #364 The @internal doc flag recognised by tsc should be an alias for @hidden 
Fixes #198 Support for @ignore?

### DIFF
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -109,7 +109,7 @@ export class CommentPlugin extends ConverterComponent {
             CommentPlugin.removeTags(comment, 'event');
         }
 
-        if (comment.hasTag('hidden')) {
+        if (comment.hasTag('hidden') || comment.hasTag('internal')) {
             if (!this.hidden) {
                 this.hidden = [];
             }


aciccarello:
Thanks for the PR! I do think we need to think through if `@internal` should always be hidden. I know that `@internal` has it's own meaning outside of TypeDoc. Is there anything in the semantics of that tag that would conflict with documentation?

BurtHarris:
Having thought about it some myself, this could get as complicated as you want to make it, especially since TypeScript's support for `@internal` is still experimental.   I don't think there's any shame in treating its support in typedoc as experimental.

For my project the documentation spew from typedoc is already enormous (primarily because it was ported from an [existing Java class library](http://www.antlr.org/api/Java/org/antlr/v4/runtime/package-summary.html) with extensive doc comments.   and I'm looking for any way I can to narrow what's generated to a limited "public" API in the **semver** sense.

Gerrit0:
Closing as this was implemented in #1114, which has now been merged + published. Somehow I missed this PR when making that..

PR #528, opened 2017-06-09
Add option to remove doublequotes in module names in TOC

nicknisi:
add `externalModulesOnly` option (boolean) which, if true, will not wrap module names in doublequotes. I was researching why these names are wrapped in quotes and stumbled across #68 discussing this issue, which [suggested a possible CLI flag](https://github.com/TypeStrong/typedoc/issues/68#issuecomment-75469700) to disable this functionality if desired.

Here's a screenshot of the change in a project where I am not using namespaces (internal modules).

napkin 06-09-17 3 04 00 pm

Also, I'm open to alternatives for the name of the option. Some others I've thought of include 
- `assumeNoNamespace`
- `removeQuotes`
- `externalModules`

Any help on the naming would be greatly appreciated.

### DIFF
--- a/src/lib/converter/plugins/DynamicModulePlugin.ts
+++ b/src/lib/converter/plugins/DynamicModulePlugin.ts
@@ -4,6 +4,8 @@ import * as Path from 'path';
 import {Reflection, ReflectionKind} from '../../models/reflections/abstract';
 import {Component, ConverterComponent} from '../components';
 import {BasePath} from '../utils/base-path';
+import {Option} from '../../utils/component';
+import {ParameterType} from '../../utils/options/declaration';
 import {Converter} from '../converter';
 import {Context} from '../context';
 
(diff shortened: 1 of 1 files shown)


aciccarello:
Yes! I think it would be great to get rid of those quotes. Honestly I'd like them to be off by default.

nicknisi:
I would too 😄, honestly. It looks like they are there to distinguish namespaces (formerly internal modules) with external modules (using the ESM `export` keyword. I personally don't use namespace all that often, so I'm not sure how big of an issue that would be to turn them off by default.

blakeembrey:
Is there any way we can check if a "module" is external (e.g. in that case, node.js `fs`) or internal?

jaredcnance:
has this PR been abandoned? what still needs to be done to get this cleaned up?

nicknisi:
Oh wow, sorry I never followed up. I'll re-orient myself and follow up properly.

Issue #533, opened 2017-06-16
Expose Reflection classes to be able to create custom theme


pragone:
Hi,

I've been trying to find a theme to produce a Markdown version of the docs. Found references to https://github.com/kimamula/typedoc-markdown-theme/ . And found that it's not working anymore as it has been mentioned in another issue. (If memory serves me well it stopped working after 0.6)

The issue I'm having is that I want it to produce .md files (instead of .html) and for this I need it to create a theme that goes beyond simply replacing the template files of the Default Theme. I need to actually include a theme.js file with a subclass of DefaultTheme.

This leads to my problem. The Reflection classes are not exported by the typedoc project. And a lot of the logic of the DefaultTheme (which I'm using as a base) depends on "instanceof" logic. Without these classes being exposed I'm finding that I can only use imports of the type:
`import { DefaultTheme, TemplateMapping } from 'typedoc/dist/lib/output/themes/DefaultTheme';`
but this doesn't really work, as I then experience issues on runtime where the DefaultTheme class loaded by typedoc is different than the DefaultTheme class loaded by the template and instanceof checks fail.

It's not a big change, and I'll submit a pull request with the classes I think would make sense to expose for the development of custom themes, but thought it would be good to ask whether there was any rationale behind why these classes are not exposed or an alternative way to tackle this issue.

aciccarello:
@pragone Thanks for investigating that. Could you explain how this relates to #382? With #504 you should be able to use a theme.js file.

I think I follow you basic premise but I'm not understanding why the instanceof on the deep imports aren't working. Shouldn't it be pointing to the same instances?

pragone:
Hi @aciccarello.
Indeed I had looked at that PR. What it does is to include the theme.js file and load it properly. Which is OK. That's not the issue I was having. This is the issue.

Now, my theme.js file must export a class that extends DefaultTheme. But this class is not exposed by the typedoc package. Also, in its internal logic it needs other classes like DeclarationReflection for parts like:
```
            entryPoint.children.forEach((child: Reflection) => {
                if (child instanceof DeclarationReflection) {
                    TypedocMarkdownTheme.buildUrls(child, urls);
                }
            });
```
That DeclarationReflection is not exposed by the typedoc package either. This means that the only way I can access it is by doing a require like:
```
import { DeclarationReflection } from 'typedoc/dist/lib/models/reflections/declaration';
```

[...]

PR #535, opened 2017-06-17
Exposed additional classes to ease the creation of non-trivial themes (#533)

pragone:
This is related to issue #533 
With this change I'm exposing additional classes necessary to create a theme that goes beyond simply replacing template files


pragone:
Strange... Build 747 failed with a timeout.
But there was a 747.1 that passed: https://travis-ci.org/TypeStrong/typedoc/jobs/243908732

Zone-Ghost:
Purely a timeout setting issue.  Several of the tests approach 4 seconds, but don't go over.  That particular one does, so it's going over.  Runs fine on my macbook testing with node 4.2 and 8.1.

Care to crank the npm "test" script to `-t 5000`?

PR #545, opened 2017-06-30
Gracefully handle GitHub plugin git exec failures

bingenito:
If shelljs exec of git rev-parse fails at the exec level, return null to add to ignored paths and continue.


bingenito:
@aciccarello Raises a different question I suppose - is there a way to run with all default plugins but that one?  The specific situation here is an extremely non-standard node location on a unc path which causes problems for shelljs.exec.  It's easier to handle the error and continue.

PR #558, opened 2017-07-15
SOLVED tslint

aleixmorgadas:


aciccarello:
Sorry for making a mess of the commits. I was using the online GH editor.

I made some adjustments, adding a space before exports and removing a few double whitespaces. When the checks come back (they should be fine) I'll squash and merge.

PR #573, opened 2017-07-31
Support latest TypeScript version.

lilezek:
I changed a few things to update from version 2.3 to 2.5-dev.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "progress": "^2.0.0",
     "shelljs": "^0.7.0",
     "typedoc-default-themes": "^0.5.0",
-    "typescript": "2.4.1"
+    "typescript": "^2.5.0-dev.20170727"
   },
   "devDependencies": {
     "@types/mocha": "^2.2.39",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "@types/mocha": "^2.2.39",
+    "@types/node": "^8.0.17",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -46,10 +46,11 @@
     "progress": "^2.0.0",
     "shelljs": "^0.7.0",
     "typedoc-default-themes": "^0.5.0",
-    "typescript": "2.4.1"
+    "typescript": "^2.5.0-dev.20170727"
```
blakeembrey: We need to use exact versions, but I won't be merging anything that's pre-release. You need to either build a pre-release pipeline to support this into the future, or wait for a stable release.

blakeembrey:
Thanks for the contribution. I'd love to accept it, but we don't currently have a pre-release pipeline set up. Feel free to update the PR when a stable 2.5 is released. On the other hand, we could definitely use a pre-release pipeline for people who want that, but it's a fair bit of work to implement.

lilezek:
Well, I've seen a bug using the latest supported version (2.3.4? can't
remember) while documenting code which compiled perfectly after patching
this. I'll paste here the code that reproduces the bug and see if there is
a stable release that documents without crashing.

lilezek:
I've just seen that 5 days ago this project was updated to 2.4.1, so the bug may be gone.

aciccarello:
TypeScript is now at 2.5.3. We can probably update this.

aciccarello:
Resolved via #626

PR #579, opened 2017-08-09
pin @types/handlebars version

harunurhan:
* Fixes errors caused by higher version of @types/handlebars due to
npm's auto update

Reference: https://github.com/TypeStrong/typedoc/pull/384#issuecomment-289107077


aciccarello:
Thanks for the PR! I think we should pin all the `@types` dependencies since they [don't conform to semver](https://github.com/TypeStrong/typedoc/pull/384#issuecomment-289107077).

harunurhan:
@aciccarello I agree I could also do in this PR if it's asked, I just pinned the one that I have problem with. I would appreciate if this is merged soon!

harunurhan:
best open source project ever

aciccarello:
@harunurhan I was hoping we'd pin all the types. Would you mind updating that?

harunurhan:
@aciccarello done, FYI, I had to pin `@types/lodash` a bit higher due to an error with the current one.

PR #581, opened 2017-08-12
Fix type converter removal

### LINKED ISSUE #559: Possible bug removing type converters in converter.ts
I was reviewing the code and came across the [`removeTypeConverter` method](https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/converter.ts#L237-L247). This method seems to be looking up an item's position in an array and removing it. However, the method looks up the item's position in the `typeNodeConverters` array twice which leads me to think that the first check should be checking the `typeTypeConverters` array.

Does anyone else think this is wrong? Is this causing bugs other than a memory leak?

aciccarello:
Closes #559

### DIFF
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -235,14 +235,14 @@ export class Converter extends ChildableComponent<Application, ConverterComponen
     }
 
     private removeTypeConverter(converter: ConverterTypeComponent) {
-        let index = this.typeNodeConverters.indexOf(<any> converter);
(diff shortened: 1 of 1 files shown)


blakeembrey:
@aciccarello Feel free to merge/close issues/PRs you want to. I've been going offline with open source lately since it became too much for me to deal with. I'll be on vacation until October now, and may be a bit more active then, but there's no guarantee - I still have a backlog of thousands of issues and PRs that I've been slowly working through.

PR #584, opened 2017-08-22
Constructor parameters with readonly flag define properties.

### LINKED ISSUE #352: Support "readonly" as a parameter property.
### Versions

typedoc: 0.5.1
typescript: 2.0.10

### Steps to Reproduce

1. Create a project with this file named `test.ts`:

    ```
    class Foo {
        constructor(private a: string, readonly b: string) {
        }
    }

    const f: Foo = new Foo("1", "2");

    console.log(f.b);
    ```

2. Run `typedoc` on it.

### Expected Results

The documentation should list both `a` and `b` as properties of the class `Foo`.

### Actual Results

Only `a` is present as property of `Foo` but `b` is missing. (They both appear as parameters to the constructor, which is fine.)

lddubeau:
The signature ``constructor(public foo: string)`` defines a public property corresponding to the parameter. The signature ``constructor(readonly foo: string)`` defines a public readonly property but prior to this commit, typedoc would not show this property in the generated documentation.

Fixes #352.

### DIFF
--- a/examples/basic/src/classes.ts
+++ b/examples/basic/src/classes.ts
@@ -286,8 +286,9 @@ export class GenericClass<T extends BaseClass>
      * @param p2 Private string property
      * @param p3 Public number property
      * @param p4 Public implicit any property
+     * @param p5 Readonly property
      */
-    constructor(p1, protected p2:T, public p3:number, private p4:number) {
+    constructor(p1, protected p2:T, public p3:number, private p4:number, readonly p5: string) {
     }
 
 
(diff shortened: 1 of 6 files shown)


thopaw:
Is it possible to also use the doc added to the constructor property?

```
class A {
  constructor(
    /**
     * I want to see this comment in the documentation
     * /
     readonly property: string) {}
}
```

Is it possible to see the 

> I want to see this comment in the documentation

in the generated HTML?

Issue #527, opened 2017-06-08
What's wrong with my command line?


BurtHarris:
I'm trying `typedoc --mode modules --out doc .\src\antlr4ts.ts` 

The command parser seems to be treating this as if I haven't specified any input files, but obviously I have.  **.\src\antlr4ts.ts**

BurtHarris:
I think I've answered my own question, there must have been something wrong in my `tsconfig.json`.   I tried deleting it and got some output...

aciccarello:
Hi @BurtHarris, we may need to look into support of the `include` option.

mnpenner:
When I run this,

    typedoc --out docs --mode modules --module commonjs --readme none --target es6 --name jtilz ./dist/node/I_DONT_EXIST

It doesn't error out. Just completely ignores the source files I give it.

Is it overriding it in favour of my `tsconfig.json`? I'd really prefer that it didn't.

fr0:
Is there any way to tell typedoc to only generate documentation for a specific subfolder of my project? No matter what options I give it, it seems to want to document the entire folder.

natevoci:
I have a separate tsconfig.json file for running typedoc where I specify the folders I want included in the documentation in the include block.

```
typedoc ... --tsconfig tsconfig.docs.json
```

```
{
  "compilerOptions": {
    ...
  },
  "include": [
    "main/folder/**/*",
    "path/to/another/folder/**/*"
  ]
}
```

This works for me, but I only use the json output option. I have not tried the the html generated docs to see if it does the same thing.

PR #587, opened 2017-09-04
Clobber generic/default options with more specific options

thecjharries:
This fixes an issue where `typedoc` clobbers the CLI path with the found options from `tsconfig.json` or `typedoc.js`. It took me a bit to figure everything out:

* `ArgumentsReader` has [no priority](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/arguments.ts#L11), whereas the priorities for [`TypedocReader`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/typedoc.ts#L25) and [`TSConfigReader`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/tsconfig.ts#L27) are both `-100`.
* [`onApi` sorts](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/events.ts#L112) from biggest to smallest. To verify,
```javascript
[0, -100, -100].sort((a, b) => b - a)
// [0, -100, -100]
new Array(10).fill(0).map((value, index) => index).sort((a, b) => b - a)
// [9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
```
* Because `ArgumentsReader` has the largest priority, it executes first. If it encounters [an unknown argument](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/options.ts#L37), it [`push`es the filename](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/arguments.ts#L57) to `event.inputFiles`.
* Because `TSConfigReader` has a lower priority, it executes second. It searches the [`tsconfig.json` for filenames](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/tsconfig.ts#L64). To do so, it uses [`parseJsonConfigFileContent`](https://github.com/Microsoft/TypeScript/blob/release-2.4/src/compiler/commandLineParser.ts#L1104), which [checks for file specs](https://github.com/Microsoft/TypeScript/blob/release-2.4/src/compiler/commandLineParser.ts#L1145). If it doesn't find any, it returns [files matching `**/*`](https://github.com/Microsoft/TypeScript/blob/release-2.4/src/compiler/commandLineParser.ts#L1189) that do not match [common package folders](https://github.com/Microsoft/TypeScript/blob/release-2.4/src/compiler/commandLineParser.ts#L1189). In essence, it will **always** return non-empty `fileNames`. Even were `fileNames` empty, [`event.inputFiles` is still clobbered](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/tsconfig.ts#L71).
* Putting all of this together, this means [the usage docs](http://typedoc.org/guides/usage/#arguments) are wrong, as is [the index in `src/test/renderer/specs/index.html`](https://github.com/TypeStrong/typedoc/blob/master/src/test/renderer/specs/index.html#L90). Unless `ArgumentsReader`'s priority is lowered (or a different solution is used), there's absolutely no way to pass in the source directory/files via the command line.

I've been trying to figure this out for too many hours, so I don't have a demo prepped. I can make one later. This folder structure
```
/project
    <optional tsconfig.json>
    some-file.ts
    src/
        some-other-file.ts
    docs/
```
used with this command
```
typedoc --out docs/ src/some-other-file.ts
```
will always generate documentation for both `some-other-file` AND `some-file`.

### DIFF
--- a/src/lib/utils/options/readers/arguments.ts
+++ b/src/lib/utils/options/readers/arguments.ts
@@ -8,7 +8,7 @@ import { ParameterType } from '../declaration';
 @Component({name: 'options:arguments'})
 export class ArgumentsReader extends OptionsComponent {
     initialize() {
-        this.listenTo(this.owner, DiscoverEvent.DISCOVER, this.onDiscover, -299);
+        this.listenTo(this.owner, DiscoverEvent.DISCOVER, this.onDiscover, -200);
     }
 
     onDiscover(event: DiscoverEvent) {
--- a/src/lib/utils/options/readers/arguments.ts
+++ b/src/lib/utils/options/readers/arguments.ts
@@ -8,7 +8,7 @@ import { ParameterType } from '../declaration';
 @Component({name: 'options:arguments'})
 export class ArgumentsReader extends OptionsComponent {
     initialize() {
-        this.listenTo(this.owner, DiscoverEvent.DISCOVER, this.onDiscover);
+        this.listenTo(this.owner, DiscoverEvent.DISCOVER, this.onDiscover, -299);
     }
 
     onDiscover(event: DiscoverEvent) {
@@ -46,6 +46,7 @@ export class ArgumentsReader extends OptionsComponent {
             }
         }
 
+        const files = [];
         while (index < args.length) {
             const arg = args[index++];
 
@@ -54,8 +55,11 @@ export class ArgumentsReader extends OptionsComponent {
             } else if (arg.charCodeAt(0) === _ts.CharacterCodes.minus) {
                 readArgument(arg.slice(arg.charCodeAt(1) === _ts.CharacterCodes.minus ? 2 : 1).toLowerCase());
             } else {
-                event.addInputFile(arg);
+                files.push(arg);
             }
+        }
+        if (files) {
+            event.inputFiles = files;
         }
     }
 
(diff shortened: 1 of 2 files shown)


thecjharries:
This should address #527. I didn't think to check the issues until after submitting the PR. I'll search for more after I get some sleep.

thecjharries:
The branch name isn't as descriptive now. After thinking about this a bit morning this morning, I realized the duplicated priority in [`TypedocReader`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/typedoc.ts#L25) and [`TSConfigReader`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/tsconfig.ts#L27) could, in rare cases, lead to unexpected results.

By changing the priority, the load order is now `TSConfigReader` > `TypedocReader` > `ArgumentsReader`, which allows the most generic to be overwritten by the most specific (e.g. `inputFiles` start at `**/*` with a generic `tsconfig.json`, are replaced by `src` inside a `typedoc.js`, and are overwritten by a specified CLI argument).

`Application`s (and their descendants `CliApplication`s) are [`bootstrap`ped via two option reads](https://github.com/TypeStrong/typedoc/blob/master/src/lib/application.ts#L107), allowing generics to be built in order of most to least specific in the first read, then replaced by the desired objects in the second read. Discrete priorities means the order is finally well-defined and the `Options` will load in a predictable manner.

blakeembrey:
Thanks @thecjharries!

beevik:
This commit seems to have broken the processing of the `"files"` array in `tsconfig.json`.

See https://github.com/TypeStrong/typedoc/issues/617

PR #591, opened 2017-09-11
fix wrong enum type conversion

shlomiassaf:
Running the [enumerations basic example](https://github.com/TypeStrong/typedoc/blob/master/examples/basic/src/enumerations.ts) the enums are evaluated as they should but member, variables, signatures and every other reflection that contains a type will resolve the wrong type.

If we look at the enumerations example, this is a short version of the code: 
```ts
export enum Size
{
    Small,

    Medium,

    Large
}

export module Size
{
    var defaultSize:Size = Size.Medium;
    
    function isSmall(value:Size):boolean {
        return value == Size.Small;
    }
}
``` 

Looking at the JSON output (slimmed down) for the `isSmell` member:
```json
{
  "id": 367,
  "name": "isSmall",
  "kind": 64,
  "kindString": "Function",
  "flags": {
    "isExported": true
  },
  "signatures": [
    {
      "id": 368,
      "name": "isSmall",
      "kind": 4096,
      "kindString": "Call signature",
      "parameters": [
        {
          "id": 369,
          "name": "value",
          "kind": 32768,
          "kindString": "Parameter",
          "type": {
            "type": "union",
            "types": [
              {
                "type": "unknown",
                "name": "Size.Small"
              },
              {
                "type": "unknown",
                "name": "Size.Medium"
              },
              {
                "type": "unknown",
                "name": "Size.Large"
              }
            ]
          }
        }
      ],
      "type": {
        "type": "intrinsic",
        "name": "boolean"
      }
    }
  ]
}
```

The type for the parameter is wrong, it treats it as union type instead of the right type which is reference type to the enum reflection.

Also see visually from a rendered page:
image

After applying the fix:

```json
{
  "id": 205,
  "name": "isSmall",
  "kind": 64,
  "kindString": "Function",
  "flags": {
    "isExported": true
  },
  "signatures": [
    {
      "id": 206,
      "name": "isSmall",
      "kind": 4096,
      "kindString": "Call signature",
      "parameters": [
        {
          "id": 207,
          "name": "value",
          "kind": 32768,
          "kindString": "Parameter",
          "type": {
            "type": "reference",
            "name": "Size",
            "id": 200
          }
        }
      ],
      "type": {
        "type": "intrinsic",
        "name": "boolean"
      }
    }
  ]
}
```
image

### DIFF
--- a/src/lib/converter/types/enum.ts
+++ b/src/lib/converter/types/enum.ts
@@ -11,7 +11,7 @@ export class EnumConverter extends ConverterTypeComponent implements TypeTypeCon
      * Test whether this converter can handle the given TypeScript type.
      */
     supportsType(context: Context, type: ts.Type): boolean {
-        return !!(type.flags & ts.TypeFlags.Enum);
+        return !!(type.flags & ts.TypeFlags.EnumLike);
     }
 
     /**
--- a/src/lib/converter/types/union-or-intersection.ts
+++ b/src/lib/converter/types/union-or-intersection.ts
@@ -17,7 +17,7 @@ export class UnionOrIntersectionConverter extends ConverterTypeComponent impleme
      * Test whether this converter can handle the given TypeScript type.
      */
     supportsType(context: Context, type: ts.UnionOrIntersectionType): boolean {
-        return !!(type.flags & ts.TypeFlags.UnionOrIntersection);
+        return !!(type.flags & ts.TypeFlags.UnionOrIntersection) && !(type.flags & ts.TypeFlags.EnumLiteral);
     }
 
     /**


bdbergeron:
I've updated to the latest release of typedoc, but my docs still generate with the inlined union of possible values instead of using the enum type. Any suggestions on how to fix this would be great.

```ts
/**
 * Available leaderboard types
 */
enum LeaderboardType {
    /** Current context */
    CURRENT_CONTEXT = "currentContext",
    /** All contexts */
    ALL_CONTEXTS = "allContexts",
    /** All friends, all-time */
    ALL_TIME = "allTime",
}

/**
 * Show leaderboards UI.
 *
 * @param type  Leaderboard type to show.
 */
function show(type: LeaderboardType): void;
```

screen shot 2017-10-12 at 1 21 20 pm

aciccarello:
@bdbergeron I wonder if there is an issue with string enums. We should create a test for that.

bdbergeron:
@aciccarello It looks like it was an issue with gulp-typedoc specifying an older version in their package.json that was overriding the one in my package.json. I somehow must have missed them pushing an update about two weeks ago that fixed the version mismatch, but this issue is now resolved for me!

PR #592, opened 2017-09-11
Allow multiple excludes

haraldF:
Allow the exclude parameter to be an Array as well as a
minimatch string. This allows for greater flexibility when excluding
files as minimatch regexs with 'or' patterns quickly become unreadable /
unmaintainable.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -72,10 +72,10 @@ export class Application extends ChildableComponent<Application, AbstractCompone
 
     @Option({
         name: 'exclude',
-        help: 'Define a pattern for excluded files when specifying paths.',
-        type: ParameterType.String
+        help: 'Define patterns for excluded files when specifying paths.',
+        type: ParameterType.Array
     })
-    exclude: string;
+    exclude: Array<string>;
 
     /**
      * The version number of TypeDoc.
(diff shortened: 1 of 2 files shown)


haraldF:
@blakeembrey - can this be merged?

src/lib/application.ts
```diff
@@ -238,9 +238,23 @@ export class Application extends ChildableComponent<Application, AbstractCompone
...
     public expandInputFiles(inputFiles?: string[]): string[] {
-        let exclude: IMinimatch, files: string[] = [];
+        let exclude: Array<IMinimatch>, files: string[] = [];
         if (this.exclude) {
-            exclude = new Minimatch(this.exclude);
+            exclude = this.exclude.map(pattern => new Minimatch(pattern));
+        }
+
+        function isExcluded(fileName: string): boolean {
+            if (!exclude) {
```
blakeembrey: This should never happen, you're initialising it already.

src/lib/application.ts
```diff
@@ -238,9 +238,23 @@ export class Application extends ChildableComponent<Application, AbstractCompone
...
+                return false;
+            }
+
+            for (let mm of exclude) {
```
blakeembrey: This should be a simple one-liner, `return exclude.some(x => x.match(fileName))`.

src/lib/application.ts
haraldF: It was only initialized when this.exclude was set. I've updated the initializer a bit now to make it look prettier.

aciccarello:
How would this be used from the command line? I tried `--exclude "**/custom-error.ts" --exclude "**/other.ts"` which only used the last minmatch and `--exclude ["**/custom-error.ts","**/other.ts"]` which didn't exclude anything.

It'd be great to also update the [usage documentation](https://github.com/TypeStrong/typedoc-site/blob/gh-pages/_guides/usage.md) as well with this new option.

haraldF:
I am only using the option from a config file. As for documentation - we should add a general comment about all array parameters, not only this one, in a separate PR request

lednhatkhanh:
Can this being merged, we really need this for our project...

lednhatkhanh:
@haraldF Could you please add an example for this, I can't make it work.

aciccarello:
@harunurhan Yeah, some documentation is certainly needed. I believe this is only accessible programmatically. Are you passing typedoc options via the CLI?

lednhatkhanh:
I achieved that by using a config file. Thanks.

PR #593, opened 2017-09-14
ContextAwareRendererComponent, MarkedLinksPlugin and DefaultTheme  to respect absolute urls

shlomiassaf:
A plugin might set the `url` property on a reflection before the rendering phase.

The rendering plugins (and the theme) does not care, they will overwrite it.

This PR adds a check before overwrite, if the URL is absolute (http, https, ftp) the URL generation will stop since its an external link.

### DIFF
--- a/src/lib/output/components.ts
+++ b/src/lib/output/components.ts
@@ -29,6 +29,11 @@ export abstract class ContextAwareRendererComponent extends RendererComponent {
     private location: string;
 
     /**
+     * Regular expression to test if a string looks like an external url.
+     */
+    protected urlPrefix: RegExp = /^(http|ftp)s?:\/\//;
+
+    /**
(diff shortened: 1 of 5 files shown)


src/lib/output/components.ts
```diff
@@ -29,6 +29,11 @@ export abstract class ContextAwareRendererComponent extends RendererComponent {
     private location: string;
 
     /**
+     * Regular expression to test if a string looks like an external url.
+     */
+    protected urlPrefix: RegExp = /^(http|ftp)s?:\/\//;
```
blakeembrey: Do you think we should just do `/^\w{1,10}:\/\//` to ignore everything that looks absolute?

shlomiassaf: I guess you're right here, I just wanted to PR a fix, changing to such a general thing might be a breaking change

PR #611, opened 2017-10-08
Implement Abstract, Const and Let flags

shlomiassaf:
This PR add the following enums to `ReflectionFlag`:

  - Abstract
  - Const
  - Let

Additionally, each enum has it's implemented converter logic added.

# Why?
No need to explain why `Abstract` symbols should reflect in the UI.
For Let and Const, these are also important, Const is a read only symbol so the user should be aware of that. Let does not have much effect since we expose structure and not implementation hence let and var in that context are the same, however, if we state const we should state let as it's a language feature and should reflect in the UI to be consistent.

### Abstract
In `ClassConverter` added logic to set `Abstract` flag to each `DeclarationReflection` that represents an abstract class, following the same logic for each abstract member of that class (`FunctionConverter`, `VariableConverter`) 

Theming Notes:
  - The theme's does not require change for members as the `Abstract` enum is also set as a relevant flags, hence showing in native themes.
  - The `overwrites` property of a member the overwrites an abstract parent member is left as is. This means it will still show overwrites which is not the right terminology, might require a custom fix in the templates. You can see if you "overwrite" an abstract parent by checking if it has the `Abstract` flag.
  - Abstract classes have no UI representation as it does not exist in the theme's, need to implement in the themes.

### Const & Let
In `VariableConverter` added logic to set `Const` or `Let` flag to each `DeclarationReflection` that represents a variable where the typescript node's parent has Const or Let node flags set.
A `var` variable should be inferred whenever a reflection of kind `ReflectionKind.Variable` has no Let or Const flags.

Theming Notes:
  - The theme's does not require change because `Const` and `Let` enums are also set as a relevant flags, hence showing in native themes. (var will not display, as is now)

### DIFF
--- a/examples/basic/src/classes.ts
+++ b/examples/basic/src/classes.ts
@@ -50,12 +50,12 @@ export interface IPrintNameInterface extends INameInterface, IPrintInterface
  *
  * [[include:class-example.md]]
  */
-export class BaseClass implements INameInterface
+export abstract class BaseClass implements INameInterface
 {
     /**
      * This is a simple public member.
      */
-    public name:string;
+    public abstract name:string;
 
     /**
      * This is a simple protected member.
@@ -90,6 +90,7 @@ export class BaseClass implements INameInterface
         this.checkName();
     }
 
+    public abstract abstractMethod(): void;
 
     /**
      * This is a simple member function.
@@ -197,6 +198,8 @@ class InternalClass
  */
 export class SubClassA extends BaseClass implements IPrintNameInterface
 {
+    public name:string;
+
     /**
      * This is a simple interface function.
      */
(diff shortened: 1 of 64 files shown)


shlomiassaf:
image

image

shlomiassaf:
@aciccarello There's no harm in being explicit here... 

Consider a `namespace` export with nested `namespace` exports.
`var` and `let` has logical meaning in such scenario.

Of course rare scenario :) but it might be.


Anyway, we should **let** the theme decide whether to display it or not... by default it should be on.

IMHO, it should always be clear, var let or const...

aciccarello:
Makes sense to me. Thanks for your work!

PR #613, opened 2017-10-09
Reflections in the root scope of a file are exported with --mode "file"

makana:
When using --mode "file" all reflections within the root scope of a file are exported although the keyword is not written.

Actually, that's just half the story. Typically this is the case if the "module" option in a tsconfig file is set to "none". Maybe this should be analyzed instead making the mode option deprecated?


aciccarello:
Hello @makana could you explain a little more of what is going on here? I don't understand the motivation. Thanks!

makana:
Hi @aciccarello 

as we are not using any module system (but "module": "none" in tsconfig.json) we have to use --mode "file" for typedoc. So in our case, a file does not represent a module, we have to use namespaces instead.

With any kind of module system a ts file would look like this:
```typescript
class AnyNonExportedClass
{
}
export class ExportedClass
{
}
export function exportedFunction
{
}
```

In our case a ts file looks like this:
```typescript
namespace This.Is.An.Exported.Namespace
{
    export ExportedClassInNamespace
    {
    }
}
class ExportedGlobalClass
{
}
```

So anything that is written in the "root scope" of a file is exported although the "export" keyword is not written.

This PR addresses the issue that those classes/objects/etc. weren't exported before when using --mode "file".

So ``(container instanceof ProjectReflection && !context.isDeclaration && context.converter.mode === SourceFileMode.File)`` means "anything in a source file that is a direct child of the project (= not nested in any namespace) when using file mode" has to be exported.

Gerrit0:
This will be fixed with #1157.

Gerrit0:
Has been fixed with #1157, sorry this took so long to fix!

PR #614, opened 2017-10-10
add the TypeOperatorType that represents an op result on a type ("keyof")

shlomiassaf:
The `TypeOperatorType` represent the typescript `TypeOperator` node type.

```ts
class A {
  a: string;
  b: number;
}

class B<T extends keyof A> {}
class B<T extends A> {}
```

In the example above, the `Reflection` instances for **class B** and **class C** will include a `TypeParameterType` **T** that has a constraint.

For **class C** it will be a `TypeReferenceType` (reference to reflection of **A**)

For **class B** it will be a `TypeOperatorType` with a `target` property holding a `TypeReferenceType` instance (reference to reflection of **A**)

Before this PR the behaviour is to include a union type that contains `IntrinsicType` of string's that represent the properties of **class A** ('a', 'b');

While this is logically true it is not accurate, the actual type is the list of keys of another type.
The documentation should show the connection between T and A and not just show a list of strings.

Some classes or types might have 10-15 or more members, it does not scale

### DIFF
--- a/examples/basic/src/classes.ts
+++ b/examples/basic/src/classes.ts
@@ -73,7 +73,7 @@ export class BaseClass implements INameInterface
     /**
      * This is an instance member of an internal class.
      */
-    private internalClass:InternalClass;
+    private internalClass:InternalClass<keyof BaseClass>;
 
 
     constructor(name:string);
@@ -181,7 +181,7 @@ export class BaseClass implements INameInterface
 /**
  * This is an internal class, it is not exported.
  */
-class InternalClass
+class InternalClass<TTT extends keyof BaseClass>
 {
     constructor(options:{name:string}) {
 
--- a/src/lib/converter/types/index.ts
+++ b/src/lib/converter/types/index.ts
@@ -9,5 +9,6 @@ export { ReferenceConverter } from './reference';
 export { ThisConverter } from './this';
 export { TupleConverter } from './tuple';
 export { TypeParameterConverter } from './type-parameter';
+export { TypeOperatorConverter } from './type-operator';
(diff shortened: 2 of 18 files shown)


shlomiassaf:
TODO: Add serialization support to #597 or here...

shlomiassaf:
image

image

aciccarello:
@shlomiassaf Great work implementing this! Can you explain a little more your motivation for the name `TypeOperator`? Is that the TypeScript node name? If it is that makes sense, but my initial reaction was that since `keyof` is the only operator we should just refer to it as keyof.

shlomiassaf:
@aciccarello 
Yes, it is the TypeScript name, see [this line](https://github.com/shlomiassaf/typedoc/blob/28ba3497097f8935e5252c95377f273e512cd368/src/lib/converter/types/type-operator.ts#L8) at the end.

It reference to the `Node` for that type (i.e. `TypeNode`) hence the name `TypeOperatorNode`
The `SyntaxKind` for this node (i.e. `node.kind`) is `TypeOperator` and this is where the name comes from, consistent with other type names in the project.

If you look at the source code you will see that the `TypeOperatorNode` has a `operator` property, That property is fixed to `SyntaxKind.KeyOfKeyword`.

So, property is fixed to `keyof` and indeed `keyof` is the only operator there is (for types).

However, if think about it, there is no reason for the TS team to create a fixed `operator` property for this node if `keyof` is the only operator...
It must be a place holder for future types, when and if they come.

So, to be prepared I followed their structure and it means that if they add more operators we are ready for it without breaking anything.

Issue #472, opened 2017-04-07
Namespace/Type & Namespace/Interface declaration merging


m1gu3l:
It's valid in TypeScript to merge type (or interface) with namespace.
```
// declare
type A = 1;
type B = 2;
type C = 3;
export type Letter = A | B | C;
export namespace Letter {
 export function random(): Letter { return <Letter> Math.ceil(Math.random() * 3); }
}
```
```
// then use as both type and module
const letter: Letter = Letter.random();
```

TypeDoc ignores type/interface and prints documentation only for module. I understand it's not very common pattern, but since it's legal it should be handled.

aciccarello:
Agreed that this should be handled better. We need to determine if this is an issue with the parser or with the theme output.

makana:
I think, the type is sometimes generated as type/interface, sometimes as module. I encountered both. Maybe it is some kind of internal what-is-processed-first-and-what-is-merged-to-the-other.

There should be a better way to deal with those types. Maybe expanding a module kind child node below the real type/interface is best.

That would probably resolve some more odd issues here as well. Having a class that acts as a namespace as well and defining variables/functions within that namespace outputs them as static properties/methods of the class (which is not really true but kind of i-can-deal-with-that), that inherit the exported state of the class (which is clearly not true since you have to export variables/functions explicitely).

See related #615

aciccarello:
According to the TypeScript [Declaration Merging documentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), merging namespaces with enums and classes is an intentional pattern to augement the enum or class. IMO these should continue to come through as enums and classes.

As far as merging namespaces with types and interfaces, I'm not sure what the desired functionality is. [TypeScript seems to only want to use the type](http://www.typescriptlang.org/play/index.html#src=type%20A%20%3D%201%3B%0D%0Atype%20B%20%3D%202%3B%0D%0Atype%20C%20%3D%203%3B%0D%0Aexport%20type%20Letter%20%3D%20A%20%7C%20B%20%7C%20C%3B%0D%0Aexport%20namespace%20Letter%20%7B%0D%0A%20export%20function%20random()%3A%20Letter%20%7B%20return%20%3CLetter%3E%20Math.ceil(Math.random()%20*%203)%3B%20%7D%0D%0A%7D%0D%0A%0D%0Alet%20variable%3A%20Letter%3B%0D%0Avariable.random()%3B).

makana:
[...]

PR #615, opened 2017-10-10
Fix the exported flag of variables or functions in types that act as a namespace as well

### LINKED ISSUE #392: Exported flag for variable declarations not correct?
Can anyone give an explanation why the parent context is considered for the isExported flag of variable declarations?
(lib/converter/factories/declaration.ts)

```ts
if (kind == ReflectionKind.ExternalModule) {
    isExported = true; // Always mark external modules as exported
} else if (node.parent && node.parent.kind == ts.SyntaxKind.VariableDeclarationList) {
    var parentModifiers = ts.getCombinedModifierFlags(node.parent.parent);
    isExported = isExported || !!(parentModifiers & ts.ModifierFlags.Export)
} else {
    isExported = isExported || !!(modifiers & ts.ModifierFlags.Export);
}
```
    

I often use classes, interfaces or enums as namespaces as well. It is similar to nested classes in C#.

For example I have a lot of auto generated enumeration code like the following:
    
```ts
namespace Geometry
{
	export enum GeometryTypeFamily
	{
		Null = 0,
		LineString = 1,
		Point = 2,
		Polygon = 3,
		Circle = 4,
	}
	export namespace GeometryTypeFamily
	{
		const ___labels: { [index: number]: string } = {
			[GeometryTypeFamily.LineString]: "Linie",
			[GeometryTypeFamily.Point]: "Punkt",
			[GeometryTypeFamily.Polygon]: "Fläche",
			[GeometryTypeFamily.Circle]: "Kreis",
		}
		export function getLabel(v: GeometryTypeFamily): string
		{
			return ___labels[v] || GeometryTypeFamily[v];
		}
	}
}
```

With the current implementation isExported is true for ___labels which is wrong.
Even if I use the --excludeNotExported flag, ___labels is visible in the typedoc output.

Same thing for classes or interfaces that are used as namespaces at the same time.

I reduced the code above to the following lines and it works fine. Am I wrong? Is there anything I don't see?

```ts
if (kind == ReflectionKind.ExternalModule) {
    isExported = true; // Always mark external modules as exported
} else if (node.parent && node.parent.kind == ts.SyntaxKind.VariableDeclarationList) {
    isExported = !!(modifiers & ts.ModifierFlags.Export);
} else {
    isExported = isExported || !!(modifiers & ts.ModifierFlags.Export);
}
```

makana:
Names of types (classes, interfaces, enums) can be used as a namespace as well to support some sort of nested types.

```typescript
    /**
     * TestClass class comment.
     */
    export class TestClass
    {
        /**
         * This is a property.
         */
        public property = 'property';
        
        /**
         * This is a method.
         */
        public method(): void
        {

        }
    }

    /**
     * TestClass module comment.
     */
    export module TestClass
    {
        /**
         * This is an exported variable.
         */
        export const exportedVar = 'exported';

        /**
         * This is an internal variable.
         */
        const internalVar = 'internal';

        /**
         * This is an internal function.
         */
        function internalFunction(): void
        {

        }
    }
```

Variables or functions defined in this namespace are no longer exported unless the export keyword is written.

fixes #392 

This PR does not address the issue that those variables or functions actually are generated as static properties or methods. Anyhow I would start a discussion on this topic. How to deal with those nested types? Wouldn't it be better to leave them variables/functions? Or to expand a "Module" reflection kind child inside the class?


src/lib/converter/factories/declaration.ts
```diff
@@ -45,7 +63,8 @@ export function createDeclaration(context: Context, node: ts.Node, kind: Reflect
 
     // Test whether the node is exported
     let isExported: boolean;
-    if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])) {
+    if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule]) || considerInheritedExportKinds.indexOf(kind) === -1
+        || node.kind === ts.SyntaxKind.VariableDeclaration || node.kind === ts.SyntaxKind.FunctionDeclaration) {
```
aciccarello: Can you help me understand this line a bit? Part of the confusion is the existing code but I'm not following why this logic is added here.

From what I can see these lines are turning `isExported` off not just for modules but also for kinds that inherit their export status and for variable declarations and for function declarations. Then below `isExported` can be marked true by a modifier flag.

Because we are reassigning the `isExported` flag multiple times it is unclear what each step is checking. Can you give a bit higher level description of the logic here?

src/test/converter/nested-modules/nested-modules.ts
```diff
@@ -0,0 +1,45 @@
...
+         * This is a method.
+         */
+        public method(): void
+        {
+
+        }
+    }
+
+    /**
+     * TestClass module comment.
```
aciccarello: What happens if this comment is different than the one on the class? Could we also add an example of merging properties (between the class and the module) just so it is clear how that is handled?

src/test/converter/nested-modules/nested-modules.ts
```diff
@@ -0,0 +1,45 @@
+module TestModule
```
aciccarello: :+1: for adding a test case. Doesn't TypeScript recommend using the `namespace` keyword now? My understanding was that those two keywords were aliases for each other.

src/lib/converter/factories/declaration.ts
makana: Thanks for your detailed reply.

I was unsure about removing the original line ``container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])``. I didn't want to destroy something but I think it can be removed.

The code is directly related to #472 (see discussion there). If you have merged types/namespaces the container kind sometimes isn't a module but the type (class/interface...). So any sub-declaration inherits the class's export state. But that's not true for variables/functions that are defined within the merged namespace.

So my point here is that it is in fact the declaration's kind that specifies if the exported state could be inherited, not the container's kind.

The second line (VariableDeclaration/FunctionDeclaration) addresses the issue that variables/functions inside the namespace are treated as static properties/methods (kind is Property/Method that actually would inherit the export state of the class).

See [converter/nodes/variable.ts](https://github.com/TypeStrong/typedoc/blob/3102b2b2ce6936410c05a56f517b1f3fa96e4a5f/src/lib/converter/nodes/variable.ts#L62) or [converter/nodes/function.ts](https://github.com/TypeStrong/typedoc/blob/3102b2b2ce6936410c05a56f517b1f3fa96e4a5f/src/lib/converter/nodes/function.ts#L29)

Generally this part of the code is pretty confusing (I came upon it as well) and surely could be done better. But as far as I can say the last reassignment just handles the "export" keyword, so if it is explicitely exported, the element is in fact exported.

src/test/converter/nested-modules/nested-modules.ts
makana: Some existing test cases use the "module" keyword. And no linter argued ;) So...

According to [TypeScript](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html) it is distinguished between internal and external modules.

So I would say if module system is "none", everything should be "namespace", although it makes no real difference in output.

src/test/converter/nested-modules/nested-modules.ts
makana: As far as I could see, both comments are joined together (by a line feed) as "shortText" of the comment block.

I did not check how further paragraphs are handeled. I guess they are joined as well.

It's all about "how to handle type and namespace merging" (#472).

But apart from that I think it is not good style to document the namespace here since almost all it is good for is to expose nested types (as nested classes in C# for example, and here you don't have any further element besides the outer class to attach a documentation).

src/lib/converter/factories/declaration.ts
```diff
@@ -45,8 +63,16 @@ export function createDeclaration(context: Context, node: ts.Node, kind: Reflect
 
     // Test whether the node is exported
     let isExported: boolean;
-    if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])) {
-        isExported = false; // Don't inherit exported state in modules and namespaces
+    if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])
+        || ((considerInheritedExportKinds.indexOf(kind) === -1
+            || node.kind === ts.SyntaxKind.VariableDeclaration || node.kind === ts.SyntaxKind.FunctionDeclaration)
+        && !container.kindOf([ReflectionKind.ObjectLiteral, ReflectionKind.TypeLiteral]))) {
```
aciccarello: This logic is confusing, can it be simplified? I wouldn't know how to fix this if there was a problem.

makana: @aciccarello 

I am not sure about the meaning of "exported" here.

Considering the following source file that is a real-life-example (comments removed):
```typescript
namespace IduIT.Core.Geometry
{
	export enum GeometryTypeNames
	{
		EmptyGeometry = 0,
		Point = 1,
		MultiPoint = 2,
		LineString = 3,
		MultiLineString = 4,
		Polygon = 5,
		MultiPolygon = 6,
		Circle = 7,
	}
	export namespace GeometryTypeNames
	{
		export type TName = 'EmptyGeometry' | 'Point' | 'MultiPoint' | 'LineString' | 'MultiLineString' | 'Polygon' | 'MultiPolygon' | 'Circle';
		export const $isFlags = false;
		const ___labels: { [index: number]: string } = {
			[GeometryTypeNames.EmptyGeometry]: "Leer",
			[GeometryTypeNames.Point]: "Einzel-Punkt (Point)",
			[GeometryTypeNames.MultiPoint]: "Mehrere Punkte (MultiPoint)",
			[GeometryTypeNames.LineString]: "Linie (LineString)",
			[GeometryTypeNames.MultiLineString]: "Mehrere Linien (MultiLineString)",
			[GeometryTypeNames.Polygon]: "Fläche (Polygon)",
			[GeometryTypeNames.MultiPolygon]: "Mehrere Flächen (MultiPolygon)",
			[GeometryTypeNames.Circle]: "Kreis (Circle)",
		}
		export function getLabel(v: GeometryTypeNames): string
		{
			return ___labels[v] || GeometryTypeNames[v];
		}
	}
}
```

Here everything except the "___labels" object is exported.

But as the enum GeometryTypeNames is merges with the namespace GeometryTypeNames without this PR everything (except direct children of modules) inherit the exported state of its parent (``container.flags.isExported``), so would ___labels. This is wrong.

I don't like this part of code, maybe it could be simplieifed, but it works as far as I can see and consideres the following things:

- if anything could inherit the containers isExported flag, it just could be
  - properties, accessors and methods of a class or interface.
  - signatures and parameters of its function/method
  - index signatures
  - sub properties of an exported type or object literal
- the exported flag can explicitely not be inherited for variables/consts, functions, classes, interfaces, namespaces.
- the "excludeNotExported" option

I think the main issue is, that currently classes and namespaces are merged together to a single reflection model and variables and functions within the namespace are treated as static properties or methods of the class.

See src/lib/converter/nodes/variable.ts
```typescript
const kind = scope.kind & ReflectionKind.ClassOrInterface ? ReflectionKind.Property : ReflectionKind.Variable;
const variable = createDeclaration(context, node, kind, name);
```
I think this should be refactored, keeping variables variables and functions functions.

aciccarello:
@makana Thanks for your additional comments. I think I've wrapped my head around the problem. So the [changes to the enum spec](https://github.com/TypeStrong/typedoc/pull/615/files/df9dfa9f7de19c05f0efb9f4689d5d7463d6ed57#diff-4e309f777ca92482dc6b17ea4547ad3c) are what we want to be aiming for but I'm not sure about the current approach.

The problem seems to stem from the merging of containers with the namespace. Instead of checking for variables and functions, the logic should be able to recognize if the symbol was originally defined in a module.

PR #621, opened 2017-10-12
Provide typing information for destructuring parameters

Maximkaaa:
Currently if a parameter of function or constructor is a distructured object, the type of the parameter is always rendered as "object", even if the type is explicetely stated. My use case for typed destructured parameters is like this:

```
interface InitializationParameters {
  param: string;
}

class MyClass {
  constructor({param = "defaultValue"}: InitializationParameters) { ... }
}
```

After converting this code to reflections, there is no typing information for "__namedParameters" arguments, so it's impossible to show argument type with doc template. This is how the result looks currently (it is simple example from the test case added in this PR):

```
/**
 * A function with a destructuring argument that implements interface.
 */
export function functionWithTypedDestr({name}: classes.INameInterface) { }
```

typed_desctr

This PR fixes the lack of typing information of destructured argument. A follow up default themes PR actually uses this information to show the type.

Unfortunately this PR does not fix the same issue when type alias is used instead of interface for type definition. But since recomended way in this case is using interface, this solution should suffice for a while.

### DIFF
--- a/examples/basic/src/functions.ts
+++ b/examples/basic/src/functions.ts
@@ -189,3 +189,8 @@ export function createSomething() {
  * Taken from http://usejsdoc.org/tags-inline-link.html.
  */
 export function functionWithDocLink():void { }
+
+/**
+ * A function with a destructuring argument that implements interface.
+ */
+export function functionWithTypedDestr({name}: classes.INameInterface) { }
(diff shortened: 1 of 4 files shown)


Maximkaaa:
Default theme PR: https://github.com/TypeStrong/typedoc-default-themes/pull/51

brendo:
This would be extremely useful, almost all projects I work with use "object bags" and destructuring for methods, so it's a little disheartening having to dive back into the code to understand further.

Issue #585, opened 2017-08-22
Show in the generated documentation which properties are readonly.


lddubeau:
I noticed while working on #584 that the fact that a property is marked `readonly` does not carry to the generated documentation. At first I thought it was just a matter of modifying the templates but now I think that the information is lost along the way from input to output. At first glance, I'm thinking that [`ReflectionFlag`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/models/reflections/abstract.ts#L68) and associated types would need to be extended to record the information and modifications made to record the `readonly`-ness and to act on it when producing the output.

I'd like `readonly` properties to be immediately recognizable and so would have a circle marked in some way to indicate `readonly`-ness. Is `typedoc` getting its icons from some already existing convention from which I'd be able to find something for `readonly` properties or are the icons made up specially for `typedoc`?

I may find time to produce a PR but I'd like someone in the know reviewing my thinking here before I start so that any caveats can be addressed from the get go.

PR #623, opened 2017-10-13
Added "Readonly" flag for readonly properties

makana:
Readonly reflection flag for readonly properties is added.

This PR is related to #611 

The way, the abstract flag is added to the reflection is changed. It fixes the issue that abstract accessors weren't considered so far.

To extend the discusson on #611:
Does it make sense to remove the Const flag in favor of the Readonly flag?


makana:
Const flag is added to const enumerations.

aciccarello:
Thanks for the PR. This would close #585

shlomiassaf:
@makana
TypeScript has a clear separation between `Const` and `Readonly` as so it is better that we follow their logic. When we are aligned with the language it's easier to understand the code and less chance of getting surprised.

When creating the UI it's easier as well, you want to show readonly on classes and const on variables and if they are consolidated you have to do some logic which is not optimal.

It's better to stick to the same path as TS and not create local terminology...

makana:
@shlomiassaf 
Agreed. At least with my last commit (add const flag to enumerations) removing the const flag option is obsolete.

@aciccarello 
Since the flag is added to the "relevantFlags" array, it will be automatically rendered in the default themes.

Gerrit0:
#1268 made these changes and has been merged in.

PR #625, opened 2017-10-16
Include ambient types (declare keyword) as exported types.

makana:
``Ambient`` reflection flag is added to reflections that are explicitely declared using the ``declare`` keyword.

This isn't only possible in declarations (*.d.ts) but also in source files.

The context's ``isDeclaration`` flag is pushed onto the scope stack to keep track of ambient types. All types inside an isDeclaration context are exported.


Gerrit0:
I did some testing and have concluded that the current behavior is correct. `declare type` does not export that type. Trying to import it from another file and use it doesn't work. The only case when it is available is when the file is not a module (no import/export) in the first place (in which case it still isn't exported from some scope, everything is global)

Sorry this took so long to get to, if I'm wrong about the behavior please feel free to open an issue with some code samples showing me why :)

Gerrit0:
Coming back to this, I was partially correct. Just because something is `declared` doesn't mean it is exported, but some declared symbols without an export may be "exported" so long as they appear in a global file with no imports/exports

PR #632, opened 2017-10-23
refactor `removeReflection` + future work notes

shlomiassaf:
- set `removeReflection` as core operation
  - allow auto-removal of stale parents when running `removeReflection`

This PR try's to normalize the `removeReflection` operation and add it as a core method for `Context` instead of it being part of `CommentPlugin` which makes no sense.

Because `Context` has a `registerReflection` method it makes sense to have a `removeReflection` method as well.

**Additionally**
An optional parameter has been added to `removeReflection` called **removeStaleParent**.
When set to true it will automatically remove stale parents.
A stale parent is a `DeclarationReflection` that is invalid without child reflections.
When a reflection is removed and `removeStaleParent` is true the parent is checked to see if it is invalid (stale), if so it is removed.

#### Examples:
  - Removing a reflection of kind `ReflectionKind.GetSignature` or `ReflectionKind.SetSignature`
will remove the signature from the parent, `ReflectionKind.Accessor`. If the removal
leaves the parent without a signature (get or set) it is stale/invalid thus requires removal as well.

  - Removing a reflection of kind`ReflectionKind.CallSignature` or `ReflectionKind.ConstructorSignature`
will remove the signature from the parent, function, method or constructor.
If the removal leaves the parent without a single signature it is stale/invalid thus
requires removal as well.

## Notes:
  - CommentPlugin#removeReflection` is still supported but with a `@deprecated` flag.
  - The added optional parameter `removeStaleParent` does not break API as it is optional and off by default (same behaviour unless opt-in)

# Future work:
This is the first step towards a safe reflection removal operation, further fixes are required to make it safe.

The main issue i've identified is that **plugins are not aware of a reflection removal operation (they are aware of adding)**

This has clear impact with the following:
  1. groups added by the `GroupPlugin` 
  2. `ReferenceType` instance.

(1) `GroupPlugin` set reflections in the `groups` property, if a reflection is removed and it's a child of a group it does not get removed in the group which leaves a stale reflection.
To solve this, each project needs to have a list of all the removed reflections so at the end the plugin can apply changes.

(2) Some of the converters as well as the `TypePlugin` create `ReferenceType` instances to reflections (`type`, `overwrites`, `inheritedFrom`, `extendedTypes`, `extendedBy`, etc.). When a reflection is removed they are unaware and so leave the reference in place.
To solve this, we can use the solution for (1) and traverse all reflection to remove types with references to removed reflections.

The solutions might not be optimal as they require full traversal of all reflections and perhaps someone can find a better idea.
A remove event will not help because the removed reflection has no reference to reflections/types pointing at it....

### DIFF
--- a/src/lib/converter/context.ts
+++ b/src/lib/converter/context.ts
@@ -2,7 +2,18 @@ import * as ts from 'typescript';
 import { Minimatch, IMinimatch } from 'minimatch';
 
 import { Logger } from '../utils/loggers';
-import { Reflection, ProjectReflection, ContainerReflection, Type } from '../models/index';
+import {
+    ContainerReflection,
+    DeclarationReflection,
+    IntrinsicType,
+    Reflection,
+    SignatureReflection,
+    ParameterReflection,
+    ProjectReflection,
+    TypeParameterReflection,
+    TraverseProperty,
+    Type
+} from '../models/index';
 import { createTypeParameter } from './factories/type-parameter';
 import { Converter } from './converter';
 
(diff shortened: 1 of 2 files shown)


shlomiassaf:
@aciccarello @blakeembrey any suggestions/ideas on the required future work?

jonchardy:
@shlomiassaf @aciccarello Also see https://github.com/TypeStrong/typedoc/issues/877. Maybe we could modify this PR to also trigger events, just for convenience? I also think it's important to properly clean out references when removing a reflection.

Issue #187, opened 2016-02-03
decorators in code blocks are being treated as tags


Necroskillz:
Given this doc:

``````
/**
 * Some Description
 *
 * ```
 * @RouteConfig([
 *   { path: '/about', component: AboutComponent, name: 'About' }
 * ])
 * ```
 */
``````

Output looks like this:
![image](https://cloud.githubusercontent.com/assets/346477/12790209/7a294f00-caa1-11e5-8285-5f431645132a.png)

Necroskillz:
I fixed it for now by inserting invisible space after the @ sign.

http://www.fileformat.info/info/unicode/char/200B/browsertest.htm

mattflix:
I encountered this behavior too.

The comment parser should not be looking for doc tags inside the code block maybe?

aciccarello:
Related to #97

aciccarello:
Should be fixed when #633 is released

PR #633, opened 2017-10-23
dont parse comment for tags inside a code block

shlomiassaf:
This PR applies markdown code block areas and prevents parsing for tags in them.

The following comment
```
/**
 * A Comment for a class
 *
 * ## Some Markup
 * **with more markup**
 *
 * A example with decorators that should not parse to tag
 * ```
 * @myDecorator
 * @FactoryDecorator('a', 'b', 'c')
 * export class CommentedClass {
 *   myProp: string = 'myProp';
 *
 *   @PropDecorator() decoratedProp: string;
 *
 *   constructor(@ParamDecorator public param: string) { }
 *
 *   myMethod() { }
 * }
 * ```
 * @deprecated
 * @todo something
 */
```

Will result in 2 tags parsed, `deprecated` and `todo`

Before this commit, 3 more tags are parsed: `myDecorator`, `FactoryDecorator` and `PropDecorator`

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -160,16 +160,23 @@ export function parseComment(text: string, comment: Comment = new Comment()): Co
         comment.tags.push(currentTag);
     }
 
+    const CODE_FENCE = /^\s*```(?!.*```)/;
+    let inCode = false;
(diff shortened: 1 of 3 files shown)


aciccarello:
Thanks! This would close #187

PR #638, opened 2017-10-27
Add two plugins

christopherthielen:
Adds two plugins to the list:
- Internal/External Module
- Single Line Tags


aciccarello:
Thanks!

PR #640, opened 2017-10-30
make `traverse` safe for delete operations

shlomiassaf:
Currently, traverse does not play nice when mutating the children, signature, etc, of a reflection while traversing.

A current issue happens silently [here](https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/plugins/CommentPlugin.ts#L319)

Every child is removed, but the iteration is done on the children collection:
```ts
    traverse(callback: TraverseCallback) {
        if (this.children) {
            this.children.forEach((child: DeclarationReflection) => {
                callback(child, TraverseProperty.Children);
            });
        }
    }
```
So the first child with **index 0** is removed, then the index increments to 1 but because we altered the collection index 1 points to what used to be in index 2, i.e. we skipped an item.

The fix is simple, when working with collection work on a copy.
```ts
    traverse(callback: TraverseCallback) {
        if (this.children) {
            this.children.slice().forEach((child: DeclarationReflection) => {
                callback(child, TraverseProperty.Children);
            });
        }
    }
```

### DIFF
--- a/src/lib/models/reflections/container.ts
+++ b/src/lib/models/reflections/container.ts
@@ -46,7 +46,7 @@ export class ContainerReflection extends Reflection {
      */
     traverse(callback: TraverseCallback) {
         if (this.children) {
-            this.children.forEach((child: DeclarationReflection) => {
+            this.children.slice().forEach((child: DeclarationReflection) => {
                 callback(child, TraverseProperty.Children);
             });
         }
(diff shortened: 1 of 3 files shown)


shlomiassaf:
@aciccarello @blakeembrey can we **kindly :)** speed up PR merges?
I'm starting to pile up PR's and it creates a backlog 👍 
thanks!

aciccarello:
@shlomiassaf Thanks for the PRs. I'm really glad someone is making improvements. Sorry things are a bit behind. I don't have the time or knowledge of the codebase to process things very quickly. I'll see if I can get through the other 3 PRs sometime soon. Let me know if there is a priority. I'd guess #597 is the main blocker.

I'm also usually available on gitter if you need to discuss something.

Issue #516, opened 2017-05-25
Document imported types


Gozala:
Right now if I have code like this:

```ts
import {Maybe} from "maybe.ts"

export const toMaybe =
  <x, a>(result: Result<x, a>): Maybe<a> =>
    result.isOk ? result.value : null

export type Result<x, a> = ...
```

typedoc generates docs such that `Maybe<a>` is completely undocumented.

<img width="498" alt="screen shot 2017-05-25 at 12 05 55 pm" src="https://cloud.githubusercontent.com/assets/21236/26466281/0a9d1350-4143-11e7-8dc2-b9c178de98a9.png">

It would be great to have an option to either generate docs for imports as well or alternatively have a way to provide external links.

vekexasia:
:+1: for this

Gerrit0:
In 0.20, TypeDoc will document whatever you export. Changing the code to:

```ts
import {Maybe} from "maybe.ts"

export * as MaybeTS from "maybe.ts"

export const toMaybe =
  <x, a>(result: Result<x, a>): Maybe<a> =>
    result.isOk ? result.value : null
export type Result<x, a> = { isOk: true, value: a } | { isOk: false, err: x }
```

Will result in documentation like this, with links going where you would expect:

![image](https://user-images.githubusercontent.com/19329837/100523572-afc4b580-316e-11eb-9e3f-f6fe3c37a306.png)

Issue #498, opened 2017-05-11
Type Alias hyperlinks do not work


yohei1126:
The type of the parameter `callback: callbacks.Callback` on setCallback  is documented as type `Callback` but no link to the type alias `Callback` is created. 
Here is my code.
https://gist.github.com/yohei1126/238650e60378e669b86cf19debf29abc

src/Callbacks.ts
```
export type Callback = () => void;
```

src/Sample.ts
```
import * as callbacks from "./Callbacks";
export class Sample {
    private callback: callbacks.Callback;
    setCallback(callback: callbacks.Callback): void {
        this.callback = callback;
    }
}
```

Issue #435, opened 2017-02-24
Import aliases don't work


asakusuma:
```JavaScript
import { Foo as MyType } from './foo';
export function bar(foo: MyType) {}
```

The docs don't recognize that `foo` is an instance of `Foo`. There is no link from `MyType` to `Foo`, and if you export the docs as JSON, the parameter `foo` is missing the reference id, unlike parameters who's types are not an alias.

aciccarello:
Thanks for the report! I wonder if this is at all related to #393 since it involves renaming symbols.

Issue #428, opened 2017-02-20
Interface doesn't work with import or reference


WahlbergRu:
Hello. I use 

`
import {ICallbackRow, IwidgetSettings, Iinvoke} from "./types";
`

or 

`
/// <reference path="./types.d.ts"/>
`

but this doesn't compile in documentation. 

But if I copy interface in file, then interface compile working. 

Where I have error?

aciccarello:
Hello @WahlbergRu, what version of typedoc are you using?

PR #646, opened 2017-11-10
Document and resolve imported types.

makana:
Import declarations of any kind are included into the docs. Used imported types are resolved as long as they are part of the source files that are documented.

Commit 01f040c includes all the logic.

Commit ad39cce modifies the theme builder to point the URLs of import declarations to their original exported declaration. I am not sure if it is the intended behavior as I only use the JSON output but I tried a few things with the default theme and it makes more sense to me to do so.

Related to #516 #498 #435 #428 #159 and maybe fixes all of them.


fd:
@makana what is the release timeline for this PR?

Gerrit0:
#1157 has been merged with support for handling imported types.

Thank you for the PR! It was a lot of help in determining how best to handle imported types.

PR #654, opened 2017-12-04
Update to latest Typescript version

Gerrit0:
Simple enough - I just fixed the build errors.


src/lib/utils/options/sources/component.ts
```diff
@@ -33,24 +33,6 @@ export class ComponentSource extends OptionsComponent {
             }
         }
     }
-
-    private removeComponent(component: AbstractComponent<any>) {
```
aciccarello: Why was this method removed?

Gerrit0: It was breaking the build as it was defined, but never used: from #641:

```text
src/lib/utils/options/sources/component.ts(37,13): error TS6133: 'removeComponent' is declared but its value is never read.
```

aciccarello: Ah, I see. I probably should review this class to ensure that there is not an implementation bug here before I merge the PR.

aciccarello: Looks like `removeComponent` should have been called in `onComponentRemoved`. I probably should put together a separate PR and do some testing. Looks like the minimal theme triggers removing components.

https://github.com/TypeStrong/typedoc/blob/6edcbd53e0c85a063bb91cb6732ced7c21e44007/src/lib/output/themes/MinimalTheme.ts#L21-L24

There also looks to be some bugs in `removeComponent` (i.e. `slice` on this array will do nothing).

https://github.com/TypeStrong/typedoc/blob/6edcbd53e0c85a063bb91cb6732ced7c21e44007/src/lib/utils/options/sources/component.ts#L40-L42

RangerMauve:
What's blocking this PR? Is there anything I could do to help move it forward?

Gerrit0:
I believe now that #664 has been merged, this should be good to go. Unfortunately since the tests on the project right now are completely broken (no useful info if something changes) it's dangerous to merge anything.

RangerMauve:
That's a shame. For now I'm using your fork directly for my project and it appears to be working fine. I don't have a particularly fancy use-case, though.

aciccarello:
There aren't any major blockers right now but I'll want to merge a couple more PRs and do a big dependency update before a release.

bilby91:
Any update on this?

aciccarello:
@Gerrit0 I didn't merge this PR but I wanted to say thanks for figuring out what needed to be changed :heart:

PR #682, opened 2018-01-22
Fix --plugin argument parsing

tomdale:
Currently, the `--plugin` argument simply doesn't work because it tries to treat the provided string as an array. For example, if I pass `--plugin foo`, it will try to load three npm packages named `f`, `o`, and `o` respectively.

This commit changes the option declaration to use the `ParameterType.Array` type, which parses a comma-separated string into an array. It also removes the otherwise-unused `isArray` property from the `DeclarationOptions` interface.

Despite the, err… "enterprise-grade" architecture of Typedoc, I had trouble writing a good test for this. Even with the hyper-abstraction of `Component`s, it's impossible to construct an `Application` without kicking off side-effects in the constructor, and it's non-trivial to instantiate and test other classes (like `PluginHost`) without an `Application`, due to lots of implicit coupling.

That said, I did manage to write a test for this, but it does instantiate an app to do so, which actually tries to load the non-existent npm packages specified in the test. Ideally, it would be possible to test the options declarations of components without having to run the application.

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -26,7 +26,6 @@ export interface DeclarationOption {
     scope?: ParameterScope;
     map?: {};
     mapError?: string;
-    isArray?: boolean;
     defaultValue?: any;
     convert?: (param: OptionDeclaration, value?: any) => any;
 }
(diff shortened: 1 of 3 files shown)


tomdale:
The build is failing due to the issue resolved in #683.

tomdale:
This is passing now after rebasing against `master` with #683 merged.

aciccarello:
Thanks @tomdale!

PR #683, opened 2018-01-22
Upgrade tslint to 5.9.1

tomdale:
A minor release of tslint is currently causing the Typedoc build to fail. The failure is caused by a malformed configuration value that was being silently swallowed before, but causes an exception in tslint 5.9.0 and later.

This PR re-enabled the malformed rule (`max-line-length`) and updates the minimum version of tslint to 5.9.0. To avoid a mega-PR I bumped the maximum line length from 120 to 170, as many lines beyond 120 characters have snuck in since this stopped working.

Given how frustrating it can be for new contributors for builds/tests to fail on a fresh clone, would you be open to including the `yarn.lock` or `package-lock.json` files that are currently in `.gitignore`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "grunt-tslint": "^5.0.1",
     "istanbul": "^0.4.1",
     "mocha": "^4.0.0",
-    "tslint": "^5.1.0"
+    "tslint": "^5.9.1"
   },
   "files": [
     "bin",
--- a/src/lib/converter/types/array.ts
+++ b/src/lib/converter/types/array.ts
@@ -18,7 +18,12 @@ export class ArrayConverter extends ConverterTypeComponent implements TypeConver
      */
     supportsType(context: Context, type: ts.TypeReference): boolean {
         // Is there a better way to detect the {"type":"reference","name":"Array","typeArguments":{...}} types that are in fact arrays?
-        return !!(type.flags & ts.TypeFlags.Object) && !!type.symbol && type.symbol.name === 'Array' && !type.symbol.parent && !!type.typeArguments && type.typeArguments.length === 1;
+        return !!(type.flags & ts.TypeFlags.Object)
+          && !!type.symbol
+          && type.symbol.name === 'Array'
+          && !type.symbol.parent
+          && !!type.typeArguments
+          && type.typeArguments.length === 1;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


aciccarello:
Thanks @tomdale. From the limited experience with lock files I've had I haven't found it particularly helpful. I'd be open to adding the `package-lock.json` file though. Can you help me understand the benefits?

tomdale:
@aciccarello Thanks for taking a look at this. The benefit of a lock file is that they encode the set of dependencies used at a particular moment in time, that can be deterministically reproduced in the future.

Let's use this case as an example. As a first time contributor, it was very frustrating to clone TypeDoc from GitHub and have the test suite fail. I wasted a lot of time assuming it was something wrong with my environment, or because I hadn't followed instructions correctly. A lockfile would have prevented this.

The root cause is that the `package.json` specified a version range of  `^5.1.0` for `tslint`, meaning any version equal to or above 5.1.0, but less than 6.0.0—in other words, declaring `tslint` "safe" to upgrade in the 5.x series because it follows Semantic Versioning.

I'm a big fan of SemVer, but particularly in dynamic languages like JavaScript, mistakes can happen. In this case, `tslint@5.9.0` introduced a change that made a previously-ignored malformed configuration value turn into a hard error.

At some point, the set of npm packages on disk was working for a maintainer. But because dependencies release new versions (and their transitive dependencies also release new versions), the set of dependencies I got after cloning and `npm install`ing was different than the set of dependencies _you_ had when you last had a working system.

This is what a lockfile encodes—"here is a set of dependencies that are known to work." The `package.json` asserts what _should_ work; the lockfile asserts what _is known_ to work. The difference may seem subtle, but I hope it's obvious why having a known-good set of dependencies to compare against makes it much easier to debug issues like this.

In this case, I was lucky because it was a single dependency that caused the tests to fail. But I have run into more complex situations in the past where the interaction between two incompatible dependencies caused errors, and if all I want to do is work on a bugfix, having to recreate a working set of dependencies is always time-consuming and sometimes effectively impossible.

*TL;DR*—lockfiles allow you to keep the project working for new contributors, even if incompatible dependencies have been released in the meantime that break the build. They allow you to be intentional about when dependencies get upgraded (and work really well in conjunction with Greenkeeper, which I see is already set up for this repo—awesome!).

I hope this explanation helps. If you have additional questions or concerns, I'm happy to answer them. Lockfiles really have been a godsend for the other open source projects I work on, so consider me just a satisfied customer. :)

Issue #655, opened 2017-12-04
Upgrade TS version


chriskrycho:
Since TypeDoc hooks into the TS version, it can be quite frustrating to use when it's behind by several versions.

I have it running as a pre-commit hook via [husky](https://github.com/typicode/husky), and it's currently failing because I have intentionally (for the sake of the shape of my exports in the `index.ts`) imported some unused items. In TS 2.6 I can ignore those unused items by doing `// @ts-ignore` on the preceding line; but TypeDoc being only at 2.4 can't do that.

Related: is the underlying API being used so unstable that TypeDoc can't treat TS as a peer dependency rather than a direct dependency? It would be great if it was a peer dep with a version requirement for >= whatever the minimum version it needs is.

aciccarello:
@chriskrycho, PR #654 should update the TS version in master. Once that is merged and a release is made the `// @ts-ignore` issue should be resolved.

As to why we can't use peer dependencies, TypeDoc currently relies on some internal TS APIs so it is very unstable. It'd be great if the project could move away from those but that seems like a significant undertaking.

Related #369

chriskrycho:
Makes sense. Thanks!

PR #685, opened 2018-01-23
Typescript Dep change

dfenstermaker:
I am having an issue with the current version in the npm repository. Because there is a version of typescript local to the module, I can't take advantage of the version of typescript I am using. This is breaking my builds. When I remove the local version and fall back to my version of typescript it works great. I removed typescript as a local dep and made it a dev dependency so it installs when developing, then I made it a peer dependency so that it tells those installing it needs a version of typescript installed.

Lastly I threw in a editor config so that no one else contributing has to go back and reformat cause they don't have their editor setup right, like me :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = 120
+trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


Gerrit0:
See #655 as to why this isn't a safe thing to do - Typedoc relies on internal APIs which are unstable and may change even between minor Typescript versions.

dfenstermaker:
@Gerrit0 makes sense. However, peerDependency would allow for others to try out the latest version of Typescript and see if it still works without waiting on maintainers to do so. The peer dependency could recommend a working version while also allowing the flexibility to try and use another. If the consumer upgrades and breaks then they know it doesn't work for them and that they need to adhere to the recommended version set in the peer dep.

aciccarello:
I'm starting to warm up to the peerDependency idea so long as we make it very clear that it is only guaranteed to work with that exact version. I probably want to wait for a major release to make that change though.

Thanks for the discussion.

PR #687, opened 2018-01-31
fix typedoc for TypeScript 2.7-RC

enko:
All I had to do, was change `ts.Diagnostic[]` to `ReadonlyArray<ts.Diagnostic>`. I tried the build with my private project and seemed fine. The tests went all without a hitch. I had to remove the abstract keyword from the `BaseClass` from the basic examples, because this does not seem to be allowed any more.


aciccarello:
@enko I didn't merge this PR but I wanted to say thanks for figuring out what needed to be changed :heart:

enko:
@aciccarello No Problem at all :-)

I'm glad the dependencies are updated.

PR #705, opened 2018-02-23
Don't write BOM for search index file

bwasty:
The search doesn't work when `search.js` has a BOM - loading it fails with 
```
Uncaught SyntaxError: Invalid or unexpected token
```
See also https://github.com/cginternals/webgl-operate/issues/22.


iamam34:
I've just encountered this issue too, keen for a merge. :+1:

PR #711, opened 2018-03-01
Allow exclusion of directories starting with a dot

### LINKED ISSUE #710: How to exclude .git directory?
I want to generate documentation starting from the root directory of my project. I'm currently using:

    typedoc --out ./docs --excludeExternals --externalPattern "**/node_modules/**" --ignoreCompilerErrors --exclude "**/{__mocks__,\.git,__tests__}/**" --name bitname .

However, .ts files in the git heads are still included in the generated documentation. It's the same without the escape string. Is there any way to achieve this behavior?

colatkinson:
Resolves #710 

Essentially just adds a test for this behavior, and sets `{dot: true}` for the exclude file `Minimatch` instance.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -248,7 +248,7 @@ export class Application extends ChildableComponent<Application, AbstractCompone
      */
     public expandInputFiles(inputFiles?: string[]): string[] {
         let files: string[] = [];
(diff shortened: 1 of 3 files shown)


aciccarello:
Thanks for the PR @colatkinson! Sorry I took so long to merge.

PR #714, opened 2018-03-03
Bug Fix: Allow deferring file inclusion to tsconfig.json

fabioberger:
Currently you **must** specify a path to your typescript project at the end of the `typedoc` command. Many projects already define `include` in their `tsconfig.json` files, describing which files they wish included on TS compilation. Fixing this bug allows those configs to be used rather then specifying them again from the command-line.

Example `include` in `tsconfig.json` taken from [0x.js](https://github.com/0xProject/0x-monorepo/tree/development/packages/0x.js):

```
"include": [
        "./src/**/*",
        "./test/**/*",
        "../../node_modules/types-bn/index.d.ts",
        "../../node_modules/types-ethereumjs-util/index.d.ts",
        "../../node_modules/web3-typescript-typings/index.d.ts",
        "../../node_modules/chai-typescript-typings/index.d.ts",
        "../../node_modules/chai-as-promised-typescript-typings/index.d.ts"
    ]
```

I believe this is just a bug fix since the relevant conditional could never be `false` given that `files` is initialized to an empty array.


aciccarello:
Hello @fabioberger. I'm likely going to merge #610 which should fix this issue (see the change to [arguments.ts](https://github.com/TypeStrong/typedoc/pull/610/files#diff-87db7f7b6da1b1ed71d3ad841e6e33b8). Thanks for your work though. Feel free to create an issue or reach out on [gitter](https://gitter.im/TypeStrong/typedoc) so that we can make sure we aren't duplicating effort.

fabioberger:
Awesome, #610 should fix my issue as well. Np, hope to contribute more soon!

Issue #722, opened 2018-03-15
don't case file names to lowercase


dkent600:
Currently `Reflection.getAlias` cases filenames to lower case.  I don't know why this is, but for my needs it would be great if it didn't do this, and instead would leave the file names with the same case as the types they document.

Perhaps this could be an option?

Gerrit0:
Closing in favor of #1585

PR #723, opened 2018-03-15
don't cast files to lower case

dkent600:
Addresses: https://github.com/TypeStrong/typedoc/issues/722

Don't cast files to lower case.  This enables me to dynamically use the file names to describe what the files contain when I list them in an index on my documentation page.

This change also fixes three lint errors.

### DIFF
--- a/src/lib/models/reflections/abstract.ts
+++ b/src/lib/models/reflections/abstract.ts
@@ -434,12 +434,12 @@ export abstract class Reflection {
      */
     getAlias(): string {
         if (!this._alias) {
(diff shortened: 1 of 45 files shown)


dskvr:
+1

demurgos:
It should probably be attached to an option and documented.
Mixed case can be the source of many issues so I don't want the default behavior to change in a simple update. I don't mind if it's opt-in for those who want to use it.

This PR also changes so standard newlines `\n` to Windows style `\r\n`, it should be fixed.

PR #730, opened 2018-03-28
Add support for excluding directories and refactor `expandInputFiles`.

meowtec:
I am creating this pull request because I have the problem when using TypeDoc.

Command:

```bash
typedoc --exclude "**/node_modules/**" --ignoreCompilerErrors --out tsdoc .
```

Stderr:

```
ENAMETOOLONG: name too long, stat '/Users/myObject/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser/node_modules/@types/express/node_modules/@types/body-parser'
```

 It was due to a symbolic link recursion, caused by `npminstall`, a alternative `npm install` tool, which I have to use for some reasons.

This pull request make `isExcluded` checking for both directories and files. We can stop the file walking when the directory is excludes. It is also better for IO performance.

PR #731, opened 2018-03-28
Add realpath method to compiler host

meowtec:
I got a large number of errors after run `typedoc [myoptions]`, those errors are just like:
```
...
Error: /project/node_modules/@types/node/index.d.ts(6945)
 Cannot redeclare block-scoped variable 'HTTP_STATUS_UPGRADE_REQUIRED'.
Error: /project/node_modules/@types/node/index.d.ts(6946)
 Cannot redeclare block-scoped variable 'HTTP_STATUS_PRECONDITION_REQUIRED'.
...
```

But I can compile my project using `tsc` with no errors.

Even though I can add `--ignoreCompilerErrors` arg to the command, those errors always make me uncomfortably.

After debugging, I have found the reason: the `realpath()` method of `CompilerHost` class is missing.
And there is some  symbolic links in my `node_modules`. (also see #730)

This PR is to fix it by adding the `realpath` method.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -254,28 +254,22 @@ export class Application extends ChildableComponent<Application, AbstractCompone
             return exclude.some(mm => mm.match(fileName));
         }
 
-        function add(dirname: string) {
-            FS.readdirSync(dirname).forEach((file) => {
-                const realpath = Path.join(dirname, file);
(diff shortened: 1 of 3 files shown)


meowtec:
Just ignore this PR. I have created new one

PR #742, opened 2018-04-06
General correction of glob patterns and config file locating

Tokimon:
This PR main purpose is to optimize the options that uses glob patterns, as they were not working optimally, as well as making sure config files are found correctly.

(unit test added for the general Minimatch function)

**Glob pattern options**
`--externalPattern` and `--exclude` options now behaves according to expectations of many developers.
- _Relative paths_:  
`./my/path/**`, `../my/path/**` and `my/path/**` all correctly matches paths relative to the project root.
- _Absolute paths_:  
`/unix-style/path/**`, `C:/win-style/path/**` and `**/my/path/**` all still matches full paths as before.
- `{ dot: true }` has been added to the Minimatch as per PR #711 

**typedoc.js(on)**
- If no typedoc file is indicated with the `--options` option any `typedoc.js` or `typedoc.json` will automatically be searched for in the root of the project (with `typedoc.js` taking precedence).
- If a directory is defined in the `--options` option any `typedoc.js` or `typedoc.json` in the root of that directory will be searched for (with `typedoc.js` taking precedence). If no file is found in the directory an error is raised.
- Specifying the exact file is still viable and result in an error if not found

**tsconfig.json**
Original functionality has not been altered, but `--tsconfig` can now also take a directory path instead of only a direct `tsconfig.json` path.


aciccarello:
Looks great! Thanks for the contribution. I had started working on getting the typedoc.json file working again a while ago but never finished.

I'll take a little more time to look this over before merging but from what I see, it looks great!

package.json
```diff
@@ -33,34 +33,34 @@
     "@types/fs-extra": "5.0.1",
     "@types/handlebars": "4.0.36",
     "@types/highlight.js": "9.12.2",
-    "@types/lodash": "4.14.104",
+    "@types/lodash": "4.14.106",
```
aciccarello: Next time, avoid updating dependencies unnecessarially. It creates merge conflicts with other contributions.

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -65,14 +76,9 @@ export class TSConfigReader extends OptionsComponent {
...
     load(event: DiscoverEvent, fileName: string) {
-        if (!FS.existsSync(fileName)) {
-            event.addError('The tsconfig file %s does not exist.', fileName);
-            return;
-        }
-
         const { config } = ts.readConfigFile(fileName, ts.sys.readFile);
         if (config === undefined) {
-            event.addError('The tsconfig file %s does not contain valid JSON.', fileName);
+            event.addError('No valid tsconfig file found.', fileName);
```
aciccarello: I changed the text since it now is used when there is no file as well as an invalid file

src/lib/utils/paths.ts
```diff
@@ -0,0 +1,17 @@
...
+
+function createMinimatch(pattern: string): IMinimatch {
+  if (pattern[0] === '.' || pattern[0] === '/' || /^\w+(?!:)($|[/\\])/.test(pattern)) {
+    pattern = Path.resolve(pattern);
+  }
+
+  return new Minimatch(pattern.replace(/[\\]/g, '/'), { dot: true });
+}
+
+export function pathToMinimatch(pattern: string | string[]): IMinimatch | IMinimatch[] {
```
aciccarello: Is this function really needed? From the usage I saw, it seems that `createMinimatch` would be sufficient. The overloading of a parameter that is sometimes an array means that you frequently need to cast the result.

I'd prefer to inline the `arr.map(createMinimatch)` in files.

src/lib/utils/paths.ts
```diff
@@ -0,0 +1,17 @@
+import * as Path from 'path';
+import { isArray } from 'util';
+import { Minimatch, IMinimatch } from 'minimatch';
+
+function createMinimatch(pattern: string): IMinimatch {
+  if (pattern[0] === '.' || pattern[0] === '/' || /^\w+(?!:)($|[/\\])/.test(pattern)) {
```
aciccarello: Can you explain this regex? I don't have the skills to understand it.

Is there any of this that Minimatch handles already?

src/test/utils.paths.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import isEqual = require('lodash/isEqual');
+import Assert = require('assert');
+
+import { pathToMinimatch } from '..';
+
+// Used to ensure uniform path cross OS
+const absolutePath = (path: string) => Path.resolve(path).replace(/[\\]/g, '/');
+
+describe('Paths', () => {
```
aciccarello: Thanks for writing tests!

src/lib/utils/options/readers/tsconfig.ts
Tokimon: So the text `The tsconfig file %s does not contain valid JSON` is used for both?

package.json
Tokimon: ok got it... I can revert it

src/lib/utils/paths.ts
Tokimon: Ok so this `if` is to check if a path like these:
- `./some/path`
- `../some/path`
- `some/path`
- `path`

Which are all relative paths and should be parsed relative to `process.cwd()` (`Path.resolve(path)`).

The regex is to test that windows absolute paths like `C:/some/path` wont be confused for a relative path like `c/some/path`. So I use a negative look ahead to ensure there is no `:` after the letters in the first part of the path (`path` will be a relative path and `path:` won't)

I could have used one regex to test for all of these cases, but it is a lot faster to do the simple tests first and only do the regex if really needed.

The main issue is that generally in other libraries that uses GLOB expressions you can just type `./some/path` and that would be treated as relative to the project root, but this have not been the case for this project where `**/some/path` was needed to ensure correct path matching, which has been counter-intuitive for many developers.

Tokimon: But I can see I made a mistake: `pattern[0] === '/'` would match `/some/path` as relative, which isn't the case. I will fix this.

src/lib/utils/paths.ts
Tokimon: It is true that the two methods could be combined. I separated them in an attempt to make the code less bulky and easier to read. But if you prefer to combine them I can. It is also a bit slower to recreate a function inside a function, so there is also a minor performance gain.

Tokimon:
I created the general function for the minimatch to uniform the the way the minimatching is done (right now it is used in two places), so the change is basically to allow paths to be written as relative to the project root (`process.cwd()`) as is the norm in many other modules and it have confused several developers which have asked how to do the `exclude` for example only to find that they needed `**` in front of the path to make it work (my self included).

I will check up on the indentation as well.

src/lib/utils/paths.ts
Tokimon: regarding the `pattern: string | string[]` is because the returned path can be one or several paths. When only one path is given it is just a string otherwise an array of strings is given, hence the double type definition.

src/lib/utils/paths.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Minimatch, IMinimatch } from 'minimatch';
+
+function convertToMinimatch(pattern: string): IMinimatch {
+    // Ensure uniform absolute path cross OS
+    // (POSIX would resolve c:/path to /path/to/repo/c:/path without this check)
+    if (Path.isAbsolute(pattern) && Path.sep === '/') {
+        pattern = pattern.replace(/^\w:/, '');
+    }
+
+    if (pattern[0] !== '*') {
```
aciccarello: Should this be checking for `**`? If the path is not resolved for `*/test.js` then I think it won't work with `/project/base/*/test.js` which is what I would expect

src/lib/utils/paths.ts
```diff
@@ -0,0 +1,28 @@
...
+        pattern = Path.resolve(pattern);
+    }
+
+    // Unify the path slashes before creating the minimatch, for more relyable matching
+    return new Minimatch(pattern.replace(/[\\]/g, '/'), { dot: true });
+}
+
+export function createMinimatch(pattern: string[]): IMinimatch[];
+export function createMinimatch(pattern: string): IMinimatch;
+export function createMinimatch(pattern: string | string[]): IMinimatch | IMinimatch[] {
```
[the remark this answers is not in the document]
Tokimon: ok... if you prefer...

src/lib/utils/paths.ts
```diff
@@ -0,0 +1,28 @@
+import * as Path from 'path';
+import { isArray } from 'util';
+import { Minimatch, IMinimatch } from 'minimatch';
+
+function convertToMinimatch(pattern: string): IMinimatch {
+    // Ensure uniform absolute path cross OS
+    // (POSIX would resolve c:/path to /path/to/repo/c:/path without this check)
+    if (Path.isAbsolute(pattern) && Path.sep === '/') {
+        pattern = pattern.replace(/^\w:/, '');
```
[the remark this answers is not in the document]
Tokimon: Well i corrected this as it was a bit confusing (even for me). But as written in my tests this should now be correctly resolved according to the current project (with or without drive notation). I generally leverage the functionality  in `Path.resolve` which on Windows will resolve `/some/path`, by adding the drive letter of your current repo (eg. with a repo at `D:/my-repo/some/path` `Path.resolve('/some/path')` would return `D:/some/path`). There is no active validation whether the path actually exists or not here, but just a correction of the possible path.

However the drive letter removal is only necessary on systems using POSIX path system (everybody else but windows), hence the `Path.sep === '/'` check, as `C:/some/path` would otherwise resolve to `/myrepo/C:/some/path`

(you can try it for yourself by resolving paths using `Path.posix.resolve` or `Path.win32.resolve`)

src/lib/utils/paths.ts
Tokimon: hm... yeah you might be right... I will correct it.

src/lib/utils/paths.ts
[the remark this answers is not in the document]
aciccarello: okay, that sounds good to me

src/lib/utils/paths.ts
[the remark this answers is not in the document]
Tokimon: Just an update on the UNIX pathing system. I asked around and for UNIX systems (Linxus + MAC) you change drive by typing something like `/volumes/drive`. So absolute paths always start with a '/' on those systems. So it should be good like this.

Tokimon:
Sure thing! Thanks for your input.

Tokimon:
There we go. Tests are passing now.

Gerrit0:
Thanks for the contribution! I'm sorry this took so long to merge.

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -40,21 +40,32 @@ export class TSConfigReader extends OptionsComponent {
             return;
         }
 
+        let file: string | undefined;
+
         if (TSConfigReader.OPTIONS_KEY in event.data) {
-            this.load(event, Path.resolve(event.data[TSConfigReader.OPTIONS_KEY]));
+            const tsconfig = event.data[TSConfigReader.OPTIONS_KEY];
+
+            if (/tsconfig\.json$/.test(tsconfig)) {
```
Roaders: I think that this line is responsible for making typedoc 0.14 a breaking change for my projects.

The path that I use for my config file is something like `node_modules/my-config-project/config/typedoc.config.json`

This worked fine in 0.13 but breaks in 0.14.

I now need to update the path on 60+ projects..

PR #756, opened 2018-04-20
fix(plugins): traverse symlinked node_modules when scanning for plugins

christopherthielen:
When Typedoc is scanning `node_modules` for plugins, it checks if `node_modules` exists, and is a directory.  It does this using `lstatSync(..).isDirectory()`.  However, this doesn't return `true` when `node_modules` is a symlink.  

I'm staging docs generation in a temporary directory and want to symlink the project's node_modules in there.  To support this, I propose changing `lstatSync()` to `statSync()`.  Tested locally and seems to work.

### DIFF
--- a/src/lib/utils/plugins.ts
+++ b/src/lib/utils/plugins.ts
@@ -76,7 +76,7 @@ export class PluginHost extends AbstractComponent<Application> {
             let path = process.cwd(), previous: string;
             do {
                 const modules = Path.join(path, 'node_modules');
-                if (FS.existsSync(modules) && FS.lstatSync(modules).isDirectory()) {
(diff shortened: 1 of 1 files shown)


christopherthielen:
a workaround I am using:

- create a `node_modules`
- symlink each of `src/node_modules/*` to `target/node_modules/*`

aciccarello:
Thanks for the PR @christopherthielen!

remyrylan:
@aciccarello Thank you so much for all the work handling the releases recently -- is there any chance a new release will be cut in the next few days that has this included?

aciccarello:
@remyrylan I'm going to traveling the next few days but I'll try to do a release after that. Right now releases are blocked on some issues that came up in testing with the JSON serialization output.

PR #760, opened 2018-04-21
Renderer.getInstalledTheme(themeName)

bluelovers:
```
typedoc --out docs --theme color
```

will looking `typedoc-themes-color` module and if exists use theme folder

make us without install out theme every time every project

and can install as global module for reuse

### DIFF
--- a/src/lib/output/renderer.ts
+++ b/src/lib/output/renderer.ts
@@ -196,8 +196,11 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
             if (!FS.existsSync(path)) {
                 path = Path.join(Renderer.getThemeDirectory(), themeName);
                 if (!FS.existsSync(path)) {
-                    this.application.logger.error('The theme %s could not be found.', themeName);
-                    return false;
+                    path = Renderer.getInstalledTheme(themeName);
+                    if (path === null || !FS.existsSync(path)) {
+                        this.application.logger.error('The theme %s could not be found.', themeName);
+                        return false;
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/lib/output/renderer.ts
```diff
@@ -282,6 +285,21 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
         return Path.dirname(require.resolve('typedoc-default-themes'));
     }
 
+    /**
+     * Return the path to the target installed theme.
+     *
+     * @returns The path to the default theme.
+     */
+    static getInstalledTheme(themeName: string): string {
+        if (/[\/\\]/.test(themeName)) {
```
aciccarello: what exactly is this regex looking for?

src/lib/output/renderer.ts
```diff
@@ -282,6 +285,21 @@ export class Renderer extends ChildableComponent<Application, RendererComponent>
...
+            return null;
+        }
+        try {
+            return Path.join(Path.dirname(require.resolve(`typedoc-themes-${themeName}`)), 'theme');
```
aciccarello: should this be generalized to allow for themes other than the default two?

src/lib/output/renderer.ts
bluelovers: @aciccarello for chk theamname not a path, make sure it is name only

src/lib/output/renderer.ts
bluelovers: the thame looke like this https://github.com/bluelovers/typedoc-themes-color

Gerrit0:
I'm going to go ahead and make the call to not merge this in. Its a neat idea, but hides more magic from the user, which I don't like. Users can still specify the path with node_modules, and with 1.0, will be able to directly specify a npm module with just its name. Thanks for the PR anyways bluelovers!

PR #768, opened 2018-05-11
CommonMark code block class names

jayrylan:
This is a very simple fix that changes TypeDoc to follow the [CommonMark](http://commonmark.org) spec for code blocks:
<https://spec.commonmark.org/0.27/> Example: <https://spec.commonmark.org/0.27/#example-109>

Right now TypeDoc prints out code block prefixes with `lang-` as the prefix, which is far I understand it, is legacy from Highlight.js. Since TypeDoc uses Highlight.js, `lang-`, `language-` and even un-prefixed values are supported... But by enforcing the CommonMark standard developers will be able to leverage other CommonMark tools with TypeDoc.

### DIFF
--- a/src/lib/output/plugins/MarkedPlugin.ts
+++ b/src/lib/output/plugins/MarkedPlugin.ts
@@ -87,7 +87,8 @@ export class MarkedPlugin extends ContextAwareRendererComponent {
         Handlebars.registerHelper('relativeURL', (url: string) => url ? this.getRelativeUrl(url) : url);
 
         Marked.setOptions({
-            highlight: (text: any, lang: any) => this.getHighlighted(text, lang)
+            highlight: (text: any, lang: any) => this.getHighlighted(text, lang),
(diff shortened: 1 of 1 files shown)


remyrylan:
Looking at the tests and there are a ton of failings in there that are unrelated, looks like the CI environment needs some attention.

aciccarello:
The changes you've included so far sound good. Yeah, I'm aware of the test issues. #762 includes some fixes to how the tests are run so they aren't overwritten. I'll need to review why the models changed in the snapshot tests and correct things in a separate PR so leave things as they are until those are fixed.

aciccarello:
Sweet. Looks like your markedjs/marked#1265 PR ended the need for this. Thanks for your contributions! The updated specs now use the updated prefix

https://github.com/TypeStrong/typedoc/blob/eace04a441bdd558a859ac84ff78cd8f59775e1c/src/test/renderer/specs/index.html#L78-L81

PR #771, opened 2018-05-14
Added support for linked readme files.

archfz:
Hello there!

I needed all other readme pages linked to the main one compiled with typedoc, so I added support for this. This code is not perfect and unfortunately I don't have time to work on this further so someone will have to take over. PM me.

But this works nice till now. Note that I have set the default themes to be installed from the forked repository atm. Also note that media is not handled by this implementation yet, only the readme files.


archfz:
I am not sure why this is breaking, it passes all tests locally.

aciccarello:
Thanks for sharing your changes. Unfortunately this PR needs some updating to be mergeable including some changes to the approach. For now I am going to close this but if you would like to update it, please leave a comment and we can discuss what changes need to be made. Thanks!

PR #772, opened 2018-05-16
Prevent unnecessary outer directory searches

pro-src:
When searching up a directory tree without the added checks, every directory may be scanned until the outermost directory is reached even when the project files have already been found. This can cause avoidable permission errors when using the default options and thus typedoc to break on out-of-the-box builds for the simplest of projects. Without these checks, typedoc will always error on some setups.

* Stops the search when readme.md and package.json are found.
* Stops the search when `--readme = none` and package.json is found.

### DIFF
--- a/src/lib/converter/plugins/PackagePlugin.ts
+++ b/src/lib/converter/plugins/PackagePlugin.ts
@@ -83,43 +83,31 @@ export class PackagePlugin extends ConverterComponent {
      * @param node  The node that is currently processed if available.
      */
     private onBeginDocument(context: Context, reflection: Reflection, node?: ts.SourceFile) {
+        let packageAndReadmeFound = () => (this.noReadmeFile || this.readmeFile) && this.packageFile;
+        let reachedTopDirectory = dirName => dirName !== Path.resolve(Path.join(dirName, '..'));
+        let visitedDirBefore = dirName => this.visited.indexOf(dirName) !== -1;
+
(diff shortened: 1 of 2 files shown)


src/lib/converter/plugins/PackagePlugin.ts
```diff
@@ -98,6 +98,14 @@ export class PackagePlugin extends ConverterComponent {
                 break;
             }
 
+            if ((this.noReadmeFile || this.readmeFile) && this.packageFile) {
+                break;
+            }
+
+            if ((this.noReadmeFile || this.readmeFile) && this.packageFile) {
```
pro-src: I'm not sure how that check got duplicated. Feel free to spend 10 seconds to fix this :rofl:

aciccarello:
I tried to clean things up but now it doesn't seem to read the "typedoc" name from the package.json in the tests. Need to do some more digging.

pro-src:
Sorry, I don't have time to investigate this. You've either changed the logic, have a bad test, or it's only reproducible on your system. Nothing jumped out at me when reviewing the changes thus it's probably not the problem but rather a symptom. I'd review the test case very meticulously and hope it's not aberrant behavior. Καληνύχτα!

aciccarello:
@pro-src Thanks for contributing this. @spiltcoffee has cleaned up the issues so I'm going to close this PR.

PR #788, opened 2018-06-08
Copy inherited parameter doc to target if not already present

### LINKED ISSUE #787: @inheritDoc is not copying parameter documentation
When using `@inheritDoc` then only the method documentation is copied. Parameter documentation is not copied.

**Versions:**
* TypeDoc v0.11.1
* Node v10.4.0

**Reproduction code:**

```typescript
// test.ts
interface Base {
    /**
     * Base doc.
     *
     * @param p  Param doc.
     * @return Return value doc.
     */
    method(p: number): string;
}

class Test implements Base {
    /** @inheritDoc */
    method(p: number): string {
        return "";
    }
}
```

```
$ typedoc --out doc test.ts
```

**Expected behavior:**

The parameter `p` of the `Test` class must be documented with the text `Param doc`.

**Actual behavior:**

Parameter `p` is not documented at all. Only the name and type is printed to the HTML output.

The reason for this can be found in `ImplementsPlugin.copyComment`. This method only copies the parameter documentation when the parameter is already documented on the target method. This isn't the case when the method is simply annotated with `@inheritDoc`. There must be at least a `@param p` comment so the parameter documentation gets copied from the interface.

This can easily be fixed by reversing the logic. When target parameter is not already documented then copy documentation from the interface. This fixes the issue and also allows to write a more specific documentation for the parameter in the class.

This can be fixed with a few lines of code. Maybe I'll come up with a PR soon.

kayahr:
This change reverses the logic when copying inherited parameter comments. Before this change the source comment was only copied when there was already a target comment. This did not make any sense. Now the source comment is copied when target comment is missing. In this way all parameter documentation is copied when simply using `@inheritDoc` and it is possible to override specific parameter comments by documenting them with `@param` on the target method.

PR also includes a unit test for a few common scenarios.

This fixes #787

### DIFF
--- a/src/lib/converter/plugins/ImplementsPlugin.ts
+++ b/src/lib/converter/plugins/ImplementsPlugin.ts
@@ -3,6 +3,7 @@ import { Type, ReferenceType } from '../../models/types/index';
 import { Component, ConverterComponent } from '../components';
 import { Converter } from '../converter';
 import { Context } from '../context';
+import { Comment } from '../../models/comments/comment';
 
 /**
  * A plugin that detects interface implementations of functions and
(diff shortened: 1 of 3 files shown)


MaximilianLackaw:
Has there been any progress yet?

StephanBijzitter:
@Gerrit0 any chance you can get this merged in the coming weeks?

Gerrit0:
@StephanBijzitter fixing the conflicts would help a ton, I haven't had time to work on TypeDoc for a few weeks and it'll probably take me a while to get to this otherwise.

StephanBijzitter:
@Gerrit0 No worries, I already read that you didn't have a lot of spare time :-)

I re-applied the changes in a new branch and all tests (linting/unit) pass.
Next to that, I also checked the output (specs) myself and the JSON is as I would expect it to be.

new pull request: #1303

PR #794, opened 2018-06-11
Implement and use isTypeListAssignable to find super methods 

### LINKED ISSUE #793: @inheritDoc does not find method in implemented interface when method signature is different
When annotating a method implementation with `@inheritDoc` then it only works when the method signature exactly matches the one in the implemented interface. It does not work when widening the signature with optional parameters or union types or using generic types.

**Example:**

```typescript
// test.ts
interface Base<T> {
    /**
     * Some method doc.
     */
    someMethod(a: T): void;
}

class Implementation implements Base<string> {
    /** @inheritDoc */
    someMethod(a: string): void {}
}
```
```
$ typedoc --out doc test.ts
```

**Version:** TypeDoc 0.11.1

**Expected behavior:**
Generated documentation must show that `someMethod` is an implementation of the interface and it must copy the documentation from the interface to the implementation.

**Actual behavior:**
TypeDoc doesn't recognize the interface method and therefor doesn't link the interface and doesn't copy the documentation.

**Analysis:**
The reason for this problem is that TypeDoc does a very simple check to find the super method. It compares the array of parameter types with an `equal` check. So `string` does not match a typed parameter restricted to `string`.

This could be solved by implementing a more complicated `isAssignableFrom` check instead of simply comparing the types with `equals`.

kayahr:
The previously used isTypeListEqual method only works when method signature is exactly the same so it breaks when widening a signature or using generic types. This new method instead checks if the type used in the interface can be assigned to the type used in the implementation.

Fixes #793.

### DIFF
--- a/src/lib/converter/plugins/ImplementsPlugin.ts
+++ b/src/lib/converter/plugins/ImplementsPlugin.ts
@@ -65,7 +65,7 @@ export class ImplementsPlugin extends ConverterComponent {
                 interfaceMember.signatures.forEach((interfaceSignature: SignatureReflection) => {
                     const interfaceParameters = interfaceSignature.getParameterTypes();
                     classMember.signatures.forEach((classSignature: SignatureReflection) => {
-                        if (Type.isTypeListEqual(interfaceParameters, classSignature.getParameterTypes())) {
+                        if (Type.isTypeListAssignable(interfaceParameters, classSignature.getParameterTypes())) {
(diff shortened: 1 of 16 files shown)


aciccarello:
❤️ Thanks for this PR! I'll have to look at the code more later.

DanielSWolf:
It would be great if this could be merged. I'm having the same problem.

aciccarello:
Sorry for the delay, but I will need to find time to review this PR.

PR #805, opened 2018-06-22
 use soft break 

bluelovers:


aciccarello:
Thanks for the PR, but I don't think that we should include breaks. If you have a long line of text you should be able to break that up into multiple lines in the code without those breaks affecting the result.

Feel free to continue discussion if you disagree.

PR #824, opened 2018-08-02
Update for typescript v3.1.x

### LINKED ISSUE #822: New release on NPM
Hi, 

Love the project, just having a minor issue with regards to the Typescript version. Could we get a new release which at the very least updates Typescript? It seems there are a fair few issues (and PRs) that address the problem but there hasn't seemed to be any traction.

ahocevar:
I am trying to update TypeDoc to use `typescript@next`, which has some new features that we will be relying on in OpenLayers (see openlayers/openlayers#8449).

~~I think I was able to update the code properly, but I'm struggling with one failing test:~~

~~One thing I noticed is that none of the specs have the `isExported: true` flag set any more when regenerated with my code. Maybe that's related?~~

~~Any help here would be much appreciated.~~

I'm also willing to continue working on this pull request if there are more requirements for updating to a newer version that I might be missing. I'd be thankful for any comments or reviews.

Fixes #822.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "progress": "^2.0.0",
     "shelljs": "^0.8.1",
     "typedoc-default-themes": "^0.5.0",
-    "typescript": "2.7.2"
+    "typescript": "^3.1.0-dev.20180802"
   },
   "devDependencies": {
     "@types/mocha": "2.2.48",
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -24,7 +24,7 @@ const nonStaticKinds = [
  * @param name  The desired name of the reflection.
  * @returns The resulting reflection.
  */
-export function createDeclaration(context: Context, node: ts.Node, kind: ReflectionKind, name?: string): DeclarationReflection {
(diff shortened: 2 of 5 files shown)


ahocevar:
Thanks to looking at #825, I was able to fix the `isExported` issue. All test pass now.

VictorQueiroz:
This is a must have, we can't generate docs for projects with lower version of TypeScript right now. What's missing to accept this PR?

lf94:
I too need this. Lets do anything we need to get this in.

cedx:
TypeDoc should try to use the local TypeScript package (i.e. declared in package.json file), if any one is already installed. Relying on a development version of TypeScript is not a good idea from my point of view...

lf94:
> TypeDoc should try to use the local TypeScript package (i.e. declared in package.json file), if any one is already installed. Relying on a development version of TypeScript is not a good idea from my point of view...

Agreed for everything. IMO your suggestion should be another MR.

This MR should still be merged to encourage people to update. Typescript 3 - Typescript ~2.5 or whichever version typedoc is on, is not a huge difference.

Who do we poke to move this forward?

ahocevar:
When merged, this could simply be published as typedoc@next.

aciccarello:
@lf94 you poke me. I made some testing updates yesterday. I hope to get dependencies updated soon

PR #825, opened 2018-08-02
chore: upgrade to typescript 3.0.x

raymondfeng:
The PR allows typedoc to be used with typescript 3.0.x. Once the PR is merged, it should be released as new minor version.

### DIFF
--- a/src/lib/ts-internal.ts
+++ b/src/lib/ts-internal.ts
@@ -98,7 +98,7 @@ export function isBindingPattern(node: ts.Node): node is ts.BindingPattern {
 
 // https://github.com/Microsoft/TypeScript/blob/v2.1.4/src/compiler/utilities.ts#L1729
 export function getClassExtendsHeritageClauseElement(node: ts.ClassLikeDeclaration | ts.InterfaceDeclaration) {
-  return tsany.getClassExtendsHeritageClauseElement.apply(this, arguments);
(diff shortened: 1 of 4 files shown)


src/lib/ts-internal.ts
```diff
@@ -98,7 +98,7 @@ export function isBindingPattern(node: ts.Node): node is ts.BindingPattern {
 
 // https://github.com/Microsoft/TypeScript/blob/v2.1.4/src/compiler/utilities.ts#L1729
 export function getClassExtendsHeritageClauseElement(node: ts.ClassLikeDeclaration | ts.InterfaceDeclaration) {
-  return tsany.getClassExtendsHeritageClauseElement.apply(this, arguments);
+  return tsany.getClassExtendsHeritageElement.apply(this, arguments);
```
aciccarello: This should probably be a separate PR and the function name should be updated

raymondfeng: +1 to rename the exported function in a separate PR.

raymondfeng: @aciccarello Do you think it's better to keep both old and new exported functions for backward compatibility?

aciccarello: Since it is only used internally, I'd say change the calling function (I think there is only one) and get rid of any `getClassExtendsHeritageClauseElement` references.

aciccarello: This gave me an idea for #830 which is related

PR #834, opened 2018-08-12
Allow installing `typedoc` from git repo

grncdr:
The `prepare` script was introduced in npm 5 to distinguish between build steps and other prepublish tasks (which should be run from the `prepublishOnly` script.

npm no longer runs the `prepublish` script when installing from git repositories, but it will install dev deps and run `prepare` if it's present. Both are run before publishing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "test": "mocha -t 10000 --exit dist/test",
     "build": "grunt build_and_test",
     "prepublish": "npm run build",
+    "prepare": "grunt",
     "grunt": "grunt",
     "clean": "rm -rf node_modules package-lock.json lib coverage .tscache"
   },


demurgos:
I confirm `prepare` is the right place to build the final package. It would be nice to have it merged.

aciccarello:
Thanks! Shouldn't `prepublish`be removed then so it isn't built twice? I think there should also be a `prepublishOnly` script to run the tests.

grncdr:
I ended up tweaking `.travis.yml` a bit to get CI working with old versions of npm. This unfortunately means installing typedoc from git with an npm version < 4 doesn't actually work, but it seemed less invasive than putting checks inside npm scripts or grunt tasks.

aciccarello:
Ah, I didn't think about older versions of npm. I think it is reasonable to require npm 3+. I made a few changes to simplify things into a build and a test step. PRs welcome.

aciccarello:
Oh, I missed that it was npm 4. I'll have to think of a better solution.

aciccarello:
Posting for later reference
https://resin.io/blog/safely-migrating-away-from-prepublish-with-npm-4/

aciccarello:
As I've thought about this more. Requiring npm 4+ for installing via git doesn't seem that bad. Normal users won't be affected. I'll clean up this PR then merge.

Issue #836, opened 2018-08-14
Migrate off of grunt


aciccarello:
I imagine that build and configuration will be clearer/easier using npm scripts and js task files. Tests could also be executed using ts-node.

demurgos:
I had good experience using the Typescript, Mocha and ts-node, especially with Typescript 3 and project references.

Looking at your grunt file, you still have some use cases that may be hard to implement manually: text replacement in a watch task, some manual file updates, and some file copies. These are harder to migrate, but possible.
I still think it's good to have a proper task manager for more complex projects. In practice I find Gulp easier to configure so it may be worth a try (disclaimer: I'm a Gulp maintainer). The main issue I had with Grunt is that you lack flexibility if you cannot express your needs declaratively in their config object.

blink1073:
nb about TypeScript 3 project references, there is currently no output in [watch mode](https://github.com/Microsoft/TypeScript/issues/26147).

PR #840, opened 2018-08-17
Stricter lint config

Gerrit0:
Also turns on strictFunctionTypes, which required an update to the `ComponentClass` interface.

This change makes typedoc less likely to break if types change in the future since it removes type assertions which aren't currently needed.

Running `grunt build_and_test` will **not** lint files with the new `no-unnecessary-type-assertion` rule as it appears the plugin does not support that option. Once #836 is resolved, this should no longer be an issue. It is also important to be aware that some editor plugins (notably tslint for vscode) do not show rules which require type information.

I have added a `npm run lint` task which will run tslint directly and show lint errors from tslint rules which require type information and configured Travis to run this task when building.

### DIFF
--- a/src/lib/utils/component.ts
+++ b/src/lib/utils/component.ts
@@ -12,8 +12,8 @@ export interface Component extends AbstractComponent<ComponentHost> {
 
 }
 
-export interface ComponentClass<T extends Component> extends Function {
-    new(owner: ComponentHost): T;
+export interface ComponentClass<T extends Component, O extends ComponentHost = ComponentHost> extends Function {
+    new(owner: O): T;
 }
 
 export interface ComponentOptions {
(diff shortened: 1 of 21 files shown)


src/lib/utils/component.ts
```diff
@@ -12,8 +12,8 @@ export interface Component extends AbstractComponent<ComponentHost> {
 
 }
 
-export interface ComponentClass<T extends Component> extends Function {
-    new(owner: ComponentHost): T;
+export interface ComponentClass<T extends Component, O extends ComponentHost = ComponentHost> extends Function {
```
aciccarello: Why can't this type stay `CompilerHost`?

Gerrit0: Turning on `strictFunctionTypes` requires a stronger type for the `addComponent` method. In particular, the application component in [`application.ts`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/application.ts#L100-L104) adds several components which expect their owner to be of type `Application`. This change makes it possible to specify the owner type in [`component.ts`](https://github.com/TypeStrong/typedoc/pull/840/files#diff-9b7f6610719fb95a47e042dd5d47b905L215) and thus resolves the type incompatibility.

Issue #837, opened 2018-08-14
Add contrubuting.md file


aciccarello:
## Add basic info on how to contribute

- [x] Checking out git repo (basic git instructions)
- [x] How to build/test the code
- [x] update spec script (`npm run grunt -- update-specs`)
- [ ] rebase instructions

## Note some coding guidelines
- [x] Code should align with tslint rules
- [x] Build should pass
- [x] Don't bump the project version number

## Other
- [x] Info about this being a community run project
- [x] Add link to contributing.md to readme (and update specs)
- [x] Recommend gitter group for discussions about implementation

blink1073:
In JupyterLab we run [prettier](https://github.com/prettier/prettier) as a [precommit](https://github.com/jupyterlab/jupyterlab/blob/master/package.json#L26) hook, and it has taken away a large part of the maintenance burden of reviewing PRs.  As a developer, it is also nice to think less about the formatting of the code you are writing and more about the content.

webmaster128:
I also recommend prettier but as part of the build step (scripts.prebuild in package.json). I think and author should always explicitly see what he or she checks-in. If you push unlinted content, the CI should reject that.

blink1073:
Another option is as a pre-commit hook using [husky](https://github.com/typicode/husky).

aciccarello:
tslint should be failing the CI if there is a problem already. Autofixes are nice though. I haven't had great experiences with precommit hooks in the past but I'd be open to it or anything else that improved the contribution experience.

This issue is mostly focused on the contribution docs though :smile:

Gerrit0:
I've taken an initial crack at this in #841 - feedback welcome! I really wasn't sure how far to go when describing installation & setup. I don't think someone brand new to coding could/should pick up TypeDoc as their first project, but someone might prove me wrong.

Gerrit0:
Looking through the checklist in the original comment here, it looks like everything is complete, except perhaps the "rebase instructions" - The contributing.md file does include [updating your branch](https://github.com/TypeStrong/typedoc/blob/master/CONTRIBUTING.md#updating-your-branch) instructions, but doesn't use `git rebase`. Is this something that should be resolved before this issue can be marked as done?

aciccarello:
That should be sufficient. The [official GitHub guide](https://help.github.com/articles/syncing-a-fork/) recommends doing a merge. With GitHub's PR squashing it's less important for contributors to know how to rebase.

PR #841, opened 2018-08-18
#837 Add contributing.md

Gerrit0:
I really don't think this is complete - but its a start!

This PR adds contributing.md, adds a link to it in the readme, and rebuilds the specs to include the updated readme.

It also includes a `.gitattributes` file which overrides the `core.autocrlf = true` setting that most Windows Git setups will have. This is important because TypeDoc does not normalize line endings when parsing files. If a file is checked out with crlf line endings and a `specs.json` file is generated, tests can fail because multiline strings will include `"\r\n"` instead of `"\n"`. Most editors (Notepad being the one exception that comes to mind) will correctly handle either line endings, so this shouldn't cause any issues there.


contributing.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Linting
+
+To lint the TypeDoc code, run `npm run lint`. This will start tslint and check all files for stylistic problems. You can also install a tslint plugin for your editor to show most style problems as you type.
+
+You can automatically fix some style problems by running `npm run lint -- --fix`.
+
+#### Building
+
+To compile the TypeDoc source, run `npm grunt`. This will start the TypeScript compiler and output the compiled JavaScript to the `dist` folder. If you want to build and test in one step, run `npm build`.
```
blink1073: Should be `npm run grunt` and `npm run build`?

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Disable core.autocrlf's line ending conversion behavior to prevent test failures on Windows
+* text=lf
```
aciccarello: Are you testing this on windows? I thought on windows you want git to checkout files with crlf so that it matches windows editors. Does the build produce crlf endings?

If going with lf, we probably want an `.editorconfig` file with that setting as well.

contributing.md
```diff
@@ -0,0 +1,118 @@
...
+- **Provide an example where this enhancement would improve TypeDoc**
+- **If possible, list another documentation generator where this feature exists**
+
+### Code
+
+Unsure of where to begin contributing to TypeDoc? You can start by looking through the issues labeled [good-first-issue] and [help-wanted]. Issues labeled with [good-first-issue] should only require changing a few lines of code and a test or two. Issues labeled with [help-wanted] can be considerably more involved and may require changing multiple files.
+
+For instructions on setting up your environment, see the [setup](#setup---git-github-and-node) instructions in this document.
+
+If you have started work on an issue and get stuck or want a second opinion on your implementation feel free to reach out through [Gitter].
```
aciccarello: I'm wondering if there is a way to make the gitter channel more visible. It is really helpful to iterate on ideas before a PR is submitted.

.gitattributes
Gerrit0: Yes, my current environment is Windows 10 + Linux Subsystem

In general, yes, on Windows you should checkout files with crlf, however the build will use whatever line endings files are checked out with, resulting in failing tests.

It might be a better idea to normalize line endings when doing output instead of this, thoughts?

contributing.md
Gerrit0: Fixed! That's what I get for using yarn...

contributing.md
Gerrit0: I've added the Gitter link to the PR section as well, I would have liked to put it under the "How Can I Contribute" section, but struggled to come up with appropriate wording.

.gitattributes
demurgos: In practice, the CRLF situation leads to many issues across platforms. All Windows editors support LF properly so I found it easier to enforce LF.

I use the following `.gitattributes` file:
```
# Enforce `lf` for text files (even on Windows)
text eol=lf
```

End the following `.editorconfig` setting:
```
end_of_line = lf
```

In `tsconfig.json`, you should also ensure that `"newLine": "lf"` is set.

demurgos:
This is a minor nitpick, but could you rename the file to uppercase? It's one of those project files like `README` and `LICENSE` where using uppercase is the convention.

Issue #249, opened 2016-07-05
Generate links to the source codes on GitHub Enterprise


kimamula:
Currently TypeDoc can generate links to the source codes on GitHub only if they are on github.com.

https://github.com/TypeStrong/typedoc/blob/v0.4.4/src/lib/converter/plugins/GitHubPlugin.ts#L56
https://github.com/TypeStrong/typedoc/blob/v0.4.4/src/lib/converter/plugins/GitHubPlugin.ts#L109

It would be great if TypeDoc can also generate links to the source codes on GitHub Enterprise.

aciccarello:
Would be covered by #129

gdelmas:
i had the same issue and created a plugin [typedoc-plugin-sourcefile-url](https://github.com/gdelmas/typedoc-plugin-sourcefile-url)

the plugin allows to create custom URLs to source code files.

it can be simply installed with:

    npm install --save-dev typedoc-plugin-sourcefile-url

PR #843, opened 2018-08-23
Add support for source links on GitHub enterprise instances

tomratcliffe:
This adds support for GitHub Enterprise source URLs found in the list of `git ls-remote --get-url`.

The updated regex checks for URLs like `github.acme.com` instead of just `github.com`, and then uses the matched hostname when generating the source URLs.

I didn't add any tests for this as I'm not sure what your testing strategy is/I couldn't find a testing file for just this module. Happy to try and add this in though if required!

Should be backwards compatible with old URLs.

Fixes https://github.com/TypeStrong/typedoc/issues/249, which was closed in favour of https://github.com/TypeStrong/typedoc/issues/129.

This doesn't entirely solve https://github.com/TypeStrong/typedoc/issues/129, but closes the gap on allowing GHE to work out of the box.


tomratcliffe:
I had a look at writing some tests, and can get one that would test that it defaults to `github.com`, but I can't work out how to stub out the internal ShellJS module and make it return whatever I want (so I could fake having a certain remote on a git repo):

```ts
import * as github from './../lib/converter/plugins/GitHubPlugin';
import Assert = require('assert');

describe.only('GitHubRepository', function() {
    let repository = new github.Repository('', '');

    describe('Application', function() {
        describe('contructor', function() {
            it('must default to github.com hostname', function() {
                Assert.equal(repository.gitHubHostname, 'github.com');
            });
        });
    });
});
```

Any ideas @aciccarello ?

tomratcliffe:
I figured it out, with a small refactor. Should be good to go @aciccarello!

tomratcliffe:
@aciccarello Any thoughts on this? Thanks!

aciccarello:
Sorry for not following up on this @tomratcliffe. I'm hoping to look at it this week.

aciccarello:
Thanks for your contribution @tomratcliffe!

Issue #846, opened 2018-08-28
An error occurs when using "typedoc-plugin-external-module-name" to prompt "Error: The component `s'has been added."


Houfeng:
An error occurs when using "typedoc-plugin-external-module-name" to prompt "Error: The component `s'has been added."

![image](https://user-images.githubusercontent.com/1432335/44721739-728b9c00-aafd-11e8-89d7-1cbf9d9db59b.png)

aciccarello:
Could you share how you are getting this error. I'm curious why it thinks it has been added twice.

justrhysism:
@aciccarello I experience this when using _gulp watch_ to rebuild the docs (using [gulp-typedoc](https://github.com/rogierschouten/gulp-typedoc)).

The error isn't thrown on the initial build, but on subsequent builds. *gulp-typedoc* creates a [new instance](https://github.com/rogierschouten/gulp-typedoc/blob/master/index.js#L50), so one would assume that everything internally is "fresh".

*typedoc-plugin-external-module-name* simply just adds itself [via `addComponent`](https://github.com/christopherthielen/typedoc-plugin-external-module-name/blob/master/index.js#L4).

It appears as though the typedoc application lifecycle isn't ending. I'm not an expert on the application side of things, but I feel like it _shouldn't_ end and restart all the time, so if this decision _was_ deliberate and it's _by design_, that makes sense - however if this _is the case_, I don't believe that it should be up to plugin authors to check to see if their plugin/component has already been added.

aciccarello:
@justrhysism This sounds like it could be a gup-typedoc specific issue. Typedoc currently isn't set up for multiple builds.

justrhysism:
> gup-typedoc specific issue

Is it, though? I don't know for sure, but my gut feel is that it's simply _exposed_ by usage with gulp. Maybe someone with a better idea of the inner workings of gulp pipes may be in a better position to suggest otherwise, but the instance instantiation:

```
const app = new typedocModule.Application(options);
```

is encapsulated in a function, so once that function ends, `app` is discarded, and with it, I would assume, the `Application`.

> Typedoc currently isn't set up for multiple builds.

Maybe not concurrent builds, but in my particular use-case which exposed this issue, builds were only being triggered in a serial fashion (i.e. no more than one build occurring at any one time).

---

[...]

PR #847, opened 2018-08-28
fix: No error is thrown when adding plug-ins repeatedly.  #846

Houfeng:


aciccarello:
For some reason I always think I can make changes from the online editor without breaking things 😰 

We can merge this as a workaround and consider making larger improvements later.

PR #850, opened 2018-08-31
Add localization plugin to Readme.md plugin section

zdrawku:
Adding localization plugin in the plugin section. 

This plugin will ease every user, to:
1. Generate a json files with all code comments based on the project structure
2. Edit the json with the translation on the desired language
3. Re-generate the typodoc documentation with the updated jason for the new language


aciccarello:
Thanks! the spec file will also need to be updated since it includes the readme (maybe we should change that test?) the easiest thing would be to copy the missing line from the travis file to the spec

aciccarello:
Thanks @zdrawku !

PR #857, opened 2018-09-19
Use a dummy readme when testing the renderer

### LINKED ISSUE #853: Don't use the project readme when testing the renderer
The current renderer end-to-end test generates an index.html file based on the project readme. While we want to check that the renderer converts markdown to HTML, it is problematic to use the project readme. Any contributions to improve the readme will fail the CI because they also require updating the specs.

The typedoc application config in the render.ts spec file should be updated to point to a static dummy readme that does not change as frequently as the project readme.

## Steps
- Add dummy readme that has some basic markdown to convert
- Update application config in src/test/renderer.ts to point to readme
- Update specs

Towerism:
This should be able to close #853

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -1,0 +1,28 @@
+# Example
+
+> Example Documentation
+
+## Installation
+
+Example install
+
+```bash
+$ npm install examples --save-dev
+```
+
+Example executable: ``examples``. 
+
+## Usage
+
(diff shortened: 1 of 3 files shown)


Towerism:
One test timed out ¯\_(ツ)_/¯

aciccarello:
Great work, thanks!

PR #863, opened 2018-10-01
Upgrade typescript to v3.1.x

ahocevar:
This pull request updates the TypeScript dependency to the latest stable v3.1.x version (currently v3.1.1).

PR #865, opened 2018-10-02
fix(HelperStack): deactivate theme resources correctly

### LINKED ISSUE #864: Unable to deactivate the theme resources in order to add some custom Handlebars helpers
I would like to ask whether you have an approach of adding new Handlebars helpers within external plugins?
I have noticed that in order to register some new theme resource you firstly need to deactivate them. So I tried it but without any success because the deactivation was not full. More specifically I have noticed in your [helper.ts](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts) within the [HelperStack](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts#L26) class that [activate ](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts#L34) and [deactivate ](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts#L56) methods have same [if statement](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts#L57) which prevents deactivation at the most abstract level which is within the  [ResourceStack](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/stack.ts#L102) class.

I suppose that [here](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/helpers.ts#L57) you should have been invoking the deactivate method of the abstract [ResourceStack](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/utils/resources/stack.ts#L102) class. Which leads to the following code 
```typescript
if (!super.deactivate())
```
Please notify me if I'm wrong!

my source:
```typescript
this.application.renderer.theme.resources.deactivate();
this.application.renderer.theme.resources.helpers.addOrigin('custom-helper', `<path-to-my-custom-helper-function>`));
this.application.renderer.theme.resources.activate();
```

Although if you have any suggestions of how I could add additional Handlebars helpers through my external plugin it would be great!

Here you can find our [Localization ](https://github.com/IgniteUI/typedoc-plugin-localization) plugin.

Aleksandyr:
Closes #864

### DIFF
--- a/src/lib/output/utils/resources/helpers.ts
+++ b/src/lib/output/utils/resources/helpers.ts
@@ -54,7 +54,7 @@ export class HelperStack extends ResourceStack<Helper> {
     }
 
     deactivate(): boolean {
-        if (!super.activate()) {
+        if (!super.deactivate()) {
             return false;
         }
 


zdrawku:
Locally are failing only two tests, that has nothing to do with the code change itself.

With or witohut the code change, these two tests are failing:

image

image

aciccarello:
Thanks for the contribution. I saw the same tests failing on another PR so its helpful to know that it is not just that one.

zdrawku:
@aciccarello what do you think of the code change? It looks just a little change, although I am sure that we must be careful with the verification. 

I wanted to ask you, what could we expect, is it possible to review and verify the PR by the end of this week, or before 10th of October? I am asking because this is a showstopper for a feature that we are currently developing.

In addition, we've notices that our change is not used anywhere in the typdoc library, it looks like it is exposed for external use only. 

Looking forward to your thoughts!

aciccarello:
That seems reasonable since it does not seem to be called. I'll merge this and try to get a release out this week including #863 if possible.

zdrawku:
Thank you @aciccarello!

zdrawku:
Hi @aciccarello, I am following up on the release that will include this PR change and #863, do you have any updates on this?

zdrawku:
@aciccarello ! :) version 0.13.0-0 is what we have been waitng for.
Thank you, everything is wokring like a charm now.

PR #871, opened 2018-10-05
Added declarationMap to list of ignored tsconfig options.

### LINKED ISSUE #844: Error if declarationMap is enabled in tsconfig.json
If `declarationMap` is set to `true` in tsconfig.json, typedoc fails with this error:
```
Error: Option 'declarationMap' cannot be specified without specifying option 'declaration'.
```
It seems like `declaration` is _unset_ by typedoc, but `declarationMap` is not.

NaridaL:
Otherwise, setting declarationMap causes an error "declarationMap requires declaration to be set".

Fixes #844

Didn't see any framework to test tsconfigs...


aciccarello:
Thanks for the fix!

PR #872, opened 2018-10-05
Fix gitattributes.

NaridaL:
text=lf isn't a valid combination and I was getting lf/crlf errors on windows.


NaridaL:
See https://stackoverflow.com/a/42135910/1980909 and https://git-scm.com/docs/gitattributes for valid attribute values.

Gerrit0:
Not entirely sure how I mixed that up, thanks!

PR #873, opened 2018-10-05
Added support for conditional types. i.e. `X extends A ? B : C`

### LINKED ISSUE #831: Ternary conditionals not supported in typescript 2.8
Can we upgrade the typescript version so that there arent any errors during compilation for newer types?

NaridaL:
Currently results in { "type": "unknown", "name": "X extends A ? B : C" }

See https://github.com/TypeStrong/typedoc/issues/831

Now fixes #831


Gerrit0:
Thanks for this! I've merged it in with #1153.

Issue #869, opened 2018-10-04
TypeDoc treats module merged with superclass as a child of a subclass


rbuckton:
TypeDoc creates the wrong reflections in the following case:

```ts
// "foo.ts"
export class Foo { }
export namespace Foo { // merges with static side of class `Foo`
  export const x = 1;
}

export class Bar extends Foo {
}
```

This results in a file named `_foo_.bar.foo.html`, as it treats the namespace `Foo` as a member of `Bar`. It is possible this may be due to the fact the `ModuleConverter` does not have a check for `context.isInherit`.

rbuckton:
I was able to patch around this behavior with this:

```ts
const savedConvert = ModuleConverter.prototype.convert;
ModuleConverter.prototype.convert = function (context, node) {
    if (context.isInherit && context.inheritParent === node) {
        const reflection = <DeclarationReflection>context.scope;
        context.withScope(reflection, () => {
            if (node.body) {
                this.owner.convertNode(context, node.body);
            }
        });
        return reflection;
    }
    return savedConvert.call(this, context, node);
};
```

PR #876, opened 2018-10-06
Fix clodule with subclass

NaridaL:
Should fix https://github.com/TypeStrong/typedoc/issues/869


src/lib/converter/nodes/module.ts
```diff
@@ -22,20 +22,14 @@ export class ModuleConverter extends ConverterNodeComponent<ts.ModuleDeclaration
...
     convert(context: Context, node: ts.ModuleDeclaration): Reflection {
-        const parent = context.scope;
-        const reflection = createDeclaration(context, node, ReflectionKind.Module);
-
+        const reflection = context.isInherit && context.inheritParent === node
+            ? <DeclarationReflection> context.scope
+            : createDeclaration(context, node, ReflectionKind.Module);
         context.withScope(reflection, () => {
-            if (parent instanceof ProjectReflection && !context.isDeclaration &&
-                (!module || module.valueOf() === ts.ModuleKind.None.valueOf())) {
```
NaridaL: `module` referred to the @types/node global. ts.ModuleKind is used only as the compilerOptions type as far as I can see. And valueOf?!

PR #890, opened 2018-10-30
Remove reflection perf

igandrews:
There's a performance bug in the removeReflection because the symbolMapping is enumerated for each call. I have a large library where there are many types and members marked as @hidden. The CommentsPlugIn stores those in the hidden array and then later invokes the removeReflection for each. Each time the method is invoked it iterates the entire symbolMapping of the project which in my case started with around 90k items. So there were about 35k hidden items which including descendants resulted in about 300k calls to removeReflection. 

This change breaks that up so that it invokes the removeReflection for each hidden item as it did except it does not touch the symbolMapping. It just tracks the id's of the items processed. Then after all the hidden items have been processed, it can just enumerate that once and use the id's of the hidden reflections and their descendants to determine which need to be deleted.

### DIFF
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -222,9 +222,6 @@ export class CommentPlugin extends ConverterComponent {
 
         if (this.hidden) {
             const project = context.project;
-            // this.hidden.forEach((reflection) => {
-            //     CommentPlugin.removeReflection(project, reflection);
-            // });
             CommentPlugin.removeReflections(project, this.hidden);
         }
     }
@@ -341,7 +338,6 @@ export class CommentPlugin extends ConverterComponent {
         reflection.traverse((child) => CommentPlugin.removeReflection(project, child, deletedIds));
 
         const parent = <DeclarationReflection> reflection.parent;
-
         parent.traverse((child: Reflection, property: TraverseProperty) => {
             if (child === reflection) {
                 switch (property) {
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -222,9 +222,10 @@ export class CommentPlugin extends ConverterComponent {
 
         if (this.hidden) {
             const project = context.project;
-            this.hidden.forEach((reflection) => {
-                CommentPlugin.removeReflection(project, reflection);
-            });
+            // this.hidden.forEach((reflection) => {
+            //     CommentPlugin.removeReflection(project, reflection);
+            // });
+            CommentPlugin.removeReflections(project, this.hidden);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/converter/plugins/CommentPlugin.ts
```diff
@@ -314,11 +312,30 @@ export class CommentPlugin extends ConverterComponent {
         }
     }
 
+    /**
+     * Remove the specified reflections from the project.
+     */
+    static removeReflections(project: ProjectReflection, reflections: Reflection[]) {
+        const deletedIds = new Array<number>();
```
jonchardy:
```suggestion
        const deletedIds: number[] = [];
```

src/lib/converter/plugins/CommentPlugin.ts
```diff
@@ -314,11 +312,30 @@ export class CommentPlugin extends ConverterComponent {
...
+    }
+
     /**
      * Remove the given reflection from the project.
      */
-    static removeReflection(project: ProjectReflection, reflection: Reflection) {
-        reflection.traverse((child) => CommentPlugin.removeReflection(project, child));
+    private static removeReflection(project: ProjectReflection, reflection: Reflection, deletedIds: number[]) {
+        // keep track of the reflections that have been deleted
+        deletedIds.push(reflection.id);
```
jonchardy: I'd move this to after the delete just to make sure the delete actually took place.

jonchardy:
This looks like a pretty solid improvement. I just tried this out with our project and it reduced our generation time by ~50% given the amount of hidden declarations we have. I suspect many large projects will benefit from this. Made some minor coding suggestions above.

aciccarello:
Thanks for reviewing @jonchardy!

aciccarello:
Thanks everyone!

jonchardy:
I just noticed one of the linked plugins in the readme uses CommentPlugin.removeReflection. I think this change will break that. Should we consider keeping the API the same, or should plugin authors avoid using internal methods that could change?

igandrews:
Ultimately it would be best if the implementation was moved out of the CommentPlugin if its meant for general usage. I think I saw some other PRs relating to trying to move that to context (and possibly raising notifications) which may be worth considering at some point. As to this issue, I guess one simple option would be to check for the deletedIds being provided and if not (i.e. its null/undefined) then perform the loop it used to.

jonchardy:
Oh yea. https://github.com/TypeStrong/typedoc/pull/632.

@aciccarello Thoughts? I think moving it out of a plugin makes sense, and I also think we should keep the performance improvements in place when doing a mass removal.

igandrews:
For now though if you want to avoid breaking that plugin then I think just re-inserting the loop when the deletedIds is null is the right thing to do. I can try to create another PR with that later or tomorrow unless you want to do it.

aciccarello:
👍 I think both the compatibility fix and moving the method out of the plugin makes sense.

PR #896, opened 2018-11-04
Avoid documenting excluded files.

### LINKED ISSUE #319: Not able to exclude node_modules folder
I run the following command on an Angular2 project:

`typedoc --experimentalDecorators --target 'es5' --module 'commonjs' --externalPattern node_modules --excludeExternals  --ignoreCompilerErrors --out docs/ web/`

I am apparently not able to properly exclude the node_modules folder from the documentation. Whatever I do, the node_modules packages get documented. This is the folder structure
/web
_ _/app
_ _/assets
_ _/e2e
_ _/node_modules
_ _/typings

I tried using exclude and externalPattern + excludeExternals. I tried many combinations of patterns, but apparently I am doing it wrong. I tried values like this:
node_modules
**/node_modules/**
node_modules/**
node_modules/*_/_.ts
etc.

I would be very grateful if someone pointed out my mistake and made a couple of examples how the patterns should be defined?

### LINKED ISSUE #839: Can't Exclude anything!?
I've been trialing and erroring for hours and nothing I do will exclude the node_modules from the output. Specifically a package called `@types/bluebird` is causing errors with its `index.d.ts`. 

typedoc version: 0.11.1
node version: 10.8.0
typescript version: 3.0.1

project structure is like so:
- my_project
-- src
-- package.json
-- docs
-- tsconfig.json
-- typedoc.json

I have a typedoc.json like below:
```
{
  "mode": "modules",
  "out": "docs",
  "exclude": "node_modules",
  "theme": "default",
  "ignoreCompilerErrors": true,
  "excludePrivate": true,
  "excludeNotExported": "true",
  "target": "ES6",
  "moduleResolution": "node",
  "preserveConstEnums": "true",
  "stripInternal": "true",
  "suppressExcessPropertyErrors": "true",
  "suppressImplicitAnyIndexErrors": "true",
  "module": "commonjs",
  "excludeExternals": true,
  "externalPattern": "node_modules",
  "includeDeclarations": false
}
``` 

then I run:
`typedoc --json typedoc.json ./src`
or
`typedoc --json typedoc.json`

No output is ever achieved because there is a ton of useless errors from the bluebird package. Nothing else is in the way of generating the docs as they worked fine up until something required the `@types/bluebird`.

Why does the excludes not work? I can't even get excludes to exclude my own files in my own project. Can't there be an option like `excludeNodeModules: true` so we don't have to worry about them at all.

Gerrit0:
This is an attempt at correcting how the exclude option works. I'd like to get some feedback from people running into this problem before merging these changes in.

I previously [raised](https://github.com/TypeStrong/typedoc/issues/319#issuecomment-435104571) some concerns about how a change like this would affect errors/output, so here's how I decided to handle each concern:

1. If TypeScript reports that a diagnostic message is not linked to a source file, it is still logged as an error.
1. If TypeScript reports that a diagnostic message is linked to a source file returned by [Application.expandInputFiles](https://github.com/TypeStrong/typedoc/blob/master/src/lib/application.ts#L249) it is still logged as an error.
1. Otherwise, the error will be suppressed since it belongs to a file which was not originally included by the user.

I ran the project on the TypeDoc repo with the utils folder excluded and noted that classes defined in the utils folder were not present in the JSON output. In the HTML output, their name will be printed as plain text (much like `Promise`) if they are referred. I also introduced errors both in the utils folder and outside of the utils folder and noted that only errors outside of the utils folder were reported. Errors in the utils folder did not prevent output.

Fixes #319 
Fixes #839

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,7 +5,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -10,6 +10,7 @@ import { ConverterComponent, ConverterNodeComponent, ConverterTypeComponent, Typ
 import { CompilerHost } from './utils/compiler-host';
 import { Component, Option, ChildableComponent, ComponentClass } from '../utils/component';
 import { normalizePath } from '../utils/fs';
+import { createMinimatch } from '../utils/paths';
 
 /**
  * Result structure of the [[Converter.convert]] method.
@@ -384,26 +385,33 @@ export class Converter extends ChildableComponent<Application, ConverterComponen
     private compile(context: Context): ReadonlyArray<ts.Diagnostic> {
         const program = context.program;
 
-        program.getSourceFiles().forEach((sourceFile) => {
+        const exclude = createMinimatch(this.application.exclude || []);
+        const isExcluded = (file: ts.SourceFile) => exclude.some(mm => mm.match(file.fileName));
+
+        const includedSourceFiles = program.getSourceFiles()
+            .filter(file => !isExcluded(file));
+        const isRelevantError = ({ file }: ts.Diagnostic) => !file || includedSourceFiles.includes(file);
+
+        includedSourceFiles.forEach((sourceFile) => {
             this.convertNode(context, sourceFile);
         });
 
-        let diagnostics = program.getOptionsDiagnostics();
+        let diagnostics = program.getOptionsDiagnostics().filter(isRelevantError);
         if (diagnostics.length) {
             return diagnostics;
         }
 
-        diagnostics = program.getSyntacticDiagnostics();
+        diagnostics = program.getSyntacticDiagnostics().filter(isRelevantError);
         if (diagnostics.length) {
             return diagnostics;
         }
 
-        diagnostics = program.getGlobalDiagnostics();
+        diagnostics = program.getGlobalDiagnostics().filter(isRelevantError);
         if (diagnostics.length) {
             return diagnostics;
         }
 
-        diagnostics = program.getSemanticDiagnostics();
+        diagnostics = program.getSemanticDiagnostics().filter(isRelevantError);
         if (diagnostics.length) {
             return diagnostics;
         }
--- a/src/test/module/a.ts
+++ b/src/test/module/a.ts
@@ -1,0 +1,5 @@
+import { multiply } from './b';
+
+export function add(a: number, b: number) {
+    return a + multiply(b, 1);
+}
--- a/src/test/module/b.ts
+++ b/src/test/module/b.ts
@@ -1,0 +1,3 @@
+export function multiply(a: number, b: number) {
+    return a * b;
+}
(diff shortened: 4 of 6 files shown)


.editorconfig
```diff
@@ -5,7 +5,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
```
slinkardbrandon: Do we need to change indent size?

slinkardbrandon:
When I npm link this branch in typedoc to my project I'm running into an issue with the ConvertComponent invocation.

```
typedoc

Error: The plugin /Users/bslinkard/someproject/node_modules/typedoc-plugin-markdown could not be loaded.

TypeError: Class constructor ConverterComponent cannot be invoked without 'new'
    at new MarkdownPlugin (/Users/bslinkard/someproject/node_modules/typedoc-plugin-markdown/dist/plugin.js:31:42)
    at Converter.addComponent (/Users/bslinkard/www/typedoc/dist/lib/utils/component.js:123:19)
    at Converter.addComponent (/Users/bslinkard/www/typedoc/dist/lib/converter/converter.js:28:33)
    at module.exports (/Users/bslinkard/someproject/node_modules/typedoc-plugin-markdown/dist/index.js:36:19)
    at PluginHost.load (/Users/bslinkard/www/typedoc/dist/lib/utils/plugins.js:36:21)
    at CliApplication.bootstrap (/Users/bslinkard/www/typedoc/dist/lib/application.js:42:22)
    at CliApplication.bootstrap (/Users/bslinkard/www/typedoc/dist/lib/cli.js:24:30)
    at new Application (/Users/bslinkard/www/typedoc/dist/lib/application.js:31:14)
    at new CliApplication (/Users/bslinkard/www/typedoc/dist/lib/cli.js:22:1)
    at Object.<anonymous> (/Users/bslinkard/www/typedoc/bin/typedoc:4:1)

Using TypeScript 3.1.1 from /Users/bslinkard/www/typedoc/node_modules/typescript/lib
Rendering [========================================] 100%
Error: Documentation could not be generated due to the errors above.
```

In my package.json:

```json
    "docs": "typedoc --exclude node_modules"
```

In my tsconfig.json:
```json
"typedocOptions": {
    "mode": "file",
    "out": "docs",
    "theme": "minimal",
    "exclude": ["test", "node_modules"],
    "excludePrivate": true,
    "excludeNotExported": true,
    "moduleResolution": "node",
    "module": "commonjs",
    "hideGenerator": true,
    "readme": "none"
  }
```

.editorconfig
Gerrit0: The indent size in the tslint.json configuration (which predates the editorconfig file) is 4 spaces, so this is just correcting a discrepancy.

Gerrit0:
That's... odd to say the least. I don't think it's related to this PR since the source of the error appears to be the markdown plugin you have installed.

```text
at new MarkdownPlugin (/Users/bslinkard/someproject/node_modules/typedoc-plugin-markdown/dist/plugin.js:31:42)
```

I'd be willing to bet that `npm link` won't work with plugins since there are multiple `node_modules` folders, but I haven't tested to confirm this

balupton:
When will this land?

9oelM:
Got merged!

balupton:
Great! When will a release land?

aciccarello:
I'm planning on merging #922 before pushing a release

PR #897, opened 2018-11-05
Remove reflection update

jonchardy:
Minor update for #890 so CommentPlugin.removeReflection doesn't have incompatible changes.

@aciccarello @igandrews


Gerrit0:
The private modifier should probably be removed to restore the original visibility. JS users of this method will receive the fix, but anyone who uses TypeScript will still encounter compile errors.

Gerrit0:
Thanks!

PR #900, opened 2018-11-12
Enable rendering of tables in Markdown

### LINKED ISSUE #899: Allow different markdown flavors
When rendering Markdown files that contain tables (in the GitHub syntax), they are not rendered as HTML by TypeDoc

ItalyPaleAle:
(Closes #899)

This enables rendering of tables in Markdown files (e.g. README.md files) in the resulting documentation.

Note: tables aren't part of the standard Markdown specifications, but are an extension of the Github Flavored Markdown. In order to enable [rendering of tables in Marked.js](https://marked.js.org/#/USING_ADVANCED.md#options), I also had to enable GFM support. While all tests for typedoc succeeded, this change _might_ break some Markdown currently in use.

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -26,3 +26,11 @@ $ examples --out path/to/documentation/ path/to/example/project/
 
 * `--example <project>`<br>
   Example explanation.
+
+### List
+
+| Tables | Are | Cool |
+|----------|:-------------:|------:|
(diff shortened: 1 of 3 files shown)


aciccarello:
Thanks for the contribution. However I'll need to close this as I don't think typedoc should be tied to GFM.

PR #906, opened 2018-11-21
Store repository lookup path as lower case

Roaders:
As discussed here:

https://stackoverflow.com/questions/53396610/why-does-git-rev-parse-show-toplevel-return-different-casing-than-actual-path

the casing of paths on windows cannot be guaranteed. If a differently cased is returned from `git rev-parse --show-toplevel` than is reported as the first part of the filename's path then for every single file a new repository will be created and a new call to the git command line will be made - taking approximately 2 seconds per file. This makes it appear like typedoc has crashed (in my example I had over 1000 files).

### DIFF
--- a/src/lib/converter/plugins/GitHubPlugin.ts
+++ b/src/lib/converter/plugins/GitHubPlugin.ts
@@ -201,18 +201,20 @@ export class GitHubPlugin extends ConverterComponent {
 
         // Check for known repositories
         for (let path in this.repositories) {
-            if (!this.repositories.hasOwnProperty(path)) {
+            const lookupPath = path.toLowerCase();
+
+            if (!this.repositories.hasOwnProperty(lookupPath)) {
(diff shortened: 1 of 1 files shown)


src/lib/converter/plugins/GitHubPlugin.ts
```diff
@@ -201,18 +201,20 @@ export class GitHubPlugin extends ConverterComponent {
 
         // Check for known repositories
         for (let path in this.repositories) {
-            if (!this.repositories.hasOwnProperty(path)) {
+            const lookupPath = path.toLowerCase();
```
aciccarello: Shouldn't this already be lowercase?

Roaders:
I removed the extra variable.
I have tested this on my work PC that had the issue but no way of testing it if you have a computer without path casing weirdness.

Issue #595, opened 2017-09-20
typedoc Restructure


shlomiassaf:
I have been using typedoc for a while, it has great potential and it works great but it's not developer friendly when it comes to plugins...

Here are some things I think would help:

- [x] Allow ES6/ES2015-style import syntax for plugins (PR ready, #594)
- [ ] Host library under an organisation (npm scope)
- [ ] Create monorepo structure for the repository, moving non-core plugins to a separate package under the org.
- [ ] Arrange `extensibility` related symbols in a single module (see info below)
- [ ] Simplify external plugin registration. (`@Component` wont work outside of the core package)
- [x] Refactor JSON generation process to support a plugin architecture (PR Ready, #597)
- [ ] Support `object` type for options (might come from js file, or CLI stringified argument). It will help complex plugins to "own" an object for options and not share the root options object.
- [ ] Think how to avoid auto-plugin registration (via `@Component`). This will allow inheriting plugins without registering them (or manually unreging them)
- [ ] Better docs... (cobbler, shoes, not there... )

The approach is to simplify the codebase, making the core small and allowing plugins to add features.

I have several plugins already built, thing like:
  - **3rd party modules plugin**: Make documentation aware of 3rd party modules either by linking symbols to external documentation or by fully embedding type info from `d.ts` files (2 modes)
     I was able to include the whole `@angular` docs, creating selective mashups etc... 2nd mode is not that useful but it's nice.
> Supports binding import declaration (`import X from 'Y'`) and namespace import declaration (`import * as Y from 'Y'`)
 
  - **library-mode plugin**: multi-mode plugin that allows exporting libraries from a single endpoint, creating a multi-lib documentation (monorepo) and other stuff...

 - **objectify**: A plugin approach for creating JSON (`toObject()`). Reflection, Type, Source, Group, etc are all serialized to JSON via plugins that hook into events using the core plugin architecture. This plugin provides the core plugin logic and routing plus core `toObject()` plugin replacements for all native symbols (reflection, types, source etc..).

 - **json-datasource**: Using **objectify** to tap in an convert the JSON into a datasource/table like structure, having all reflection in one reflection table (hash), types in type table,  sources in source table etc... The graph is kept in tact by having pointers (id numbers) instead of objects.
This is structure is suitable for SPA apps that require lookpups, quick access etc.
In addition, the plugin will normalize the output so flags will output in their numeric form and not verbose form. It will add a table for each flag enum as a reference.

[...]

PR #914, opened 2018-12-10
Adds developer / contributor documentation explaining typedoc architecture

cspotcode:
This PR adds some developer-facing documentation for typedoc contributors.  I figure it's better to have something in master than to require contributors to google around for github threads and gists.

@mootari I copied the gist you posted last year: https://github.com/TypeStrong/typedoc/issues/595#issuecomment-353752599  Is that ok?

### DIFF
--- a/docs/developer/events.md
+++ b/docs/developer/events.md
@@ -1,0 +1,70 @@
+*Copied from https://gist.github.com/mootari/d39895574c8deacc57d0fc04eb0d21ca*
+*Original author: https://github.com/mootari*
+
+# Events
+
+## Components
+
+- `ComponentEvent.ADDED (event: ComponentEvent)`
+- `ComponentEvent.REMOVED(event: ComponentEvent)`
+
+## Options
+
+- `DiscoverEvent.DISCOVER (event: DiscoverEvent)`
+
+## Conversion
+
+### General Events
+
+- `Converter.EVENT_BEGIN (context: Context)`:
+  Triggered when the converter begins converting a project.
+- `Converter.EVENT_END (context: Context)`:
+  Triggered when the converter has finished converting a project.
+
+### Factory Events
+
+- `Converter.EVENT_FILE_BEGIN (context: Context, project: ProjectReflection, node: TypeScript.SourceFile)`:
+  Triggered when the converter begins converting a source file.
+- `Converter.EVENT_CREATE_DECLARATION (context: Context, declaration: DeclarationReflection, node?: TypeScript.TypeReferenceNode)`:
+  Triggered when the converter has created a declaration reflection.
+- `Converter.EVENT_CREATE_SIGNATURE (context: Context, signature: SignatureReflection, node: TypeScript.SignatureDeclaration)`:
+  Triggered when the converter has created a signature reflection.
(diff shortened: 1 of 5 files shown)


mootari:
> @mootari I copied the gist you posted last year: #595 (comment) Is that ok?

Fine by me, although I'd rather see the information added to the wiki (if @aciccarello would be willing to enable it for this repo).

Two things to keep in mind:
- Most of the component descriptions have been pulled directly from the source (and only slightly altered where necessary).
- The gist hasn't been updated in quite a while. The information would need to be checked against the current state of typedoc.

docs/developer/options.md
```diff
@@ -0,0 +1,53 @@
+*Copied from https://gist.github.com/mootari/d39895574c8deacc57d0fc04eb0d21ca*
+*Original author: https://github.com/mootari*
+
+# Component options
```
aciccarello: What about adding this to the `Options` TSDoc?

https://github.com/TypeStrong/typedoc/blob/1c4c5151cec938ef03cae79c33c12be8d587ea74/src/lib/utils/options/options.ts#L62-L67

docs/developer/events.md
```diff
@@ -0,0 +1,70 @@
+*Copied from https://gist.github.com/mootari/d39895574c8deacc57d0fc04eb0d21ca*
+*Original author: https://github.com/mootari*
+
+# Events
```
aciccarello: What about adding this to the `Event` TSDoc comment. You could also add links to the other events listed.

hinell:
@mootari Is the PR still alive? It would be best to have a script that would do this every time you build it.

mootari:
@hinell No idea, I didn't open it. Looks rather dead, though.

cspotcode:
Is anyone opposed to merging this as-is?  It is meant to be a work-in-progress, improved gradually over time.  All of the suggestions in this thread can be implemented via follow-up PRs.

Ultimately, the goal of this PR is to help developers.  In that sense, if it is an improvement over the status quo, I think it should be merged.

mootari:
I'm currently not working with typedoc (and haven't in a while), so take the following with the appropriate amount of salt:
- The gist on which this PR is based is 2 1/2 years old. It was created during a time where typedoc underwent a major rewrite.
- My biggest fear would be that, once this gets committed, noone is going to take a second look. In my opinion any major clhanges should still happen in this PR.
- At the very, very least the docs need to be fact-checked to ensure that they contain no false information. If the docs are incomplete, that's bearable. But wrong information would be highly frustrating for devs.
- @aciccarello's comments about integrating the docs into the source still stand. Imo they shouldn't just get ignored.
- The sensible thing may be to split this PR up into several (one for each section), so that the easier docs can be embedded into the source, while the more complex stuff has a place to be discussed exclusively.

cspotcode:
I should clarify I don't have any time allocated for typedoc these days, so anyone who wants is free to run with this PR, close it, or whatever.

Gerrit0:
I've mostly just left this since parts of what it describes is going away, and parts have already changed to not be up to date. I still reference it occasionally though, so... I do want to pull at least some of these into comments.

Gerrit0:
Having looked at this a bit more, I'd rather point contributors to https://typedoc.org/guides/development/ for a high level overview of TypeDoc's architecture, and comments within the source for extra details. Thanks though!

PR #916, opened 2018-12-10
Fix indentation rules in .editorconfig

cspotcode:
I notice that the .editorconfig specifies 2-space indents but should be specifying 4.


Gerrit0:
Thanks!

PR #932, opened 2018-12-29
Fix serialize plugin

### LINKED ISSUE #930: Serializer plugin not working as expected
I could be approaching this incorrectly, but I have been tirelessly trying to get this to work to no avail. My aim is to add a custom key/value to the generateJson output schema. I started by adding the key to the reflection using a converter plugin on the Converter.EVENT_CREATE_DECLARATION event. See below for simplified version: 
```
declare class decReflection extends Reflection {
  test: string;
}

@Component({ name: "test-plugin" })
export class TestPlugin extends ConverterComponent {
 

  initialize() {
    this.listenTo(this.owner, {
      [Converter.EVENT_CREATE_DECLARATION]: this.onDeclaration,
    });
  }

  private onDeclaration(
    context: Context,
    reflection: decReflection,
    node?: any
  ) {
    if (!node) {
      return;
    }

    reflection.test = 'test value';
  }
}
```

It wasn't until I noticed that the output did not contain my custom key/value that, by process of elimination, I narrowed the problem down to serialization, which (apparently) determines the output schema. 

I feel like I am doing everything correctly when creating the serialization plugin, starting with adding the component. It seems as though for whatever reason the `support` function is not getting processed as it is advertised in the /lib/serialization/components.ts file. 

> Like [[Converter]] plugins each [[Serializer]] plugin defines a predicate that instructs if an
> object can be serialized by it, this is done dynamically at runtime via a `supports` method.

So here is my attempt at getting the serializer plugin to work, starting with the index file which adds the component: 

```
module.exports = function(PluginHost: any) {
  var app = PluginHost.owner;

  app.converter.addComponent("test-plugin", TestPlugin);
  app.serializer.addComponent("serializer:test-plugin", TestPluginSerializer);
};
```

then the serializer plugin class: 

```
@Component({ name: "serializer:test-plugin" })
export class TestPluginSerializer extends SerializerComponent<TestReflection> {
  static PRIORITY = 1000;

  serializeGroup = (instance: unknown): boolean => {
    return instance instanceof TestReflection;
  };

  supports = (t: unknown) => {
    return true;
  };

  serializeGroupSymbol = TestReflection;

  toObject(testReflection: TestReflection, obj?: any): any {
    obj = obj || {};

    if (testReflection.test) {
      obj.test = testReflection.test;
    }

    return obj;
  }
}
```

After much digging, I can see that the serializer gets added as a component but for whatever reason the support function is not called. 

Again, I am simply trying to add a key/value to the json schema generated by the generateJson method. I could be approaching this completely wrong, but if the serializer plugin is required then it is a lot less intuitive than it needs to be or something is buggy.

Any insight/help would be appreciated!

tilersmyth:
Fixes #930. Conditional comparing functions always returned falsey. Utilizing toString() on the functions fixes the issue.

### DIFF
--- a/src/lib/serialization/serializer.ts
+++ b/src/lib/serialization/serializer.ts
@@ -36,7 +36,7 @@ export class Serializer extends ChildableComponent<Application, SerializerCompon
       let match = this.router.get(component.serializeGroup);
 
       if (!match) {
-        match = Array.from(this.router.values()).find( v => v.symbol === component.serializeGroupSymbol)
+        match = Array.from(this.router.values()).find( v => v.symbol.toString() === component.serializeGroupSymbol.toString())
           || { symbol: component.serializeGroupSymbol , group: [] };
         this.router.set(component.serializeGroup, match);
         this.routes.push(component.serializeGroup);


Gerrit0:
I'm... wary about comparing stringified functions. This seems like a hack that corrects some other mistake in the serialization code. Why are the functions (instances?) different in the first place?

tilersmyth:
@Gerrit0 you could be right. This is the only fix I could come up with for the way the serializer currently adds plugins to the proper group. When the serializer `addComponent` function is executed it is comparing two distinct instances of the same function (object), which is why equality will never be true. 
Unless the serializer is refactored at a deeper level, using the .toString() method here might be adequate as we just want to check if it is the same function (it does not have to be same instance) to push it the right group for serialization.

tilersmyth:
@Gerrit0 you're right, this is not a great solution. After continuing to dig into the issue, it seems that the functions are not equal because the serialization plugin is referencing the class declaration where as internally the the `serializeGroupSymbol` is referencing the class (due to it being referenced prior to being compiled - I think). See output comparison below.. 

For my purposes of getting the serializer to work, I have now simply updated the `match` function to look for equality in `serializeGroupSymbol` names with the following: 

```
match = Array.from(this.router.values()).find(v => v.symbol.name === component.serializeGroupSymbol.name)
```
This has been working well, but let me know if you think this solution is worth a PR. Otherwise, I will just leave it alone until a better fix comes along. Thanks for your input on this. 


## Current serializeGroupSymbol output (Plugin vs. Internal)
**Plugin:** 
```
function Reflection(parent, name, kind) {
        this.name = '';
        this.flags = [];
        this.id = REFLECTION_ID++;
        this.parent = parent;
        this.name = name;
        this.originalName = name;
        this.kind = kind;
    }
```

**Internal:**
```
class Reflection {
    constructor(name, kind, parent) {
        this.name = '';
        this.flags = new ReflectionFlags();
        this.id = REFLECTION_ID++;
        this.parent = parent;
        this.name = name;
        this.originalName = name;
        this.kind = kind;
    }
    kindOf(kind) {
        const kindArray = Array.isArray(kind) ? kind : [kind];
        return kindArray.some(kind => (this.kind & kind) !== 0);
    }
    getFullName(separator = '.') {
        if (this.parent && !this.parent.isProject()) {
            return this.parent.getFullName(separator) + separator + this.name;
        }
        else {
            return this.name;
        }
    }
    setFlag(flag, value = true) {
        this.flags.setFlag(flag, value);
    }
    getAlias() {
        if (!this._alias) {
            let alias = this.name.replace(/[^a-z0-9]/gi, '_').toLowerCase();
            if (alias === '') {
                alias = 'reflection-' + this.id;
            }
            let target = this;
            while (target.parent && !target.parent.isProject() && !target.hasOwnDocument) {
                target = target.parent;
            }
            if (!target._aliases) {
                target._aliases = [];
            }
            let suffix = '', index = 0;
            while (target._aliases.indexOf(alias + suffix) !== -1) {
                suffix = '-' + (++index).toString();
            }
            alias += suffix;
            target._aliases.push(alias);
            this._alias = alias;
        }
        return this._alias;
    }
    hasComment() {
        return this.comment ? this.comment.hasVisibleComponent() : false;
    }
    hasGetterOrSetter() {
        return false;
    }
    getChildByName(arg) {
        const names = Array.isArray(arg) ? arg : arg.split('.');
        const name = names[0];
        let result;
        this.traverse((child) => {
            if (child.name === name) {
                if (names.length <= 1) {
                    result = child;
                }
                else if (child) {
                    result = child.getChildByName(names.slice(1));
                }
            }
        });
        return result;
    }
    isProject() {
        return false;
    }
    findReflectionByName(arg) {
        const names = Array.isArray(arg) ? arg : arg.split('.');
        const reflection = this.getChildByName(names);
        if (reflection) {
            return reflection;
        }
        else if (this.parent) {
            return this.parent.findReflectionByName(names);
        }
    }
    traverse(callback) { }
    toObject() {
        const result = {
            id: this.id,
            name: this.name,
            kind: this.kind,
            kindString: this.kindString,
            flags: {}
        };
        if (this.originalName !== this.name) {
            result.originalName = this.originalName;
        }
        if (this.comment) {
            result.comment = this.comment.toObject();
        }
        Object.getOwnPropertyNames(ReflectionFlags.prototype).forEach(name => {
            const descriptor = Object.getOwnPropertyDescriptor(ReflectionFlags.prototype, name);
            if (typeof descriptor.get === 'function' && this.flags[name] === true) {
                result.flags[name] = true;
            }
        });
        if (this.decorates) {
            result.decorates = this.decorates.map((type) => type.toObject());
        }
        if (this.decorators) {
            result.decorators = this.decorators.map((decorator) => {
                const result = { name: decorator.name };
                if (decorator.type) {
                    result.type = decorator.type.toObject();
                }
                if (decorator.arguments) {
                    result.arguments = decorator.arguments;
                }
                return result;
            });
        }
        this.traverse((child, property) => {
            if (property === TraverseProperty.TypeLiteral) {
                return;
            }
            let name = TraverseProperty[property];
            name = name.substr(0, 1).toLowerCase() + name.substr(1);
            if (!result[name]) {
                result[name] = [];
            }
            result[name].push(child.toObject());
        });
        return result;
    }
    toString() {
        return ReflectionKind[this.kind] + ' ' + this.name;
    }
    toStringHierarchy(indent = '') {
        const lines = [indent + this.toString()];
        indent += '  ';
        this.traverse((child, property) => {
            lines.push(child.toStringHierarchy(indent));
        });
        return lines.join('\n');
    }
}
```

Gerrit0:
Thanks for looking in to this more! Reviewing the serialization has been on my list for a while and this should be helpful in finding what should be done to fix this. Checking names is better than `.toString`, but I'd still like to resolve the root issue.

tilersmyth:
Ok cool! I will be following.

PR #922, opened 2018-12-19
Use typescript 3.2.x

Wenzil:

PR #934, opened 2018-12-31
update typescript to 3.2.x

just-boris:
The changes in destructuring handling are caused by this PR in Typescript: https://github.com/Microsoft/TypeScript/pull/27587

Rest parameters are defined by this code, so it now returns an object type with multiple type parameters (roughly as `Object<number, number>`), instead of `Array<number>` as it was before

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "progress": "^2.0.0",
     "shelljs": "^0.8.2",
     "typedoc-default-themes": "^0.5.0",
-    "typescript": "3.1.x"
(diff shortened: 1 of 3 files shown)


Gerrit0:
#922 also updates to 3.2.x, and changes the test behavior to display a tuple type instead of just `Object`, which I believe is more useful. Thanks for the link to the relevant TS PR!

just-boris:
I probably should have searched better... 
I will close this, as there is something in progress already

Issue #928, opened 2018-12-24
add an issue template


9oelM:
I see many issues that could possibly be very helpful, but some of them have no proper reproduction (like #925, #911, #904, #893, and more).  And it's also a time consuming job to reproduce it ourselves just by reading from the issue or ask each author to reproduce his/her problem.
 
The best thing to do, I believe, is to make a common issue template so that authors could know before posting that they should reproduce their problems as well in many cases. 

Probably an issue template would be some kind of this format:

## Steps to Reproduce the Problem

  1.
  1.
  1.

## Specifications

  - Version:
  - Platform:
  - Subsystem:

And the same thing for PR templates & Code of conduct. It would make it easy for people to contribute.

aciccarello:
👍 I agree a issue template would be helpful especially with bug reports.

My request is that it is simple/easy enough to use and understand that we don't trade _issue reports with little information_ for _issue reports with little information sprinkled in a lot of boilerplate_.

9oelM:
@aciccarello I cannot propose (make a PR) on the issue template. Guess it's only restricted to collaborators. Could you propose one instead? I guess a bit of tweaks would work if you would want to make the suggested format above simpler/easier.

Gerrit0:
Creating the file `ISSUE_TEMPLATE.md` should let you propose a template according to the [GitHub docs](https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/)

9oelM:
@Gerrit0 Cool! Thanks for pointing that out.

9oelM:
I will write one and submit a PR. I think better direction could be making multiple issue templates for bug reports and feature requests.

Issue #904, opened 2018-11-20
@category not working for functions within modules


geethika18:
**Requirement**
Within a module, there should be a way to classify/categorize the functions - ex: content creation functions, content editing functions etc.

Tried out something similar to what was mentioned in https://github.com/TypeStrong/typedoc/pull/566

But @category tag was taking effect only on classes.

geethika18:
Hi @aciccarello,

I have a ts file defined as below

namespace ABC{

/**
* @category abc.1
*/
export function1() {
}

/**
* @category abc.2
*/
export function2() {
}
...
}

My requirement is to have these categories produce different pages or even different sections within the same page. Please help me out with the same.

geethika18:
Can I get some help here?

Gerrit0:
I don't believe this is supported with the default theme yet, #866 might resolve this.

jonchardy:
Yes, see #866. I'm willing to help make some changes here, but need some input on desired behavior.

geethika18:
Hi @jonchardy ,

Any luck with this? Please let me know. Thanks.

geethika18:
Output generated 

![image](https://user-images.githubusercontent.com/13620937/50688220-76e73900-104a-11e9-8680-658f5112f9c3.png)

Expected output 

![image](https://user-images.githubusercontent.com/13620937/50688238-89617280-104a-11e9-8c6d-7d200f821b86.png)


The theme used is same, but in 2nd case I had to duplicate my input file which is basically a weird work-around. Please suggest me the right approach.

jonchardy:
See https://github.com/TypeStrong/typedoc/pull/938

geethika18:
Hi @jonchardy ,

What is the ETA for releasing this change?

jonchardy:
I don't know, I'm just a contributor. @aciccarello

aciccarello:
Sorry, there is no ETA at the moment. The PR will need to be reviewed. Additionally there are some issues with releasing themes at the moment.

geethika18:
Hi @aciccarello ,

Could you please share the ETA for releasing this change? I see that the change is merged. Thank you.

Issue #902, opened 2018-11-16
Feature Request: Instead of showing links to types/aliases/interfaces, show their information instead


balupton:
I am converting a fairly large package to typescript today, and got typedoc generations docs. However the documentation requires a lot more clicking around to get the information that the jsdoc solution had previously.

Here is the declarations:
https://github.com/bevry/safeps/blob/4c9497e2b9eb53903ba12598e192c5ce1b931af5/source/index.ts#L25-L98

As you can see, there is a lot of shared properties between the interfaces. For three of them, I've already made what typedoc calls type aliases for them, which have the jsdoc descriptions.

However, when I navigate the documenation, I must click through several steps before I get the information. Each screenshot/visual-box is an additional click with typedoc:

<img width="932" alt="screen shot 2018-11-16 at 11 52 37 pm" src="https://user-images.githubusercontent.com/61148/48622515-24e74700-e9fb-11e8-9a1c-f973a38896f8.png">
<img width="919" alt="screen shot 2018-11-16 at 11 52 55 pm" src="https://user-images.githubusercontent.com/61148/48622517-26b10a80-e9fb-11e8-93c5-94ebca18cd1d.png">
<img width="926" alt="screen shot 2018-11-16 at 11 53 02 pm" src="https://user-images.githubusercontent.com/61148/48622521-29abfb00-e9fb-11e8-8a18-fe828a716857.png">
<img width="933" alt="screen shot 2018-11-16 at 11 53 12 pm" src="https://user-images.githubusercontent.com/61148/48622526-2d3f8200-e9fb-11e8-836d-db229baa7f51.png">
<img width="940" alt="screen shot 2018-11-16 at 11 53 16 pm" src="https://user-images.githubusercontent.com/61148/48622528-2fa1dc00-e9fb-11e8-9c23-dea912631b31.png">
<img width="935" alt="screen shot 2018-11-16 at 11 53 20 pm" src="https://user-images.githubusercontent.com/61148/48622535-329ccc80-e9fb-11e8-9a5d-15c89be58505.png">

I really like the way JSDoc embeds documentation, like it has done here for a different project: http://master.editions.bevry.surge.sh/docs/global.html#Edition

Or here via documentation.js for an earlier version of that project http://master.safeps.bevry.surge.sh/docs/#.safeps.exec

aciccarello:
[...]

Issue #901, opened 2018-11-14
No reflection on object literals


williamkbentley:
From TypeDoc src\lib\utils\component.ts:
`const childMappings: {host: ChildableComponent<any, any>, child: Function}[] = [];`

Generates:
![image](https://user-images.githubusercontent.com/38542573/48509384-f4e34c80-e81e-11e8-9d59-0245a93c475d.png)

No information on the object literal, just _object_.

Likewise on this example:
`/** Email details */  `
`public email?: { id: string, isVerified?: boolean }`

Generates:
![image](https://user-images.githubusercontent.com/38542573/48510069-d2eac980-e820-11e8-8109-fe086a1635e1.png)

Again, just _object_. Is it possible to get reflection information on the object literal?

9oelM:
It's partly working. See below:

![default](https://user-images.githubusercontent.com/22465806/50904587-03876200-1464-11e9-885e-7e3f02d98e6a.PNG)

And my test code to produce docs is this:
```ts
class Test {
    public email?: { id: string, isVerified?: boolean }
}

const childMappings: {host: number, child: Function}[] = [];

const test2 = {
    hi: "hi",
    test: "test"
}
```

We can see that `email` property is shown with its own object properties. Am I reproducing something wrong, or is typedoc updated to make some changes on this matter? 

On the other hand, `childMappings` would be rendered the same as what @williamkbentley suggested:
![image](https://user-images.githubusercontent.com/22465806/50904789-94f6d400-1464-11e9-81ab-f74793f2adaf.png)

So I really don't know what's happening here. I need some more verification...!? @williamkbentley

williamkbentley:
I think narrowed part of the issue down to `--excludeNotExported`. When that option is on the object literal properties are not included:
![image](https://user-images.githubusercontent.com/38542573/50918310-df993180-140d-11e9-9f85-8c8980aaaccf.png)

Same class without `--excludeNotExported`:
![image](https://user-images.githubusercontent.com/38542573/50918379-15d6b100-140e-11e9-8bef-eecff1aba680.png)

Using v0.14.0

9oelM:
Weird :worried: 
I will have a look at this if I have some time.

th0r:
Are there any news guys? I face the same problem and here is a minimal reproduction:

#### Test code
```ts
export interface MyEvent<Type extends string, Payload> {
  type: Type;
  payload?: Payload;
}

export type FooEvent = MyEvent<'foo', {
[...]

PR #935, opened 2019-01-01
Add issue templates

9oelM:
## Rationale
Currently, Typedoc does not have issue templates, so I have raised an issue on that (#928)
Many people are graciously giving contributions, but sometimes it's hard to understand their intentions or their problems. Therefore it seems that it's a better approach to make an issue template.

## Templates
* Typedoc's issues can largely be divided into two: bug reports (i.e. #893, #901, #904, ...) and suggestions (i.e. #898, #895, #902, ...). That's the reason I made separate templates for each.
* I believe these templates are fully in line with what @Gerrit0 is saying in [`CONTRIBUTING.md`](https://github.com/TypeStrong/typedoc/blob/master/CONTRIBUTING.md).

## Goal
I tried to lead contributors to the right direction but did not want to annoy them with a bunch of lines of comments. Plus, as @aciccarello said, I wanted to keep it as simple as possible for the risk of submissions of issues with almost no information hidden under an already-well-written template of length.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,8 +4,8 @@ about: Tell about typedoc's bug
 ---
 <!--
     Thank you for contributing to Typedoc! 
-    We ask you to give a description for each part,
-    so that you may provide enough information.
+    
+    First, please use a clear and descriptive title for the issue to identify the bug.
 
     Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
 -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,31 @@
+---
+name: Bug report
+about: Tell about typedoc's bug
+---
+<!--
+    Thank you for contributing to Typedoc! 
+    We ask you to give a description for each part,
+    so that you may provide enough information.
+
+    Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
+-->
+
+## Expected Behavior
+<!--
+    According to the code or documentation, how is Typedoc supposed to work?
+-->
+
+## Actual Behavior
+<!--
+    What does Typedoc fail to do?  
+-->
+
+## Steps to reproduce the bug
+<!--
+    Tell us how we could reproduce the bug. 
+-->
+
+## Specifications
+  - Typedoc version:
+  - Node.js version:
+  - OS:
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -10,6 +10,8 @@ about: Tell about typedoc's bug
     Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
 -->
 
+- [ ] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates
+
 ## Expected Behavior
 <!--
     According to the code or documentation, how is Typedoc supposed to work?
--- a/.github/ISSUE_TEMPLATE/suggestion.md
+++ b/.github/ISSUE_TEMPLATE/suggestion.md
@@ -4,9 +4,9 @@ about: Suggest an improvement or a feature
 ---
 <!--
     Thank you for contributing to Typedoc! 
-    We ask you to give a description for each part,
-    so that you may provide enough information.
-
+    
+    First, please use a clear and descriptive title for the issue to identify the suggestion.
+    
     Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
 -->
 
--- a/.github/ISSUE_TEMPLATE/suggestion.md
+++ b/.github/ISSUE_TEMPLATE/suggestion.md
@@ -10,6 +10,8 @@ about: Suggest an improvement or a feature
     Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
 -->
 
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+-->
+
+## Actual Behavior
+<!--
+    What does Typedoc fail to do?  
+-->
+
+## Steps to reproduce the bug
+<!--
+    Tell us how we could reproduce the bug.
```
aciccarello: Could we add some suggestions in this comment? How `typedoc` is invoked, configuration options, code that does not get documented correctly, minimal project to reproduce the issue.

.github/ISSUE_TEMPLATE/suggestion.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Suggestion
+about: Suggest an improvement or a feature
+---
+<!--
+    Thank you for contributing to Typedoc! 
+    
+    First, please use a clear and descriptive title for the issue to identify the suggestion.
+    
+    Please be as specific as possible! Provide codes, sreenshots, repositories, explanations, issues, links ... anything that could describe your point.
```
aciccarello: Minor typo
```suggestion
    Please be as specific as possible! Provide codes, screenshots, repositories, explanations, issues, links ... anything that could describe your point.
```

.github/ISSUE_TEMPLATE/suggestion.md
```diff
@@ -0,0 +1,23 @@
...
+-->
+
+- [ ] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates
+
+## Problem
+<!--
+    What is making you feel inconvenient without this improvement or feature?
```
aciccarello:
```suggestion
    What is missing or inconvenient without this improvement or feature?
```

.github/ISSUE_TEMPLATE/suggestion.md
```diff
@@ -0,0 +1,23 @@
...
+-->
+
+## Suggestion
+<!--
+    What do you suggest to fix the problem?
```
aciccarello:
```suggestion
    How do you suggest fixing the problem?
```

.github/ISSUE_TEMPLATE/bug_report.md
9oelM: Sure. I will add that.

9oelM:
```suggestion
    Tell us how we could reproduce the bug. Possible ways are providing:
    - Typedoc command (typedoc --out ...)
    - Configuration (typedoc.js or your custom option file)
    - Code that causes a bug in the documentation generated
    - Project (Github repository, CodePen, ...) where the issue can be reproduced.
```
Is this alright? I think there are a few more examples of how an issue can be reproduced, but I'm just afraid it's going to be too long.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+## Steps to reproduce the bug
+<!--
+    Tell us how we could reproduce the bug. Possible ways are providing:
+    - Typedoc command (typedoc --out ...)
+    - Configuration (typedoc.js or your custom option file)
+    - Code that causes a bug in the documentation generated
+    - Project (Github repository, CodePen, ...) where the issue can be reproduced.
+-->
+
+## Specifications
```
aciccarello: Would `Environment` be a better word for this heading?

.github/ISSUE_TEMPLATE/suggestion.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Suggestion
+about: Suggest an improvement or a feature
```
aciccarello: Could we rename this template "Feature Request"?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
+---
+name: Bug report
+about: Tell about typedoc's bug
```
aciccarello: I think this should automatically set the `bug` label
```suggestion
about: Tell about typedoc's bug
labels: bug
```

.github/ISSUE_TEMPLATE/bug_report.md
aciccarello: Sorry, didn't see this comment earlier. What you ended up with looks good.

.github/ISSUE_TEMPLATE/suggestion.md
9oelM: @aciccarello actually I was hesitating for minutes on that issue. For me, "Feature Request" sounds like... a narrower name. For example, #898, #935, #917 could be suggestions, but not feature requests. Feature requests sound like only an improvement only in typedoc code to me. Is that the only case for me?

If you get what I mean, "Suggestion" sounds like a superset of "Feature Request", which might embrace more issues.

What are your thoughts?

.github/ISSUE_TEMPLATE/bug_report.md
9oelM: Yeah. Both are fine, actually

.github/ISSUE_TEMPLATE/suggestion.md
aciccarello: That's a good point. I still lean towards "Feature Request" since it is a more standard term and fits the contents of the template better. We could possibly add a more general option in the future.

.github/ISSUE_TEMPLATE/bug_report.md
9oelM:
```suggestion
## Environment
```

.github/ISSUE_TEMPLATE/suggestion.md
9oelM: Yeap!

.github/ISSUE_TEMPLATE/suggestion.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Suggestion
```
9oelM:
```suggestion
name: Feature Request
```

Issue #866, opened 2018-10-03
Suggestion: Category improvements


rbuckton:
When categorization was added in #566, there was nothing added to the built-in templates to make categories useful. The suggestion in https://github.com/TypeStrong/typedoc/pull/566#issuecomment-332204971 provides some guidance, but overall the solution has some issues:

- There is no category for uncategorized reflections.
- Uncategorized reflections are lost using the suggested template.
- Categories override groups with this approach.

As such, I would like to suggest the following improvements to categories:

1. Add categories to groups. This would let the built-in template organize by group and then category.
1. Support an "Other" category (possibly configurable). This would be the category to place uncategorized reflections when there are other reflections that have a category.
1. Support category sorting via configuration/options.
1. Consider supporting _multiple_ categories for an item (nice to have, but not strictly necessary).

I ended up implementing (1), (2), and (3) in a custom plugin I wrote for my project to work around a number of other issues:

- Add an "Other" category when needed:
  - https://github.com/rbuckton/iterable-query/blob/24f9010f33135008c57d8bd391d518057bef06d3/src/typedoc-plugin-custom/plugin.ts#L33 
- Add categories to groups:
  - https://github.com/rbuckton/iterable-query/blob/refactor/src/typedoc-plugin-custom/plugin.ts#L356
  - https://github.com/rbuckton/iterable-query/blob/refactor/src/typedoc-plugin-custom/plugin.ts#L400
- Index by group and category in theme:
  - https://github.com/rbuckton/iterable-query/blob/refactor/src/typedoc-plugin-custom/theme/partials/index.hbs#L8

Here's an example:
![categories](https://user-images.githubusercontent.com/3902892/46624363-61af4c80-cae5-11e8-9d70-a284e4b5ab97.png)

jonchardy:
Since I worked on the original categories, I'm willing to help out here. It would be nice to get an idea of where people want to see this go.

I can see different methods of categorization (and this is why I didn't implement anything on the theme side, since at the time I knew people would want to do different things with categories):
1) Lump categories, which include reflections from different groups. This is something we do in our index, where we have broad categories that include both classes and interfaces. I think this is most useful at the project level, as lump categories don't make as much sense once you're dealing with methods vs properties.
2) Group categories, as shown by @rbuckton . This would allow classes, methods, etc. to be categorized _within_ their groups. Within a group, I imagine if all categories fall into the default category, we would just output the default group title, rather than including category information.

[...]

PR #938, opened 2019-01-04
Category improvements

jonchardy:
This PR makes a handful of changes based on requests and feedback from #866 and #904. I've also made a PR over at https://github.com/TypeStrong/typedoc-default-themes/pull/61 to make categories more integrated into the theme.

Thanks to @rbuckton for the suggestions. I used some of the ideas from your implementation in your custom plugin.

**Group Categories**
Categories can now be used within groups or as lump categories. Lump categories mean if you have a method and property you want together, you can do so. The option `--categorizeByGroup` controls this. By default, reflections will be categorized by group (within properties, methods, etc).

**Default Category**
Uncategorized reflections will be placed in the default category. If everything falls into the default category, categories aren't used. The option `--defaultCategory` controls this. By default, the default category will be 'Other'.

**Category Ordering**
Categories can be ordered however the user wants. Using an array, the user can specify the order categories are displayed on their pages. The option `--categoryOrder` controls this. By default, they will be ordered alphabetically. An entry of '*' indicates where categories not in this list will be rendered, defaulting to the end.

### DIFF
--- a/src/lib/models/reflections/project.ts
+++ b/src/lib/models/reflections/project.ts
@@ -1,7 +1,6 @@
 import { SourceFile, SourceDirectory } from '../sources/index';
 import { Reflection, ReflectionKind } from './abstract';
 import { ContainerReflection } from './container';
-import { ReflectionCategory } from '../ReflectionCategory';
 
 /**
  * A reflection that represents the root of the project.
@@ -26,11 +25,6 @@ export class ProjectReflection extends ContainerReflection {
      * A list of all source files within the project.
      */
     files: SourceFile[] = [];
-
-    /**
-     * All reflections categorized.
-     */
-    categories?: ReflectionCategory[];
 
     /**
      * The name of the project.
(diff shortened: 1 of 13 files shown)


src/lib/models/reflections/project.ts
```diff
@@ -123,25 +117,4 @@ export class ProjectReflection extends ContainerReflection {
 
         return undefined;
     }
-
-    /**
-     * Return a raw object representation of this reflection.
-     * @deprecated Use serializers instead
-     */
-    toObject(): any {
```
Gerrit0: Any reason this was deleted? The `.toObject` methods are still used in tests right now.

jonchardy: ProjectReflection extends ContainerReflection, which already outputs categories in its toObject method.

BenDevelopment:
Hi,

When will this merge be released?

PR #944, opened 2019-01-10
Don't limit option files to standard names

### LINKED ISSUE #942: 0.13 supported config paths do not work in 0.14.0

## Expected Behavior
My `--options` path is:

`node_modules/my-config-project/config/typedoc.config.json`

in 0.13 this worked fine

## Actual Behavior

In 0.14 this no longer loads the config, This makes 0.14 a breaking change despite not being marked as such and has broken 60+ projects...

## Environment
  - Typedoc version: 0.14
  - Node.js version: 8.11.1
  - OS: windows

Gerrit0:
I believe this should resolve any issues with files that were accepted in v0.13 being rejected in v0.14.

Fixes #942

### DIFF
--- a/src/lib/utils/options/readers/tsconfig.ts
+++ b/src/lib/utils/options/readers/tsconfig.ts
@@ -48,7 +48,7 @@ export class TSConfigReader extends OptionsComponent {
         if (TSConfigReader.OPTIONS_KEY in event.data) {
             const tsconfig = event.data[TSConfigReader.OPTIONS_KEY];
 
-            if (/tsconfig\.json$/.test(tsconfig)) {
+            if (FS.existsSync(tsconfig)) {
                 file = Path.resolve(tsconfig);
             } else {
                 file = ts.findConfigFile(tsconfig, ts.sys.fileExists);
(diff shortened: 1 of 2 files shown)


src/lib/utils/options/readers/typedoc.ts
```diff
@@ -63,8 +63,8 @@ export class TypedocReader extends OptionsComponent {
      *   typedoc file will be attempted to be found at the root of this path
      * @return the typedoc.(js|json) file path or undefined
      */
-    findTypedocFile(path: string = process.cwd()): string | undefined {
-        if (/typedoc\.js(on)?$/.test(path)) {
+    findTypedocFile(path: string): string | undefined {
+        if (FS.existsSync(path) && FS.statSync(path).isFile()) {
```
aciccarello: 👍 That's much more flexible

src/lib/utils/options/readers/tsconfig.ts
```diff
@@ -59,8 +59,14 @@ export class TSConfigReader extends OptionsComponent {
                 return;
             }
         } else if (TSConfigReader.PROJECT_KEY in event.data) {
-            // The `project` option may be a directory or file, so use TS to find it
-            file = ts.findConfigFile(event.data[TSConfigReader.PROJECT_KEY], ts.sys.fileExists);
+            const resolved = Path.resolve(event.data[TSConfigReader.PROJECT_KEY]);
+            // If the file exists, use it
+            if (FS.existsSync(resolved)) {
```
aciccarello: Does this need to check if it is a file too? What doesn't `ts.findConfigFile` do correctly?

Gerrit0: Yes, this should also check for a file, `ts.findConfigFile` searches for `tsconfig.json` (or other file passed in to the function - 3rd argument) and doesn't check if the first argument is a file.

PR #946, opened 2019-01-11
Update & unpin highlight.js

Gerrit0:
See https://github.com/TypeStrong/typedoc/pull/845#issuecomment-453517506

Causes a change to the test files since highlight-js's language detection has changed


webmaster128:
Thank you @Gerrit0 and @aciccarello!

PR #948, opened 2019-01-12
Expand JS files if allowJS is enabled

### LINKED ISSUE #911: Feature request: Support jsdoc style typescript comments in js files
Typescript supports a way to add jsdoc-style comments (that indicate type information) within js files (https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript). This is quite a valid way to use Typescript without moving to ts files.

But these files don't seem to get picked up or processed by typedoc to generate relevant documentation. Is it possible to support this?
(TS compiler api should already have ways to understand this comments, the way VSCode does)

Gerrit0:
Fixes #911

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -257,12 +257,13 @@ export class Application extends ChildableComponent<Application, AbstractCompone
             return exclude.some(mm => mm.match(fileName));
         }
 
+        const supportedFileRegex = this.options.getCompilerOptions().allowJs ? /\.[tj]sx?$/ : /\.tsx?$/;
         function add(dirname: string) {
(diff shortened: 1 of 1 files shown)


aciccarello:
Makes sense. Does this result in duplicate files if someone has their `.ts` files compiled in place (in the same directory) while using `allowJs`?

Gerrit0:
I suppose it could result in duplicate files, but, if I understand the `allowJs` option correctly, this shouldn't be an issue since typescript would complain that input files would be overwritten. With an `outDir` set, the compiled `.ts` files should be in a directory that isn't being expanded and thus this wouldn't result in duplicate files.

PR #949, opened 2019-01-13
Reduce reliance on internal TS apis

Gerrit0:
Summary of changes:

1. Removed the custom `ts.CompilerHost` implementation - it is not required when creating a program and we just passed the calls to `ts.sys` functions (and a few internal apis) anyways.
1. Replaced internal calls with equivalent code.

I'll also make a few inline comments where code was just removed since it was no longer necessary.


src/lib/converter/factories/comment.ts
```diff
@@ -32,14 +32,7 @@ export function createComment(node: ts.Node): Comment | undefined {
...
 function isTopmostModuleDeclaration(node: ts.ModuleDeclaration): boolean {
-    if (node.nextContainer && node.nextContainer.kind === ts.SyntaxKind.ModuleDeclaration) {
-        let next = <ts.ModuleDeclaration> node.nextContainer;
-        if (node.name.end + 1 === next.name.pos) {
-            return false;
-        }
-    }
-
-    return true;
+    return node.getChildren().some(ts.isModuleBlock);
```
Gerrit0: Toss `namespace A.B.C {}` into https://ts-ast-viewer.com/ and it's immediately obvious that a module declaration is "topmost" if it has a module block as an immediate child.

src/lib/converter/nodes/variable-statement.ts
```diff
@@ -42,10 +41,14 @@ export class VariableStatementConverter extends ConverterNodeComponent<ts.Variab
      * @param node     The binding pattern node that should be analyzed.
      */
     convertBindingPattern(context: Context, node: ts.BindingPattern) {
-        (node.elements as ts.NodeArray<ts.BindingElement>).forEach((element: ts.BindingElement) => {
+        node.elements.forEach((element) => {
             this.owner.convertNode(context, element);
 
-            if (_ts.isBindingPattern(element.name)) {
+            if (!ts.isBindingElement(element)) {
```
Gerrit0: This additional check is required since the internal `isBindingPattern` check also ensures that `node` exists, while the `is*BindingPattern` do not

src/lib/converter/nodes/variable.ts
```diff
@@ -30,6 +30,7 @@ export class VariableConverter extends ConverterNodeComponent<ts.VariableDeclara
 
     /**
      * Analyze the given variable declaration node and create a suitable reflection.
+     * TODO: the type of `node` is incorrect, it should be a union of ts.PropertySignature | ts.PropertyDeclaration | ...
```
Gerrit0: Leaving resolving this to a later PR since changing to the actual type caused quite a few type errors.

src/lib/utils/options/sources/typescript.ts
```diff
@@ -72,15 +72,6 @@ export class TypeScriptSource extends OptionsComponent {
                 }
         }
 
-        switch (option.paramType) {
```
Gerrit0: No longer necessary since this is only used when printing help output, and TypeDoc hasn't printed TypeScript options in help since #569 was resolved months ago.

src/lib/utils/options/readers/arguments.ts
```diff
@@ -5,6 +5,13 @@ import { Component } from '../../component';
 import { DiscoverEvent, OptionsComponent } from '../options';
 import { ParameterType } from '../declaration';
 
+enum CharacterCodes {
```
aciccarello: Should we add a source link?

src/lib/converter/nodes/variable-statement.ts
aciccarello: Would it be better to inline this with the if block below?

src/lib/converter/converter.ts
```diff
@@ -281,11 +274,12 @@ export class Converter extends ChildableComponent<Application, ConverterComponen
      * @param fileNames  Array of the file names that should be compiled.
      */
     convert(fileNames: string[]): ConverterResult {
+        // TODO: This mutates the input array, is this side effect used anywhere? Can it be removed?
```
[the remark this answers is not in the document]
aciccarello: It doesn't look like the side effects are used anywhere since this is passed into Application

PR #954, opened 2019-01-19
Replace array `indexOf` checks with `includes`

### LINKED ISSUE #898: Refactor to use array include
Minor change to make the code more readable. `Array.includes` is available in Node 6+ and can be used to replace the less-readable `[].indexOf !== -1` or `[].indexOf > -1`.

In general, `Array.includes` is a drop in replacement but there are some slight differences in edge cases (most notably when searching for `undefined`).

https://dev.to/adroitcoder/includes-vs-indexof-in-javascript

bendtherules:
Closes #898

Changes:

Replaced `indexOf` calls to check presence or absence with `includes` (with negation operator when checking absence).

Decisions:

1. There are some indexOf calls which use the index value to both check presence and use it to remove some elements at that position. They can be potentially refactored with separate calls to `includes` and `indexOf`, but I decided not to do that as `indexOf` alone can serve both the purposes here.

2. In testcases, replaced `assert.equal` + `indexOf` check with -1 by `assert.ok` + `includes`. This will probably show less information during error, as the actual index value won't be printed. That seems like a small price to pay.

Observations:

1. Most of the tests don't seem to fail if I change the presence or absence logic (`includes` or `!includes`) with the opposite one. I often did this mistake when working on this initially, but none of the test cases failed as they should have.
Maybe we should add some more tests around that?

### DIFF
--- a/src/test/typedoc.ts
+++ b/src/test/typedoc.ts
@@ -15,55 +15,55 @@ describe('TypeDoc', function() {
             const inputFiles = Path.join(__dirname, 'converter', 'class');
             const expanded = application.expandInputFiles([inputFiles]);
 
-            Assert.notEqual(expanded.indexOf(Path.join(inputFiles, 'class.ts')), -1);
-            Assert.equal(expanded.indexOf(inputFiles), -1);
+            Assert.ok(expanded.includes(Path.join(inputFiles, 'class.ts')));
+            Assert.ok(!expanded.includes(inputFiles));
         });
         it('expands input files', function() {
             const inputFiles = Path.join(__dirname, 'converter', 'class', 'class.ts');
             const expanded = application.expandInputFiles([inputFiles]);
 
-            Assert.notEqual(expanded.indexOf(inputFiles), -1);
+            Assert.ok(expanded.includes(inputFiles));
         });
         it('honors the exclude argument even on a fixed directory list', function() {
             const inputFiles = Path.join(__dirname, 'converter', 'class');
             application.options.setValue('exclude', '**/class.ts');
             const expanded = application.expandInputFiles([inputFiles]);
 
-            Assert.equal(expanded.indexOf(Path.join(inputFiles, 'class.ts')), -1);
-            Assert.equal(expanded.indexOf(inputFiles), -1);
+            Assert.ok(!expanded.includes(Path.join(inputFiles, 'class.ts')));
+            Assert.ok(!expanded.includes(inputFiles));
         });
         it('honors the exclude argument even on a fixed file list', function() {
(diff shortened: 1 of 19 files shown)


Gerrit0:
I'll need to take a closer look at the tests, since that makes it sound like something else is broken, but otherwise this looks great! Thanks!

src/test/typedoc.ts
```diff
@@ -15,55 +15,55 @@ describe('TypeDoc', function() {
             const inputFiles = Path.join(__dirname, 'converter', 'class');
             const expanded = application.expandInputFiles([inputFiles]);
 
-            Assert.notEqual(expanded.indexOf(Path.join(inputFiles, 'class.ts')), -1);
-            Assert.equal(expanded.indexOf(inputFiles), -1);
+            Assert.ok(expanded.includes(Path.join(inputFiles, 'class.ts')));
+            Assert.ok(!expanded.includes(inputFiles));
```
aciccarello: I think I'd prefer to drop the `.ok` here and use `Assert(condition)`

bendtherules:
> See the files I found below and let me know if there is a reason we should keep those as is.
> 
> * ComentPlugin.ts:66
> * DynamicModulePlugin.ts:57
> * converter/converter.ts:315

Thanks for noticing these. I tried to looked through `Array.indexOf` references, but not for other iterables. Have fixed this.

bendtherules:
In `test/utils.path.ts:22` 👍 
```ts
    it('Minimatch can match absolute paths expressions', () => {
      const paths = ['/unix/absolute/**/path', '\\windows\\alternative\\absolute\\path', 'C:\\Windows\\absolute\\*\\path', '**/arbitrary/path/**'];
      const mms = createMinimatch(paths);
      const patterns = mms.map(({ pattern }) => pattern);
      const comparePaths = [
        absolutePath('/unix/absolute/**/path'),
        absolutePath('/windows/alternative/absolute/path'),
        absolutePath('/Windows/absolute/*/path'),
        '**/arbitrary/path/**'
      ];

      Assert(isEqual(patterns, comparePaths), `Patterns have been altered:\nMMS: ${patterns}\nPaths: ${comparePaths}`);
```

Here the third path `'C:\\Windows\\absolute\\*\\path'` might not match with `absolutePath('/Windows/absolute/*/path'),` if the project directory is not in C drive. This is probably not going to be a problem in CI system, but can fail for other developers using a different drive letter.

This is failing in my local Windows setup. We might want to change it somehow (or make this optional?).

bendtherules:
Hi @aciccarello , do this changes look good?

aciccarello:
Sorry for the delay. These changes look great!

If you are having trouble running the tests on windows, could you please open another issue so we can get that fixed separately. Thanks!

Issue #880, opened 2018-10-08
Suggestion: Allow customization of TypeScript version


rbuckton:
While TypeDoc might not be able to process some AST nodes in newer versions of TypeScript, if a project is using features of a newer version of TypeScript, TypeDoc will fail to generate documentation because it is unable to compile the project.

One option would be to allow TypeDoc API users (such as gulp-typedoc) the ability to customize the version of TypeScript to use.

Gerrit0:
TypeDoc can break even on minor updates from TypeScript as it uses internal API features, #720. It would be neat to have an option to easily test switching to a different version though.

aciccarello:
Yeah, historically we've bundled TypeScript to avoid TypeScripts breaking changes. Since TypeDoc uses some private TypeScript apis it could theoretically break on any version. I'd like TypeDoc to move to the public apis but it isn't ready for that yet. Eventually, I think that TypeDoc should move to a peer dependency.

xaviergonz:
+1 I just removed typescript from typedocs node modules (since I needed it to use 3.6) and it correctly picked up my higher typescript version from a levels above

PR #965, opened 2019-02-04
fix: allow higher versions of typescript to be used

birkir:
This is a real problem.

If I installed a different version of typescript in a project, this package will overwrite the ./node_modules/.bin/tsc with typedoc's typescript version.

I'm even thinking if this should be a dependency at all, just a peerDependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "progress": "^2.0.0",
     "shelljs": "^0.8.2",
     "typedoc-default-themes": "^0.5.0",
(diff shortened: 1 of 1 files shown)


Gerrit0:
This is not feasible for TypeDoc right now since it relies on internal APIs and can break even with a minor release. See #880

birkir:
okay sure thing! there must be some solution to not let typedoc's dependency overwrite the tsc bin.

PR #966, opened 2019-02-05
patch: use marked 0.6.0 to support id generation for headers

Beraliv:
Release of [marked@0.6.0](https://github.com/markedjs/marked/releases/tag/v0.6.0)
Added [feature with Slugger](https://github.com/markedjs/marked/pull/1401)

**Problem:** README contains links to different headers which cannot be properly generated (Github is instead working just fine with it)

**Fix:** After updating version of `marked` id generation is automatically supported

**How to reproduce:** Get README where there are some relative links `#link-to-header`, use `typedoc` and see that `id` is `-` or something else

**After fix:** id is properly added and links are working fine like in Github


aciccarello:
Thanks for the contribution. Could you change a test to demonstrate this? Let us know if you have questions.

Beraliv:
@aciccarello yeah, sure
will provide today lately

Beraliv:
@aciccarello how can I point out that I use `marked`
I don't think it's plugin. However name of `Marked` is included in `MarkedPlugin` which is confusing a bit
And the constructor of `Application` has no working example yet (i'm working on it, right 👍)

If I use

```js
  app = new Application({
        mode:   'Modules',
        logger: 'none',
        target: 'ES5',
        module: 'CommonJS',
        experimentalDecorators: true
      });
```

it will provide useless output 

```json
{
  "id": 0,
  "kind": 0,
  "flags": {}
}
```

How can it be resolved?

aciccarello:
The file isn't converted because it is not a typescript file. Only the readme markdown file is parsed by default.

You'll probably want to change the `examples/basic/README.md` file added in #857 to test.

Beraliv:
@aciccarello I think it's ready

src/test/renderer/specs/index.html
```diff
@@ -81,6 +81,15 @@ <h3 id="arguments">Arguments</h3>
 					<li><code>--example &lt;project&gt;</code><br>
 					Example explanation.</li>
 				</ul>
+				<h2 id="example-with-cyrillic">Example with Cyrillic</h2>
```
Beraliv: first case, only Latin, should be with hyphen as it was

src/test/renderer/specs/index.html
```diff
@@ -81,6 +81,15 @@ <h3 id="arguments">Arguments</h3>
...
+				<h3 id="пример-api">Пример API</h3>
```
Beraliv: second case, combined Latin and Cyrillic
it would be `-api` without patching

src/test/renderer/specs/index.html
```diff
@@ -81,6 +81,15 @@ <h3 id="arguments">Arguments</h3>
...
+				<blockquote>
+					<p>Combined Latin and Cyrillic text</p>
+				</blockquote>
+				<h3 id="только-кириллица">Только кириллица</h3>
```
Beraliv: third case, only Cyrillic
it would be `-` without patching

Beraliv:
@aciccarello is everything good?

aciccarello:
Sorry for the delay. Somehow I missed your updates. Looks great!

sirisak1985:
request

PR #970, opened 2019-02-08
feat(cli): if no out or json cmd, put out to ./docs

### LINKED ISSUE #746: default out option
Typedoc requires either an `--out` or `--json` argument to run. It should have a reasonable default option (e.g. `--out docs`) to make getting started easier.

Suggested by @ice1080

artentica:
I not really happy with this solution but idk how do another way.
If you have an idea i'll be happy to implement it

Closes #746

### DIFF
--- a/src/lib/cli.ts
+++ b/src/lib/cli.ts
@@ -7,41 +7,41 @@ import { ParameterHint, ParameterType } from './utils/options/declaration';
 import { getOptionsHelp } from './utils/options/help';
 
 export const enum ExitCode {
-    OptionError  = 1,
+    OptionError = 1,
     NoInputFiles = 2,
-    NoOutput     = 3,
+    NoOutput = 3,
     CompileError = 4,
-    OutputError  = 5
+    OutputError = 5
 }
 
 export class CliApplication extends Application {
     @Option({
-        name:  'out',
-        help:  'Specifies the location the documentation should be written to.',
-        hint:  ParameterHint.Directory
+        name: 'out',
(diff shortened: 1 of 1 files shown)


src/lib/cli.ts
```diff
@@ -74,6 +71,11 @@ export class CliApplication extends Application {
                 if (this.json) {
                     this.generateJson(project, this.json);
                 }
+                if (!this.out && !this.json) {
+                    this.logger.log("You must either specify the 'out' or 'json' option.");
```
aciccarello: If we are automatically setting a default "must" sounds strong here.
```suggestion
                    this.logger.log("No 'out' or 'json' option has been set");
```

src/lib/cli.ts
```diff
@@ -74,6 +71,11 @@ export class CliApplication extends Application {
...
+                    this.logger.log("The './docs' directory have been set in default output");
```
aciccarello:
```suggestion
                    this.logger.log("The './docs' directory has be set as the output location by default");
```

artentica: Of course, I correct that and repush
Sorry for the bad grammar :/
Thx for your suggestions =)

artentica:
I accept the modification (directly on github (didn't know it was possible ^^))

artentica:
I'm not sure i ask: I cannot merge even if you (@aciccarello) approved no ?

aciccarello:
I'm leaving it open in case someone else wanted to chime in. I'll merge before the next release.

PR #973, opened 2019-02-12
add documentation contributing instructions

aciccarello:
Small section in CONTRIBUTING.md indicating where documentation comes from. Main improvement is that we now explicitly point to [TypeStrong/typedoc-site](https://github.com/TypeStrong/typedoc-site/). I also thought it was worthwhile to note our other documentation sources.

While writing this up, I was reminded that options are documented in three different areas with varying completeness. The website now includes the output of `typedoc --help` but it is unclear what we are putting on the website vs the readme.


aciccarello:
@Gerrit0 Technically, the readme is already in the api docs.

I think of the 3 different locations like this
- Option help: short description
- Readme: Easily accessible docs
- Website: Arguments site shows more detail

I know a lot of people only look at the readme and the website currently doesn't show much information beyond the help input so I'm not sure the best way to keep everything aligned.

Issue #972, opened 2019-02-11
Add links to headers


Gerrit0:
It is possible to link to a specific part of the page using anchors, for example: https://typedoc.org/guides/doccomments/#files

However, getting the link requires either guessing at the anchor, or viewing the source to grab the header's ID. It would be great if anchors were clickable like in the [ts-morph](https://dsherret.github.io/ts-morph/) documentation

igncp:
Hi, I would like to take over this issue if ok 🙂 

I have it working here: https://github.com/TypeStrong/typedoc/compare/master...igncp:feat/add-links-in-rendered-content?expand=1

But still needs some styling, testing and polishing. Cheers

Gerrit0:
Looking good! I'm kind of surprised that works, I thought the header had to have an ID to be linked to in the page.

igncp:
Correct, using ID seems to be the recommended way to do this. Updated it, working but still not ready for PR. Thanks for the info.

PR #981, opened 2019-03-05
Add links to README.md headings

igncp:
Hi 🙂 

This would add links without underline or special colour to the files parsed by `marked`, fixing (partially?) this issue: https://github.com/TypeStrong/typedoc/issues/972

Was thinking that it may be nice to also add links in the headings in the generated API pages, e.g. `Properties`, `Functions`, etc. , and in each of the elements titles (e.g. `methodA`). The last ones are already anchors although they are still using the `name` attribute, for example:

```
<a name="method-a" class="tsd-anchor"></a>
<h3>methodA</h3>
```

However, I think that would involve updating this repository: https://github.com/TypeStrong/typedoc-default-themes with a different approach, and not sure if this issue was referring to that also.

Many thanks for the library and happy to update any code from comments

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -37,3 +37,11 @@ $ examples --out path/to/documentation/ path/to/example/project/
 ### Только кириллица
 
 > Only Cyrillic text
+
+## Repeated Heading
+
+This is used to check the generated id. First instance.
+
+## Repeated Heading
+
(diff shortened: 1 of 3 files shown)


Gerrit0:
Thanks! Still leaving #972 open as my intent there was to add links in the typedoc-site, but this is a good change to make.

Issue #521, opened 2017-06-01
Create a "How to write a typedoc plugin" for boosting community plugins


nvdnkpr:
I don't know if typedoc is currently plugin-ready, meaning the code structure exporting the right classes and methods for 3rd party developers, but since there already exist a few nice plugins it would be good to have a short "Typedoc plugin 101" so one can write our own plugins in order to boost company.

samvv:
I would be 100% for this. Making it entirely pluggable would unleash the full power of the community. I'd be willing to give it a few pull requests, as soon as I have a better understanding of the code base.

Lanfei:
Looking forward to have it

BenDevelopment:
Same for me, I'm looking for a way to write a local plugin for my project. It would be nice to specify a single plugin ts file.

Gerrit0:
You can specify plugins to be loaded with `--plugin /absolute/path/to/my-plugin.js`

I think this issue needs to wait for a stable release - or at least until the `@Component` decorator is removed since that will change everything anyways.

jayarjo:
Any progress on this? Existing plugins get outdated and it becomes hard to figure out what the process for creating one is.

Gerrit0:
I have started https://typedoc.org/guides/development/, which focuses on developing TypeDoc proper, but will likely be useful for plugin developers.

I am in the middle right now of making library mode a thing, which has required changing quite a lot of the codebase (okay, I admit some didn't *have* to be changed, but I've wanted to for a while, and while I'm breaking everything anyways....) as a part of this, I'm taking good notes for what plugin developers need to know. I'll publish a hello world of plugin development guide as a draft once it is ready for a PR.

frog-o:
gerrit0 wrote: 
>I have started https://typedoc.org/guides/development/, which focuses on developing TypeDoc proper, but will likely be useful for plugin developers.

 and 

> I am in the middle right now of making library mode a thing, which has required changing quite a lot of the codebase (okay, I admit some didn't have to be changed, but I've wanted to for a while,

if it is not to much to ask for can someone explain: 
- what typedoc proper is.
- what library mode is 
and 
- what need to been done yet to get a full documented plugin guide pushed to main repo (it would be great to have a check list)

IMHO time of arrivals are always frown upon because there almost impossible to get even close but a checklist just show what done and show about the same info.

Gerrit0:
> what typedoc proper is.

[...]

PR #988, opened 2019-03-15
Add option to include modules in the top level toc

unstubbable:
Together with [`typedoc-plugin-monorepo`](https://www.npmjs.com/package/typedoc-plugin-monorepo) this enables monorepos to generate combined API docs for their packages using "external modules".

[Example](https://deploy-preview-430--feature-hub.netlify.com/@feature-hub/)

### DIFF
--- a/src/lib/converter/components.ts
+++ b/src/lib/converter/components.ts
@@ -6,7 +6,7 @@ import { Type } from '../models/types/abstract';
 import { Context } from './context';
 import { Converter } from './converter';
 
-export {Component};
+export { Component };
 
 export abstract class ConverterComponent extends AbstractComponent<Converter> { }
 
--- a/src/lib/output/components.ts
+++ b/src/lib/output/components.ts
@@ -5,7 +5,7 @@ import { ProjectReflection, DeclarationReflection } from '../models/reflections/
 import { Renderer } from './renderer';
 import { RendererEvent, PageEvent } from './events';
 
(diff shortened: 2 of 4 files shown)


aciccarello:
Thanks for the contribution. However, I'm not sure that it makes sense to add this just for a single plugin. Can you explain why this is necessary?

unstubbable:
Hmm, I thought it could be useful in general to have the option to list modules as top level toc entries. I just mentioned the plugin because this option would be especially helpful in the context of libraries that publish multiple packages under a common scope and are organized in a monorepo, which seems to become more and more common (e.g. Babel).

If you don’t agree, is there the possibility to replace the TocPlugin with a custom one?

aciccarello:
I believe you should be able to listen to the same event as the `TocPlugin` and add items to the page navigation that way. Also, you should be able to add the option decorator within the plugin.

I like the idea but I generally try to keep the core as simple as possible. If you have success with making a plugin, please tag it on npm and share a link here so others can find it. As a side note, we could use help writing a guide to creating plugins if you are interested (#521).

unstubbable:
I published a replacement of the `TocPlugin` as part of [my fork of the aforementioned `typedoc-plugin-monorepo`](https://www.npmjs.com/package/@strictsoftware/typedoc-plugin-monorepo). It lists modules as top-level entries in the table of contents.

PR #990, opened 2019-03-19
Mixin support

SamuraiJack:
This PR adds support for mixins, as per https://www.bryntum.com/blog/the-mixin-pattern-in-typescript-all-you-need-to-know/

The most important changes:
- When doing "inheritance" in class and interface converters, consider the possibility for the base type to be intersection type: `Type1 & Type2`. In such case, "inherit" from every component of such type
- When doing reflections merge, do not merge the reflections for type with reflections for values. This is because TypeScript has different namespaces for values and types and its perfectly valid to have different things with the same name:
```ts
export type Some = number
export const Some = () => {}
```
- bumped TypeScript dependency to 3.3.x

### DIFF
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -19,6 +19,7 @@ const nonStaticKinds = [
  */
 const nonStaticMergeKinds = [
     ts.SyntaxKind.ClassDeclaration,
+    ts.SyntaxKind.ClassExpression,
     ts.SyntaxKind.InterfaceDeclaration
 ];
 
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -93,11 +93,11 @@ export function createDeclaration(context: Context, node: ts.Declaration, kind: 
         }
     }
 
-    // Check if we already have a child with the same name and static flag
+    // Check if we already have a child of the same kind, with the same name and static flag
     let child: DeclarationReflection | undefined;
     const children = container.children = container.children || [];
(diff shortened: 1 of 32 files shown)


aciccarello:
Awesome! I wonder if this will help with any other reflection mergin issues we have.

aciccarello:
So the build passed on Node 10 but not node 6 because of a tslint error. That seems strange...

```
ERROR: /home/travis/build/TypeStrong/typedoc/src/lib/converter/components.ts:9:9 - missing whitespace
ERROR: /home/travis/build/TypeStrong/typedoc/src/lib/converter/components.ts:9:18 - missing whitespace
ERROR: /home/travis/build/TypeStrong/typedoc/src/lib/output/components.ts:8:9 - missing whitespace
ERROR: /home/travis/build/TypeStrong/typedoc/src/lib/output/components.ts:8:18 - missing whitespace
```

Sorry it's so grouchy. 👿

SamuraiJack:
I was not changing those files, but added whitespace anyway, hopefully build will pass this time :)

src/lib/converter/factories/declaration.ts
```diff
@@ -158,6 +159,20 @@ function setupDeclaration(context: Context, reflection: DeclarationReflection, n
     return reflection;
 }
 
+// we should not be merging type and value with the same name,
+// because TypeScript have different namespaces for these two categories
```
aciccarello: Slight grammar adjustment
```suggestion
// because TypeScript has different namespaces for these two categories
```

SamuraiJack:
@aciccarello Thank you!

PR #997, opened 2019-03-28
Add excludeNotExported to README

fc:
The README was missing something I needed. The README is actually incomplete since there are several other arguments not shown... so I've sent a link to the TypeDoc page.

It may be easier to just copy/paste the output of `typedoc --help` to this README to keep it more easily in sync instead but the text wrapping isn't so good.


aciccarello:
Thanks for the contribution. Yeah, we should find a better way to manage this. We also need to keep the website updated to so it does take some maintenance work.

PR #1006, opened 2019-04-05
Reset the state of the comment plugin on conversion start.

### LINKED ISSUE #828: kindString dropped in JSON output changes
The last time the the specs were updated there were a number of spec files where the kindString property was dropped in the JSON output. This is most likely a bug.

```diff
diff --git a/src/test/converter/type-operator/specs.json b/src/test/converter/type-operator/specs.json
index dd6f81c..f103dfc 100644
--- a/src/test/converter/type-operator/specs.json
+++ b/src/test/converter/type-operator/specs.json
@@ -18,7 +18,6 @@
           "id": 5,
           "name": "GenericClass",
           "kind": 128,
-          "kindString": "Class",
           "flags": {
             "isExported": true
           },
```

lddubeau:
In the test suite, a single app is reused for all the converter tests. If the plugins do not appropriately reset their state at the start of a conversion, then one test may affect later tests.

`CommentPlugin` did not reset its `hidden` field. This caused all converter tests after the `comment` test to be messed up. In all subsequent tests, the reflections with `id` 4 and 5 were deleted from `project.reflections`. This prevented plugins that follow `CommentPlugin` from acting on reflections 4 and 5. They were still part of the reflection *tree* however, and were still output, just with incomplete or incorrect values. For instance, `kindString` was not populated by the `GroupPlugin`, and `fileName` was not trimmed by the `SourcePlugin`.

Commit 3716ba7918bce8165e864d40ebf365cded3e3ee1 incorrectly modified all the tests that were failing due to the incorrect code so that they'd conform to the incorrect output. This PR reverses most of that commit.

Closes #828.

### DIFF
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -124,6 +124,7 @@ export class CommentPlugin extends ConverterComponent {
      * @param context  The context object describing the current state the converter is in.
      */
     private onBegin(context: Context) {
+        this.hidden = undefined;
         this.comments = {};
     }
 
--- a/src/test/converter/constructor-properties/specs.json
+++ b/src/test/converter/constructor-properties/specs.json
@@ -125,6 +125,7 @@
               "id": 4,
               "name": "x",
               "kind": 1024,
+              "kindString": "Property",
               "flags": {
                 "isConstructorProperty": true,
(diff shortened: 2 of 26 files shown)


aciccarello:
Thanks for finding this. It would be great to eventually rework the testing model to avoid issues like this.

Issue #991, opened 2019-03-21
@category not rendered properly


wangjie-agora:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

Just tried the latest category tag, and the interfaces/functions/methods can be grouped on the index, but the @category tags and the category names still appear in the descriptions of each interface/function/method.

![image](https://user-images.githubusercontent.com/40165471/54742403-469c4900-4bfc-11e9-911e-1bf3cbd141de.png)

### The ts code

```
    /**
     * The Client object returned by the {@link createClient} method provides access to much of the core AgoraRTC functionality.
     * @category Core
     */
    interface Client
```

### Typedoc config

   ```
 "typedocOptions": {
        "mode": "file",
        "hideGenerator": true,
        "defaultCategory": "Miscellaneous",
        "categorizeByGroup": true,
        "includes": "types",
        "excludeExternals": true,
        "includeDeclarations": true,
        "target": "ES6",
}
```

## Environment
  - Typedoc version: 0.14.2
  - Node.js version: 11.12.0
  - OS: macOS Mojave

Gerrit0:
> Typedoc version: 0.14.2

The category changes merged to master 2 days ago have not yet been released. Are you running on master or from npm?

wangjie-agora:
> > Typedoc version: 0.14.2
> 
> The category changes merged to master 2 days ago have not yet been released. Are you running on master or from npm?

Running on master. I can get the index in the categories I set.

Gerrit0:
Oh, gotcha, I misread the initial report. Yea, this is a bug. It should be a fairly simple fix, we just have to add `category` to the list of tags that won't be rendered.

jonchardy:
Included a fix for this in https://github.com/TypeStrong/typedoc/pull/1008. It will remove the comment tag when the category is resolved.

PR #1008, opened 2019-04-10
Category fixes

jonchardy:
Fix an issue where lump categorization could put everything into the default category instead of not using categories.

Also fix https://github.com/TypeStrong/typedoc/issues/991 by splicing the category tag from the comment when the category is resolved.


jonchardy:
Hmm, I think because that would involve changing a runtime argument, it would require an entire new test description in https://github.com/TypeStrong/typedoc/blob/master/src/test/converter.test.ts. What do you think, worth adding?

aciccarello:
Ah, that's true. In that case, we'll hold off on testing that until we get some test helpers to help with different configs.

PR #1014, opened 2019-04-16
update typescript

just-boris:
Bring the latest typescript version in.

### DIFF
--- a/src/lib/converter/plugins/SourcePlugin.ts
+++ b/src/lib/converter/plugins/SourcePlugin.ts
@@ -104,11 +104,14 @@ export class SourcePlugin extends ConverterComponent {
         if (!reflection.sources) {
             reflection.sources = [];
         }
+
+        const sources = reflection.sources;
+
         if (reflection instanceof DeclarationReflection) {
             file.reflections.push(reflection);
         }
 
-        reflection.sources!.push({
+        sources.push({
             file: file,
             fileName: fileName,
             line: position.line + 1,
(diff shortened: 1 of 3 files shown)


src/lib/converter/plugins/SourcePlugin.ts
```diff
@@ -104,11 +104,14 @@ export class SourcePlugin extends ConverterComponent {
...
         }
+
+        const sources = reflection.sources;
+
         if (reflection instanceof DeclarationReflection) {
             file.reflections.push(reflection);
         }
 
-        reflection.sources.push({
+        sources.push({
```
just-boris: This change is needed, because `file.reflections.push(reflection)` can potentially mutate `sources` key. 

I guess, typescript 3.4 became smarter and requires us to have an explicit stored reference to the array.

just-boris:
@aciccarello do you need anything else from me before this PR can be merged?

aciccarello:
@just-boris Thanks for identifying the new compilation errors. Hopefully I've found a way to avoid the variable assignment. Honestly is surprised me that typedoc complained about the existing code but then accepted that the array would be defined. Once this passes CI we should be able to merge.

just-boris:
Awesome! Thank you!

webmaster128:
Great stuff! @aciccarello could you release typedoc with typescript 3.4 support?

styu:
@aciccarello just curious when you plan on releasing again? Would love to use this with TS 3.4!

Gerrit0:
@styu a release is in progress, there appear to be a couple problems with the default theme which we are trying to address.

ftrimble:
Seems like you guys should mark this as a peer dependency so that you don't have to make a pr every time typescript releases a new version, and to prevent blocking people using this.

Any updates on when this will get released?

PR #1022, opened 2019-04-23
Fix reference to deprecated handlebars type

adidahiya:
from `handlebars.d.ts`:

```ts
// NOTE: for backward compatibility of this typing
type HandlebarsTemplateDelegate<T = any> = Handlebars.TemplateDelegate<T>;
```

turns out it's not actually backwards compatible because ~they haven't declared it as an ambient global type~ of subtle differences in type acquisition (see below). but anyway, typedoc should use the new namespaced type. I've gone ahead and done that in this PR.

unblocks https://github.com/palantir/documentalist/pull/86
cc @styu 

---

edit: more info

Another way to solve this would be with an `import "handlebars";` statement in any typedoc module which references a global handlebars symbol like `HandlebarsTemplateDelegate`. But adding import statements just for type acquisition side effects is bad.

When handlebars switched to publishing its own types, they did a straight port of `@types/handlebars`. However, the semantics for acquiring types from `node_modules/@types/handlebars` is a little different from `node_modules/handlebars`. TypeScript will automatically pull in all ambient types from `node_modules/@types` and make symbols like `HandlebarsTemplateDelegate` available, but the same is not true for types bundled in a regular npm package.

Related to this (but external to this repo), there's no need for `declare module` statements in [handlebars' types](https://github.com/wycats/handlebars.js/blob/3ce4425056998c1db921d7b6c0662b843acb593b/types/index.d.ts) anymore. The type system will acquire types from the same location as the module system, so `declare module "handlebars"` is unnecessary.

### DIFF
--- a/src/lib/output/events.ts
+++ b/src/lib/output/events.ts
@@ -1,3 +1,4 @@
+import { TemplateDelegate } from 'handlebars';
 import * as Path from 'path';
 
 import { Event } from '../utils/events';
@@ -110,7 +111,7 @@ export class PageEvent extends Event {
     /**
      * The template that should be used to render this page.
      */
-    template?: HandlebarsTemplateDelegate;
+    template?: TemplateDelegate;
 
     /**
      * The name of the template that should be used to render this page.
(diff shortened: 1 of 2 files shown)


aciccarello:
Great! Thanks for the PR!

raymondfeng:
@aciccarello Would you please publish a new release to `npm`?

Issue #975, opened 2019-02-19
20x slowdown compared to 0.11.x


alexdima:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

## Expected Behavior
Typedoc works as before, and takes reasonable time to execute.

## Actual Behavior
Typedoc is super super slow, going from 2.5s (v0.11.1) to 49s (v0.14.2).

## Steps to reproduce the bug

- save as `monaco.d.ts` - [monaco.d.ts.txt](https://github.com/TypeStrong/typedoc/files/2879629/monaco.d.ts.txt)
- run in PowerShell:
```
 Measure-Command {..\typedoc-0.11.1\node_modules\.bin\typedoc --mode file --out out --includeDeclarations --entryPoint monaco monaco.d.ts}
TotalMilliseconds : 2499.9106

Measure-Command {..\typedoc-0.13.0\node_modules\.bin\typedoc --mode file --out out --includeDeclarations --entryPoint monaco monaco.d.ts}
TotalMilliseconds : 44939.4815

Measure-Command {..\typedoc-0.14.2\node_modules\.bin\typedoc --mode file --out out --includeDeclarations --entryPoint monaco monaco.d.ts}
TotalMilliseconds : 48930.0364
```

## Environment
  - Typedoc version: 0.14.2
  - Node.js version: 0.15.1
  - OS: Windows

aciccarello:
Thanks for the report. I wonder if this is a windows focused issue.

Related to #960

Gerrit0:
I think this must be a windows issue, I haven't run into it. Thanks for the reproduction! Time to pull out `git bisect` and see what went wrong.

Gerrit0:
`git bisect` says d4b2f56ca799cab3f3670c8301b51df34339e420 is the first bad commit, I haven't had time to look into what exactly the issue is.

Gerrit0:
Finally getting back to this, and I still have absolutely no clue why updating to TS 3.x causes this major slowdown when compilation speed remains essentially constant.

The Chrome devtools simply can't handle TypeDoc when trying to parse the above file. I get a hard crash as soon as the profiling is complete, which is making this more difficult than it ought to be.

[...]

PR #1034, opened 2019-05-07
Allow TraverseCallback to bail out early

Gerrit0:
It is used in `getChildByName` when searching for a reflection by name, which led to *always* searching all children, which is especially problematic for any deeply nested structures.

Improves convert time by 25% on the monaco.d.ts file in #975

### DIFF
--- a/src/lib/models/reflections/container.ts
+++ b/src/lib/models/reflections/container.ts
@@ -39,7 +39,7 @@ export class ContainerReflection extends Reflection {
      * @param callback  The callback function that should be applied for each child reflection.
      */
     traverse(callback: TraverseCallback) {
-        for (const child of toArray(this.children).slice()) {
+        for (const child of toArray(this.children)) {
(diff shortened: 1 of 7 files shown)


src/lib/models/reflections/container.ts
```diff
@@ -38,10 +39,10 @@ export class ContainerReflection extends Reflection {
      * @param callback  The callback function that should be applied for each child reflection.
      */
     traverse(callback: TraverseCallback) {
-        if (this.children) {
-            this.children.slice().forEach((child: DeclarationReflection) => {
-                callback(child, TraverseProperty.Children);
-            });
+        for (const child of toArray(this.children).slice()) {
```
aciccarello: Do you need to slice if you are already doing `toArray`?

Gerrit0: Good catch, I assumed `toArray` just returned the input if it was an array, but apparently it copies the array. Will remove.

PR #1039, opened 2019-05-24
Improve output for computed property names

rbuckton:
Currently, the documentation for `[Symbol.iterator](): Iterator<string>` reports the property name as `__@iterator`, and other computed property names are reported as `__computed`. This improves the name generated for computed properties where possible.


aciccarello:
Thanks @rbuckton! Could you add a simple test case? It can be in one of the existing files.

Gerrit0:
It would also be great if there was a way to get the original property name from the compiler API, instead of needing to construct it ourselves, but I'm not sure if that capability exists.

Gerrit0:
Thanks! I looked into getting the original name from the API, but it was more complicated than just computing the symbol name ourselves.

PR #1041, opened 2019-05-29
update dependencies

aciccarello:


.vscode/settings.json
```diff
@@ -1,3 +1,9 @@
 {
-  "typescript.tsdk": "./node_modules/typescript/lib"
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "workbench.colorCustomizations": {
+    "titleBar.activeBackground": "#38ceb5",
+    "titleBar.inactiveBackground": "#38ceb599",
+    "titleBar.activeForeground": "#15202b",
+    "titleBar.inactiveForeground": "#15202b99"
```
Gerrit0: Not entirely sure if this is necessary, but doesn't hurt anything ;)

examples/self/run.bat
```diff
@@ -2,6 +2,6 @@
 set curr_dir=%cd%
 chdir /D "%~dp0"
 
-node ..\..\bin\typedoc --includeDeclarations --externalPattern **/lib/** --mode file --name "TypeDoc Documentation" --out doc\ ..\..\src\
+node ..\..\bin\typedoc --includeDeclarations --externalPattern **/lib/** --mode file --name "TypeDoc Documentation" --out doc\ ..\..\src\lib\
```
Gerrit0: Should this be changed to invoke TypeDoc with the same arguments as the unix side?

package.json
```diff
@@ -31,29 +31,29 @@
...
-    "highlight.js": "^9.13.1",
+    "highlight.js": "^9.15.7",
     "lodash": "^4.17.11",
     "marked": "^0.6.2",
     "minimatch": "^3.0.0",
     "progress": "^2.0.3",
     "shelljs": "^0.8.3",
-    "typedoc-default-themes": "^0.6.0-0",
-    "typescript": "3.4.x"
+    "typedoc-default-themes": "file:../typedoc-default-themes/typedoc-default-themes-0.6.0-0.tgz",
```
Gerrit0: This probably isn't a good idea, if anyone working locally doesn't have the assumed file structure, it will break. If not npm, we should use a GitHub version https://docs.npmjs.com/files/package.json#github-urls

Gerrit0: This is also causing Travis to fail ^

.vscode/settings.json
aciccarello: Yeah, didn't mean to commit that 🤦‍♂

package.json
aciccarello: Yep, once I get a pre-release build of typedoc themes I'll update this

PR #1047, opened 2019-06-08
fix(package): update to typescript 3.5.x

tusharmath:


Gerrit0:
@aciccarello what do you think about dropping node v6 support? Looks like tests are failing again because the old npm version is picking up the wrong version of highlight.js. According to https://nodejs.org/en/about/releases/ v6 isn't even in LTS maintenance anymore.

aciccarello:
@Gerrit0 Yeah, let's drop node 6 now that it's out of LTS.

We might also be able to bump our target version to es2017 which should reduce the amount of work that TypeScript has to do and simplify the output. The only things I see that node 8 doesn't support are things we don't use:

- [template string call site caching](https://node.green/#ES2015-syntax-template-literals-TemplateStrings-call-site-caching)
- [`Array.values()`](https://node.green/#ES2015-built-in-extensions-Array-prototype-methods-Array-prototype-values)
- [shared memory and atomics](https://node.green/#ES2017-features-shared-memory-and-atomics) and a few other regex or getter-setter related features.

Gerrit0:
Thanks!

aciccarello:
We probably should set a minimal node version in our package.json so Node knows what we expect.

tusharmath:
@aciccarello Can you publish these changes? 

Also may be you can try — https://github.com/semantic-release/semantic-release

This will enable u to publish automatically from master.

Issue #1049, opened 2019-06-10
References are marked as external when the path includes a dot


tsvetomir:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

## Expected Behavior
A dot (".") in the project path should have no effect on the generated documentation.

## Actual Behavior
References such as `classes.INameInterface` are treated as external when the project path contains a dot.

## Steps to reproduce the bug
1. Rename `examples/basic` to `examples/basic.dot`
1. Execute the `run` script in the folder
1. Observe that `INameInterface` is no longer an internal reference in `json.json` ([sample](https://github.com/tsvetomir/typedoc-dots-in-path/blob/dots-in-path/basic.dot/json.json#L6511-L6514)):

```json
{
  "type": {
    "type": "reference",
    "name": "classes.INameInterface"
   }
}
```

Normally it would be ([sample](https://github.com/tsvetomir/typedoc-dots-in-path/blob/master/basic/json.json#L6511-L6515)):
```json
{
  "type": {
    "type": "reference",
    "name": "INameInterface",
    "id": 19
   }
}
```

## Environment
  - Typedoc version: 0.15.0-0
  - Node.js version: v10.15.1
  - OS: OSX 10.14.5

Gerrit0:
Huh, well that's unexpected, I'll have to find some time to look at the internal/external distinction to figure out what's going on someday..

Related: #927

tsvetomir:
I've traced the issue to [lib/converter/types/alias.ts](https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/types/alias.ts#L36):

```ts
let symbolName = checker.getFullyQualifiedName(type.symbol).split('.');
if (!symbolName.length) {
    return false;
}
if (symbolName[0].substr(0, 1) === '"') {
    symbolName.shift();
}
```

Here the `symbolName` array would contain extra parts that will not be filtered by the last check above:
```ts
[ '"/private/tmp/typedoc/examples/basic',
  'dot/src/classes"',
  'INameInterface' ]
```

The path seems to be enclosed in quotes so we can remove with a Regex (yay, we have two problems now!):

```ts
const checker = context.checker;
const fqn = checker.getFullyQualifiedName(type.symbol);

let symbolName = fqn.replace(/"[^"]*"\./, '').split('.');
if (!symbolName.length) {
    return false;
}
```

[...]

PR #1052, opened 2019-06-12
fix: references marked as external if path has dots

tsvetomir:
Closes TypeStrong/typedoc#1049


tsvetomir:
The Node v6 build fails with the same error in other PRs. Should I do anything about the error here?

```diff
-<pre><code><span class="hljs-attr">export</span> = SingleExportedClass<span class="hljs-comment">;</span></code></pre>
+<pre><code><span class="hljs-attribute">export</span> = SingleExportedClass<span class="hljs-comment">;</span></code></pre>
```

Gerrit0:
Don't worry about node 6, support has been dropped. If you merge master in, you'll see we are only testing Node 8,10,12 now.

---

I don't particularly like this solution, but looking at the rest of that function, this might be as good as it gets. I'll put this off until the weekend so that I can take a closer look. Thanks for the PR!

tsvetomir:
I've updated the regex to ignore quotes in the path. The only assumption now is that the symbol name doesn't contain quotes.

I'm not familiar with the context here - is there a function that will get us just the path? This should be more precise than the regex.

Gerrit0:
I suspect there probably is a way to get that information from the checker, but I don't know how off the top of my head. That's part of the reason I want to take some time to look at this more closely :)

Gerrit0:
I spent some time looking at this today, and have come to the realization that I don't understand the reasoning behind why some `ReferenceType`s are created by the alias converter, and some are created by the reference converter.

The code hasn't really been touched since Sebastian Lenz wrote it ~4 years ago, so I'd be surprised if anyone still working on the project knows why it was done this way. I'll hopefully have the time to figure it out tomorrow.

tsvetomir:
Okay, let me know if I can be of any help with the investigation.

Gerrit0:
Following smart people working with TS is really helpful sometimes. I have a feeling this will be helpful in determining how this ought to work. I haven't had time to look at it yet... but I'm rather inclined to just merge this PR, it is at worst an improvement over the current behavior.

https://github.com/dsherret/ts-morph/issues/653#issuecomment-506003489

tsvetomir:
Seems like the TS compiler drops information aliases early on — not good. The proposed hack in the issue is not relevant to Typedoc as it requires modifying the type definition.

I'm glad there is such an extensive test suite in place. I'd definitely be uncomfortable with fixing this wasn't the case.

tsvetomir:
I'll appreciate it if this patch makes it to production, as the original issue is really getting in our way. Thanks!

Gerrit0:
Sorry for the delay here, it's been a couple of really busy weeks.

I just tried to pull this down on a Windows machine to play with it locally before merging and ran into an issue since Windows doesn't allow quotes in file paths. I did some research and was unable to find a workaround, so I just removed that and pushed, once the build passes I'll merge.

I can't make releases, but TypeDoc is past due for one. When Anthony has the time there will be a release on the `typedoc` package. I maintain `@gerrit0/typedoc` which is just the TypeDoc master branch (+ theme master branch) if you need these changes sooner.

PR #1053, opened 2019-06-13
Run builds in current Node.js versions

tsvetomir:
Node.js 6 "Maintenance LTS" expired in April 2019.

The [currently supported releases](https://nodejs.org/en/about/releases/) are 8, 10 and 12.


aciccarello:
Great! Thanks for updating this!

PR #1075, opened 2019-07-20
Support negations in directory excludes

stoically:
Currently it's not possible to exclude a directory from the exclude pattern with a glob negation. E.g. if `--exclude '**/!(access)/'` is given, it should match all directories, except `access`. This patch allows doing that by changing the behavior of `add` slightly, if the `file` is a directory:

- Append `/` to allow globs ending with `/` to match directories only
- Don't match the entry-point against the exclude glob, as otherwise a negated glob would prevent further traversing


Gerrit0:
Thanks!

PR #1077, opened 2019-07-22
Updated lodash to 4.17.15 (to avoid prototype pollution)

DayneD89:
Actually only needs to be updated above 4.17.12 to avoid this potential issue, but may as well go to 4.17.15.


Gerrit0:
Thanks!

Issue #1074, opened 2019-07-19
Typedoc runs typedoc.js file with wrong options


Alessar:
* Repro steps:
`npm install typedoc`
Create `"typedoc":"typedoc"` in `run` section in `package.json`
Create file `typedoc.js` with possible options
`npm run typedoc`
* Expected:
Documentation should be generated
* Actual:
`typedoc.js` opened in external text editor
* Thoughts:
Maybe `typedocs.js` is executed without specifying executor like `js` or `node` and in this case `*.js` files interprets as it was setup in system by default. But in this case maybe it is needed to provide additional options for run via `js` or `node` executor for example.

Gerrit0:
This doesn't make much sense to me. TypeDoc doesn't execute the file as an external program. It requires it. I'll try to find some time to look at this either this week or this weekend.

https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/typedoc.ts

aciccarello:
@Alessar It isn't clear what is causing the problem you are encountering since the issue doesn't describe your configuration options. You may want to try configuring typedoc with a typedoc.json file. See our [Installation guide](https://typedoc.org/guides/installation/) for more information.

Since we can't help you with the information provided I'm going to close this but you are welcome to reply with more information if you still need help. We'll reopen if there is a bug.

Alessar:
@aciccarello Unfortunately I cannot understand what you wait from me.
All repro steps I already provided. Options in file is `{ }` - is it possible options?

And what additionally I found:
`typedoc --help` shows:
`--options Specify a js option file that should be loaded. If not specified TypeDoc will look for 'typedoc.js' in the current directory.`
But you and documentation mentioned `typedoc.json` file.
So I don't know to whom I should trust?

So my repro steps contains file `typedoc.JS` according self-documentation from script.
@Gerrit0 Did you try to reproduce discribed reprosteps with such file?

Gerrit0:
This is really weird, but doesn't appear to directly be a TypeDoc problem (though we probably should switch to follow the webpack/rollup/a bunch of other tools that recommend `tool.config.js` for config files, as our current recommendation clearly causes some confusion)

When you do `npm run typedoc` npm:
1. Adds `node_modules/.bin` to your path
1. Starts `cmd.exe` (or whatever you have specified with `script-shell`, see npm docs)
1. Sends the process the value of the `typedoc` key

[...]

PR #1084, opened 2019-07-31
add typedoc.json to 'options' CLI option

aciccarello:
Fixes confusing documentation noted in #1074


Gerrit0:
LGTM. We should probably specify somewhere easily discoverable that JSON is preferred.

aciccarello:
Yeah, I'm considering rewording this option to mostly refer to json.

> Specify a json configuration file path. By default TypeDoc looks for a \'typedoc.json\' or \'typedoc.js\' file in the current directory.

Looking at the website, we should probably note that JSON is preferred in the [`--options` config description](https://typedoc.org/guides/arguments/#options). From there we could point people to the [node module api](https://typedoc.org/guides/installation/#node-module) for those who need a dynamic config.

Gerrit0:
This was superseded by work done to fix options parsing.

PR #1085, opened 2019-08-01
remove content from readme and reference guides on website

aciccarello:
This simplifies the project readme to only explain the basics. This also attempts to make it clear that the readme does not contain exhaustive documentation and that users (and contributors) should refer to the website.

Documentation dropped:
- Less important options
- Gulp/grunt/webpack project references
- Plugin list
- Typedoc 0.2 breaking change note

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,15 @@
 [![NPM Version](https://badge.fury.io/js/typedoc.svg)](https://badge.fury.io/js/typedoc)
 [![Chat on Gitter](https://badges.gitter.im/TypeStrong/typedoc.svg)](https://gitter.im/TypeStrong/typedoc?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
+## Documentation
+Visit our website for more complete documentation and example API documentation:<br>
+[https://typedoc.org](https://typedoc.org).
+
+There you can find an [installation guide](https://typedoc.org/guides/installation/) explaining
+how to use typedoc from the cli, webpack, grunt, or gulp. There are additional guides explaining
+how to extend typedoc using [plugins](https://typedoc.org/guides/plugins/) and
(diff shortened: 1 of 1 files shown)


Gerrit0:
Looks good to me. I like the idea of focusing the readme (and therefore the npm package page) focus on the essentials.

README.md
```diff
@@ -35,48 +44,28 @@ generator. So in order to create a documentation for an entire project you simpl
...
-### Important note
-
-Starting with version 0.2, TypeDoc no longer can predict whether files should be treated as modules
-or whether the project should be compiled into one big namespace. You must specify the `mode` argument
-in order to change the behaviour of TypeDoc.
-
 ### Arguments
 
-For a complete list of the command line arguments run `typedoc --help` or read [here](https://typedoc.org/guides/arguments/).
+For a complete list of the command line arguments run `typedoc --help` or visit [our websit](https://typedoc.org/guides/arguments/).
```
Gerrit0: Typo

README.md
```diff
@@ -35,48 +44,28 @@ generator. So in order to create a documentation for an entire project you simpl
...
 
 * `--out <path/to/documentation/>`<br>
-  Specifies the location the documentation should be written to.
+  Specifies the location the documentation should be written to. Defaults to `./docs`
 * `--mode <file|modules>`<br>
-  Specifies the output mode the project is used to be compiled with.
+  Specifies the output mode the project is used to be compiled with. Required since verion 0.2.
```
Gerrit0: This isn't true. The mode defaults to modules. It might be best practice to specify it as we might want to change the default if the library mode is ever completed, but it isn't required.

https://github.com/TypeStrong/typedoc/blob/cdf3acdd5e56b4fb6683f2ddb8067220e44dde1b/src/lib/converter/nodes/block.ts#L22-L32

PR #1091, opened 2019-08-26
Add support for `` inside of Marked Link Brackets

forivall:


Gerrit0:
Question, does `` `[[ link ]]` `` not work as intended? With the current behavior, the text inside brackets is taken (almost) verbatim to find symbols, and while `` ` `` isn't a very common symbol, it still introduces some complexity that it would be nice to avoid.

forivall:
Hmm, I didn't try that because I would usually expect everything inside of
a code fence to not be interpreted.

socsieng:
> Question, does `` `[[ link ]]` `` not work as intended? With the current behavior, the text inside brackets is taken (almost) verbatim to find symbols, and while `` ` `` isn't a very common symbol, it still introduces some complexity that it would be nice to avoid.

Using the back ticks outside of the links results in the anchor being displayed as code as opposed to the code inside the anchor tag.

My current work-around for this is:

```
[[MyClass.attribute|`MyClass.attribute`]]
```

I would obviously prefer:

```
[[`MyClass.attribute`]]
```

Any chance of this getting merged?

Gerrit0:
I've been putting this off for a bit since I also want to update the docs surrounding linking but haven't gotten around to it. Notably, using `{@linkcode MyClass.attribute}` will also produce a link with code in the anchor tag, but isn't documented anywhere.

Gerrit0:
Released  in 0.16.11 :)

PR #1096, opened 2019-09-18
chore: update typescript dep to 3.6.x

### LINKED ISSUE #1092: Upgrade to Typescript 3.6
- [X] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem
I have a project in TS3.6 that uses the new Generator type, which is not compilable under TS 3.5

## Suggested Solution
Upgrade dependency to TS3.6

simllll:
fixes #1092 
as a side node: needed to add @types/node as an explicit dev dependency, otherwise there is a conflict with IteratorResult (see also https://github.com/microsoft/TypeScript/issues/32333).

All the best,
Thanks
Simon

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -194,9 +194,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "10.5.8",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.5.8.tgz",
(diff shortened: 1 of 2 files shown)


Gerrit0:
Thanks! Good to know about @types/node, that would have had me going around in circles for a while.

Issue #958, opened 2019-01-28
Complex types are documented as 'object' or 'any'


benstevens48:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

## Expected Behavior
Function defined as
```
expandNodes(items: {nodeOrLink: {id: string}, type: string, expandedType: string | null}[]){
...
}
```
should be documented as
```
expandNodes(items: {nodeOrLink: {id: string}, type: string, expandedType: string | null}[]): void
```

## Actual Behavior
Function defined as
```
expandNodes(items: {nodeOrLink: {id: string}, type: string, expandedType: string | null}[]){
...
}
```
is documented as
```
expandNodes(items: object[]): void
```

## Steps to reproduce the bug
Run typedoc with a ts file containing a class containing the above function. (TypeScript strict mode is on, in case that makes a difference.)

## Environment
  - Typedoc version: 0.14.2
  - Node.js version: 10.15
  - OS: Windows 10
  - TypeScript version 3.2.2

If this is due to a typedoc setting or something then I apologize but I couldn't find such a setting documented anywhere.

9oelM:
Perhaps related: #953, #901

benstevens48:
Yes, it does look like it might be the same bug, although `excludeNotExported` doesn't seem to make any difference to me.

ppoulard:
* Similar problem with explicit type definition :
```
type CastContext<C> = {
    cast: typeof UNICAST
    client ?: C
} | {
    cast: typeof MULTICAST
    clients: C[]
} | {
    cast: typeof BROADCAST
}
```
is rendered as 
```
CastContext<C>: object | object | object
```
instead of the object description

* Another example :
```
export const UNICAST: unique symbol = Symbol('UNICAST')
```
is correctly rendered as :
```
UNICAST: keyof symbol = Symbol('UNICAST')
```
however, if I refer to it with `typeof` :
```
export type Capabilities = typeof UNICAST | typeof MULTICAST | typeof BROADCAST;
```
it is rendered as :
```
Capabilities: unique symbol | unique symbol | unique symbol
```
which is somewhat useless

bobheadxi:
Another similar problem I ran into in https://github.com/ubclaunchpad/new/issues/22:

```ts
export const clubConfig: {
  memberCount: number;
  currentTeams: Team[];
  featuredTeams: Team[];
} = {
  /**
[...]

Issue #953, opened 2019-01-19
feat: Add support for documentation on nested optional object/interface


diosney:
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem
I've the following interface definition with documentation comments:

```
export interface IPaymentUrls {
  /** Some description. */
  redirect?: {
    /** Some other description. */
    success: string;

    /** Another description?! */
    cancellation: string;
  }
}
```

Sadly, in the docs it outputs the `redirect` variable as optional, which is fine, but then omits all the other nested properties from the documentation, like if they never existed.

## Suggested Solution
Even if the properties are optional, if they have comments or types, and even if they have an `any` type, they should be shown in the docs.

9oelM:
This issue needs proper reproduction. Perhaps I will have a look some time later when I have some time.

diosney:
@Gerrit0 Sorry but even with that flag turned to `false` they do not show up.

My object is the following:

```
export interface IPaymentUrls {
  /** URLs related to the buyer redirection after some specific events in the Payment Web App flow. */
  redirect?: {
    /** URL that the Merchant can set and that the Payment Web App will use to redirect the buyer when it *successfully completed* the payment. */
    success?: string;

    /** URL that the Merchant can set and that the Payment Web App will use to redirect the buyer when it *cancels* the payment. */
    cancellation: string;

    /** URL that the Merchant can set and that the Payment Web App will use to redirect the buyer when there is an error with the payment. */
    error: string;
  }
}
```

When `redirect` is optional, the documentation shows up like the following (with the frag or without it):

https://imgur.com/L9thSTo

And removing the optional attribute to `redirect` correctly outputs the following:

https://imgur.com/4DRiKu1

(I set `success` to optional to further test them)

If I have another nested object, only if it is not optional gets added to the documentation, also, the same behavior applies.

Gerrit0:
Thanks for the reproduction! This is clearly a bug.

alalonde:
I tested this with and without the `--excludeNotExported` flag.  The nested members do show up when the flag is omitted.

Gerrit0:
This has been fixed + released in v0.16.0

grit96:
I can still reproduce this on typedoc `v0.19.1`.

With optional:

![image](https://user-images.githubusercontent.com/3957608/93344420-9677c280-f829-11ea-9364-0db250295445.png)
[...]

PR #1103, opened 2019-10-01
Fixed skipping of nested object literals with --excludeNotExported

### LINKED ISSUE #901: No reflection on object literals

alalonde:
Should fix #901, #953 and #958.

### DIFF
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -56,6 +56,10 @@ export function createDeclaration(context: Context, node: ts.Declaration, kind: 
     let isExported: boolean;
     if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])) {
         isExported = false; // Don't inherit exported state in modules and namespaces
+    } else if (container.kindOf([ReflectionKind.TypeLiteral]) && context.converter.excludeNotExported) {
+        // If the container for this is a type literal, it never had the
+        // isExported flag set on it.  See https://github.com/TypeStrong/typedoc/issues/953
+        isExported = true;
     } else {
         isExported = container.flags.isExported;
     }
@@ -108,7 +112,7 @@ export function createDeclaration(context: Context, node: ts.Declaration, kind: 
         child.setFlag(ReflectionFlag.Static, isStatic);
         child.setFlag(ReflectionFlag.Private, isPrivate);
         child.setFlag(ReflectionFlag.ConstructorProperty, isConstructorProperty);
-        child.setFlag(ReflectionFlag.Exported,  isExported);
+        child.setFlag(ReflectionFlag.Exported, isExported);
         child = setupDeclaration(context, child, node);
 
         if (child) {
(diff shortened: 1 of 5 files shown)


src/lib/converter/factories/declaration.ts
```diff
@@ -56,6 +56,10 @@ export function createDeclaration(context: Context, node: ts.Declaration, kind:
     let isExported: boolean;
     if (container.kindOf([ReflectionKind.Module, ReflectionKind.ExternalModule])) {
         isExported = false; // Don't inherit exported state in modules and namespaces
+    } else if (container.kindOf([ReflectionKind.TypeLiteral]) && context.converter.excludeNotExported) {
+        // If the container for this is a type literal, it never had the
+        // isExported flag set on it.  See https://github.com/TypeStrong/typedoc/issues/953
+        isExported = true;
```
Gerrit0: This worries me. Is there a reason we can't set the `isExported` flag on containers when creating them instead of this?

alalonde: Yeah, I looked into that.  That does seem like the more "correct" way to go about it. However, after extensive debugging I discovered that this factory function doesn't ever get called for object literals themselves.  Instead, I believe the associated Declaration is created here: https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/types/reference.ts#L127 

Perhaps there is an underlying issue where object literal declarations follow a different code path.  I didn't want to go there, so I just wrote some tests to determine the least invasive fix, which is what you see.

Happy to dive back in if you can provide some additional insights about code structure.  It's a pretty challenging codebase. :)

Gerrit0: Makes sense, thanks. I would like to fix this the right way, but haven't had enough time to dig in and see where that would be yet, so unfortunately I can't provide any guidance here. Maybe next weekend.

Yes, the codebase is rather difficult to follow, we're working on it, but its slow going. :)

Gerrit0:
I spent some time digging in to this and was able to *mostly* fix it by inheriting the `isExported` flag from the parent in `createParameter`, `createSignature`, `TypeLiteralConverter.convert`, and `ReferenceConverter.convert`. I then realized that I recognized a lot of the changes I was making from #801. I suspect that PR also fixes this issue. Since I've been trying to get through it for like a month, I think I'll put my investigation here on hold and just go tackle that, and after getting it merged in see if it has fixed this issue (I'll at least want to add a test like the one you added, unless #801 adds one already)

alalonde:
OK, makes sense.  Thanks for looking into it.  Happy to help with that one where I can.

alalonde:
@Gerrit0 I pulled #801 and added my test to it. The main converter test (with my added `export-literals`) passed, but the same text with the `excludeNotExported` flag set still fails, so I imagine you'll need to add the changes you mentioned above into that branch as well. Or, if you push your changes thus-far to this branch, I can try them out.

Gerrit0:
Good to know, thanks. I'll be sure to pull your test in when I look into this this weekend.

Gerrit0:
Sorry about the delay here! I've confirmed that this is fixed in 0.16.0-8 (`@next` release, full release later today I hope) due to the work with export declarations.

PR #1106, opened 2019-10-04
ignore incremental option in tsconfig

### LINKED ISSUE #1105: Unable to compile projects with incremental mode enabled
I have the following tsconfig

```json
{
  "compilerOptions": {
    "incremental": true,
    "outDir": "dist"
  }
}
```

I am using this configuration file to build my code and also for tsdoc

```
tsc -p tsconfig.json
typedoc --tsconfig tsconfig.json --out docs
```

## Expected Behavior

With typedoc 0.14 and typescript 3.5.3 documentation generates correctly.

## Actual Behavior

With typedoc 0.15 it fails with the following error

```
Using TypeScript 3.5.3 from /Users/serdiuk/workspaces/aws-build/src/AWS-UI-Components/node_modules/typescript/lib
Error: Option '--incremental' can only be specified using tsconfig, emitting to single file or when option `--tsBuildInfoFile` is specified.
```

It happened because newer typedoc upgraded the inner version of typescript so it started picking up the `incremental` option was ignored before.

just-boris:
fixes #1105

### DIFF
--- a/src/lib/utils/options/sources/typescript.ts
+++ b/src/lib/utils/options/sources/typescript.ts
@@ -21,7 +21,7 @@ export class TypeScriptSource extends OptionsComponent {
     static IGNORED: string[] = [
         'out', 'version', 'help',
         'watch', 'declaration', 'declarationDir', 'declarationMap', 'mapRoot',
-        'sourceMap', 'inlineSources', 'removeComments'
+        'sourceMap', 'inlineSources', 'removeComments', 'incremental'
     ];
 
     initialize() {


just-boris:
@aciccarello @Gerrit0 can you please have a look at this PR?

The change is trivial, but unblocks the use of Typedoc in projects with incremental compilation enabled.

Gerrit0:
Thanks! Sorry for the delay here.

yxliang01:
Looking for this patch being release! @Gerrit0 would you consider releasing 0.15.1? Now, I need to forcibly turning off the incremental build..

Gerrit0:
I don't have permissions to make a new release on the main package, we'll have to wait for Anthony on that.

Gerrit0:
Fix has been published in 0.15.1

Issue #493, opened 2017-05-02
Support typedoc.json "extends"


DennisJaamann:
Add support for creating extended typedoc.json files. 

This functionality already exists for:
- TypeScript configs
- TsLint configs

So why not do the same for typedoc when we want for example a company wide configuration for typedocs with the same theme etc preconfigured.

Example:

_**node_modules/@somecompany/somelib/typedoc.json**_
```
{
  "mode": "file",
  "module": "commonjs",
  "theme": "default",
  "out": "docs",
  "target": "ES5",
  "moduleResolution": "node",
  "includeDeclarations":"false",
  "ignoreCompilerErrors": "true",
  "experimentalDecorators": "true",
  "externalPattern":"**/*.d.ts",
  "excludeExternals":"true",
  "emitDecoratorMetadata": "true",
  "preserveConstEnums": "true",
  "stripInternal": "true",
  "suppressExcessPropertyErrors": "true",
  "suppressImplicitAnyIndexErrors": "true"
}
```

**_someprojectusingsomelib/typedoc.json_**
```
{
  "extends": "@somecompany/somelib/typedoc.json"
}
```
**_Currently, there is a way of achieving this but it feels rather cumbersome / not community standard in comparison with tools such as typescript or tslint._**

```
 "scripts": {
    "docs": "npm run typedoc -- --options node_modules/@somecompany/somelib/typedoc.json ./src/",
    "typedoc": "typedoc"
  },
```

aciccarello:
Cool idea. PRs welcome.

PR #1115, opened 2019-10-15
feat: Add extend property

Richienb:
Fixes https://github.com/TypeStrong/typedoc/issues/493

### DIFF
--- a/src/lib/utils/options/options.ts
+++ b/src/lib/utils/options/options.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import * as Util from 'util';
 import * as ts from 'typescript';
+import * as fs from 'fs-extra';
 
 import { Event } from '../events';
 import { Component, AbstractComponent, ChildableComponent } from '../component';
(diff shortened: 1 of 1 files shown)


Gerrit0:
Is this intended to be in draft status?

Richienb:
@Gerrit0 Probably not. I made it and forgot it. I'll finish up the PR soon.

src/lib/utils/options/options.ts
```diff
@@ -81,7 +81,22 @@ export class Options extends ChildableComponent<Application, OptionsComponent> {
...
+                    data = {
+                        ...data,
+                        ...require(name)
+                    };
+                    return true;
+                } catch (e) {
+                    return false;
+                }
+            };
+            if (data.extends.startsWith('@') && !data.extends.split('/')[1]) { tryLoad(`${data.extends}/typedoc-config`); } else { tryLoad(`typedoc-config-${data.extends}`) || tryLoad(data.extends); }
```
Gerrit0: I'm not a huge fan of the many different options here. I think it over complicates the solution. It would be better to just tell the user to specify a file name and error if that file does not exist or fails to load.

We should also probably check for circularity like TypeScript does - http://www.typescriptlang.org/v2/en/tsconfig#extends

Richienb:
@Gerrit0 How should we detect circularity?

Gerrit0:
The simplest way is probably to just keep a `Set<string>` of the resolved paths that have been loaded, and then error if we see the same path twice. This would technically have some false positives (e.g. `extends: ['./a.json', './a.json']`, but that's likely a mistake in the first place.

maneetgoyal:
Hi @Gerrit0, is this PR awaiting any further corrections?

Richienb:
@maneetgoyal The last thing that needs to be implemented is circularity detection.

maneetgoyal:
Hi @Richienb, thanks for picking up this PR. It seems very useful. Any time estimate on this will be of great help.

src/lib/utils/options/options.ts
Richienb: Fixed!

Richienb:
@maneetgoyal I've just finished the last part of the PR. After a bit of testing, it will be ready.

Gerrit0:
Logic looks good to me :+1: I should be able to pull this in this weekend.

Gerrit0:
Ah, of course this was built before the options refactor - I'll pull the changes in to the refactored code.

PR #1118, opened 2019-10-18
refactor: use const instead of let where possible

SebastiaanNijland:
It's really just that (and one docblock typo fix) - `const`s instead of `let`s in a bunch of places.


Gerrit0:
Thanks!

PR #1120, opened 2019-10-22
Ignore tarball, which is the result of running npm pack

PissedCapslock:
When making local changes that you want to test with another project, one option is to create a tarball using `npm pack` and installing that in your other project.

This change will ensure you don't accidentally commit that tarball.


Gerrit0:
Thanks!

PR #1129, opened 2019-11-07
Upgrade to TypeScript 3.7

pmconne:
TypeScript 3.7 introduces [assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) which previous versions of tsc cannot compile.


Gerrit0:
While not technically wrong, it's kind of weird to convert the return type to `void`. I'd like to look into this to see if it would be an easy fix before merging, should have enough time this weekend.

Gerrit0:
Turns out that wasn't too difficult after all, thanks for the PR! I've reached out to Anthony about getting a 0.15.1 release with this in it. Just waiting for Travis to pass to merge.

If you need the changes now, I keep `@gerrit0/typedoc` up to date with master here (whenever bugs are fixed / features added, I don't bother with new releases if nothing user-facing changes), so it will have these changes in an hour or so.

package.json
```diff
@@ -32,29 +32,29 @@
...
+    "handlebars": "^4.5.1",
+    "highlight.js": "^9.16.2",
     "lodash": "^4.17.15",
     "marked": "^0.7.0",
     "minimatch": "^3.0.0",
     "progress": "^2.0.3",
     "shelljs": "^0.8.3",
-    "typedoc-default-themes": "^0.6.0",
-    "typescript": "3.6.x"
+    "@gerrit0/typedoc-default-themes": "^0.6.4",
```
blakeembrey: Can this be reverted once you publish to the main namespace?

Gerrit0: Yes, I kept it here for now so that tests would keep passing. It should be reverted before a new theme version is published.

Gerrit0: It has now been reverted

PR #1140, opened 2019-11-24
Ignore tsBuildInfoFile TypeScript compiler option

### LINKED ISSUE #1105: Unable to compile projects with incremental mode enabled

AdamVig:
Fixes #1105 (again).

### DIFF
--- a/src/lib/utils/options/sources/typescript.ts
+++ b/src/lib/utils/options/sources/typescript.ts
@@ -21,7 +21,8 @@ export class TypeScriptSource extends OptionsComponent {
     static IGNORED: string[] = [
         'out', 'version', 'help', 'emitDeclarationOnly',
         'watch', 'declaration', 'declarationDir', 'declarationMap', 'mapRoot',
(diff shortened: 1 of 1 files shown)


Gerrit0:
Thanks!

PR #1143, opened 2019-11-27
Support exports in definition files

PissedCapslock:
Exports in definition files were not marked as exported.
This patch fixes that.

The patch includes a test as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ yarn-error.log
 /typescript/
 /coverage/
 /dist/
+
+typedoc*.tgz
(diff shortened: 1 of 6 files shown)


Gerrit0:
See also #801 + discussion there. This is more complicated than it seems at first glance.

PissedCapslock:
Ow, my bad. I didn't realize there was already a PR open for this, let alone the number of things I overlooked.

Gerrit0:
Support for this has been merged to master with #1175 :) Thanks!

Issue #1148, opened 2019-12-19
Support documentation for string literal types in unions


socsieng:
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem
Normally I would use an `enum` for this, however, I am documenting an existing JavaScript library as a `.d.ts` file where the actual type doesn't exist in JavaScript.

There doesn't appear to be any support for documenting or applying an `@tag` to individual string literals in a type union.

Example:

```ts
type CallbackType = /** Fired when ready */ 'ready' |
  /** Fired when complete */ 'complete' |
  /** Fired when an error occurs */ 'error' |
  /** @hidden */ 'unknown';
```

The above example has no effect on the generated documentation.

## Suggested Solution
Add support for documenting individual items in a union type including `@tags`.

Gerrit0:
This gets complicated quickly. Right now we don't support documenting any parts of types except for properties of interfaces. If we add support for members of a union, it seems reasonable to also make it possible to document parts of an intersection, properties within those parts... etc. I'm also not really sure how this extra documentation would be rendered.

An alternative method of documenting this could be:
```ts
interface Callbacks {
  /** Fired when ready */
  ready: never;
  /** Fired when complete */
  complete: never;
  /** Fired when an error occurs */
  error: never;
  /** @hidden */
  unknown: never
}

type CallbackType = keyof Callbacks
```

socsieng:
Thanks @Gerrit0.

I ended up going with a slight variation on this:

```ts
type Callbacks = {
  /** Fired when ready */
  ready: never;
  /** Fired when complete */
  complete: never;
  /** Fired when an error occurs */
  error: never;
  /** @hidden */
  unknown: never
}

type CallbackType = keyof Callbacks
```

I preferred the output of the documentation following a `type` declaration instead of an `interface` one. 

In either case, these are the results that I've observed:

- [x] VSCode: Displays `CallbackType` values as union of strings (expected)
- [ ] VSCode: Doesn't honor the `@hidden` tag (unexpected)
- [x] TypeDoc: Honors the `@hidden` tag (expected)
- [ ] TypeDoc: Doesn't automatically link to the `Callbacks` type (unexpected)

Further to the last point, the output of the `CallbackType` documentation reads as follows:

```
T CallbackType: keyof Callbacks
```

But I expected `Callbacks` to be hyperlinked like this:

```
T CallbackType: keyof [[Callbacks]]
[...]

PR #1161, opened 2020-01-12
Resolve TypeOperatorType to allow it to be hyperlinkable

socsieng:
The change add type resolution for `TypeOperatorType` types so that it can be hyperlinkable.

This change depends on https://github.com/TypeStrong/typedoc-default-themes/pull/91 and was raised because of the following issue: https://github.com/TypeStrong/typedoc/issues/1148#issuecomment-568608150

> Not linking to callbacks in `keyof Callbacks` sounds like a bug to me.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2060,9 +2060,8 @@
       }
     },
     "typedoc-default-themes": {
-      "version": "0.7.0-2",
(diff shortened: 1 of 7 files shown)


Gerrit0:
Thanks! Released in 0.16.0

Issue #1142, opened 2019-11-25
@hidden annotation hides all documentation for overloads


fjcaetano:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

## Expected Behavior

When using the `@hidden` annotation in an overload, only that overload's documentation should be hidden.

## Actual Behavior

The documentation for all overloads, including the implementation is hidden.

## Motivation

I have a method similar to `Promise.all` which I have implemented 10 overloads, supporting 10 different generic types. Since Typescript is able to import the documentation from the implementation, there's no need to document each single overload. At the same time, adding incremental overloads pollutes the Typedoc generated documentation, therefore, most of the overloads's documentation could be hidden.

## Steps to reproduce the bug

[...]

PR #1169, opened 2020-01-14
fix: Prevent @hidden signature from removing method declaration

socsieng:
For #1142

The declaration was being treat as hidden when `applyModifiers` was being called. This change determines if the reflection is hidden during resolution instead.

### DIFF
--- a/src/test/converter/comment/specs.json
+++ b/src/test/converter/comment/specs.json
@@ -62,6 +62,78 @@
                 "type": "intrinsic",
                 "name": "string"
               }
+            },
+            {
+              "id": 22,
+              "name": "mixedHidden",
+              "kind": 2048,
+              "kindString": "Method",
+              "flags": {
+                "isExported": true
+              },
+              "signatures": [
+                {
+                  "id": 25,
+                  "name": "mixedHidden",
+                  "kind": 4096,
+                  "kindString": "Call signature",
+                  "flags": {
+                    "isExported": true
+                  },
+                  "comment": {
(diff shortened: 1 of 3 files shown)


src/test/converter/comment/specs.json
```diff
@@ -62,6 +62,78 @@
...
+                    "name": "any"
+                  }
+                }
+              ],
+              "sources": [
+                {
+                  "fileName": "comment.ts",
+                  "line": 70,
+                  "character": 13
+                },
```
socsieng: I couldn't find a way to remove sources from signatures that have been removed.

Gerrit0:
Removing reflections during resolution poses a problem - if a reference is later created pointing to that reflection it will be reported as as a broken reference.

socsieng:
> Removing reflections during resolution poses a problem

This is consistent with existing behavior. Isn't the resolution process the right time to remove reflections?

> if a reference is later created pointing to that reflection it will be reported as as a broken reference

Creating reflections after/during the resolution process seems like the wrong time to be doing this.

Gerrit0:
Sorry, just looking at the diff misled me, I thought they were being removed in the `onDeclaration` method. Yes you are correct, that's the right time to do this. Let me pull it down and play with it locally and it should be good to merge :+1:

src/test/converter/comment/specs.json
Gerrit0: I think this is OK, but will see if I notice an easy way to do it

PR #1171, opened 2020-01-14
Refactor DefaultTheme to be more easily extended

### LINKED ISSUE #1126: Modules with only non-exported constructs should not be exported either

## Expected Behavior
In a file with only `@internal` exports, the module corresponding to that file should not be exported either, and not show up in the "Globals" section.  Assuming `mode="modules"`

## Actual Behavior
The module is exported but empty.  There is a link to it in the "Globals" section.

## Steps to reproduce the bug
With this file `src/internal.ts`:
```
/**
 * @hidden
 */
export class Internal {
  ...
}
```

`typedoc --mode="modules" --excludeNotExported src/`

## Environment
  - Typedoc version: 0.15.0
  - Node.js version: 10.8
  - OS: Mac OS 10.12

alalonde:
I pulled the internal functions out of `getNavigation` into a separate `NavigationBuilder` class so that a custom theme doesn't have to re-implement all of them. 

I also changed the default behavior of `build` to no longer add links to empty modules.  This fixes #1126

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   ],
   "scripts": {
     "pretest": "node scripts/copy_test_files.js",
-    "test": "nyc --reporter=html --reporter=text-summary mocha --timeout=10000 dist/test/**/*.test.js",
+    "test": "nyc --reporter=html --reporter=text-summary mocha --timeout=10000 'dist/test/**/*.test.js'",
     "prerebuild_specs": "npm run pretest",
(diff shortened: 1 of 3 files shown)


Gerrit0:
This is awesome :+1: I don't have the time to go through and look at it today, but should be able to get to it soon. Thanks!

Gerrit0:
Well, the navigation code still scares me, but this is still an improvement. Thanks!

Issue #1177, opened 2020-01-16
Should removing/hiding a reference remove type references as well?


socsieng:
For the given example:

```ts
/**
 * @hidden
 */
class HiddenClass {
}

class AvailableClass {
}

type Classes = HiddenClass | AvailableClass;

class OtherClass {
  hidden: HiddenClass;
  available: AvailableClass;
}
```

The following is effectively the output:

```ts
class AvailableClass {
}

type Classes = HiddenClass | AvailableClass;

class OtherClass {
  hidden: HiddenClass;
  available: AvailableClass;
}
```

The hidden class is removed but there are remaining references to it.

What is the desired behavior in the example above? Should it be as-is or should it remove or remap references? (e.g. replace the references with `unknown`)

Example:

```ts
class AvailableClass {
}

// remove type if in a list
type Classes = AvailableClass;
// alternatively, remap to unknown
// type Classes = unknown | AvailableClass;

class OtherClass {
  hidden: unknown;
  available: AvailableClass;
}
```

I would personally like to see something like:

- Remove the type reference if used in a list,
- Remap the reference to `unknown` if the resulting list is empty,
- Remap the reference to `unknown` if directly referenced

If you are happy with this, I can have a go at implementing it. What are your thoughts?

Gerrit0:
This is tricky. Mapping types not included in the documentation to `unknown` poses a problem because `unknown` is the top type, `unknown | AvailableClass === unknown`. It also implies a less-strict contract when calling a function.

```ts
type Internal = { __internal: true }
declare function withInternal(arg: Internal): void
```

If this is mapped to 
```ts
declare function withInternal(arg: unknown): void
```

Then a user would expect to be able to pass whatever they like into the function... Which isn't right and will likely cause runtime errors.

---

Since TypeScript supports `@internal` for removing a symbol from the declaration file, I decided to take a look at what `tsc` does. It looks like it currently follows the same strategy that TypeDoc does, keeping the name when referenced in other locations but dropping the declaration.

```typescript
/** @internal */
type Internal = 1
type External = 2
export type Foo = Internal | External
```
When compiled with `--stripInternal` will result in

```typescript
declare type External = 2
export declare type Foo = Internal | External
```

socsieng:
> This is tricky. Mapping types not included in the documentation to `unknown` poses a problem because `unknown` is the top type, `unknown | AvailableClass === unknown`. It also implies a less-strict contract when calling a function.

[...]

PR #1184, opened 2020-01-19
Support for library mode

### LINKED ISSUE #109: Rename "modules" to "namespaces"
Starting with TypeScript 1.5 we have namespaces that will/should replace modules, see [here](https://github.com/Microsoft/TypeScript/pull/2923). I would like to rename the documentation output accordingly as I always struggled with the distinction between "modules" and "external modules".

| Old name | New name |
| --- | --- |
| Module | Namespace |
| External module | Module |

This would break existing urls and might interfere with other things but I think it's worth the trouble. Any opinions on this?

### LINKED ISSUE #639: Document only a library's exported files from a given starting point
Currently, `typedoc` exports all internally exported entities which are not part of the exports in the given starting point. I've been trying to find a way to get past that using available options but looks like this feature is not available.

For example, if starting point is `src/index.ts`, then only exports from this file should be documented and not all entities. Restricting files in config file or marking every internal utility function as `@hidden` makes things unmaintainable.

If I have missed anything, please let me know.

### LINKED ISSUE #880: Suggestion: Allow customization of TypeScript version

Gerrit0:
This turned out to be far simpler to implement than I expected, despite the crazy large diff (mostly spec updates)

## :package: Changes

1. Added `--mode library`
1. Moved TypeScript to a peer dependency
1. Renamed modules to namespaces and external modules to modules

## :rocket: Try it out

The changes in this PR have been published as `typedoc@next`. You can install it and try it out with:

```
npm install typedoc@next typescript
npx typedoc --mode library --out docs src/index.ts
```

### 📚 Library Mode

This is the flagship feature for <del>0.17</del><ins>0.18</ins>. With this mode, TypeDoc will take each of the specified input files and generate documentation for that input file as one module. It will resolve any re-exports of that module and treat them as if they were declared in that file. Note that if you have multiple entry points that export the same symbol from a helper file, this will result in multiple copies of that symbol in your documentation

Note that input files are still expanded before conversion. If you specify a directory, every file in that directory and subdirectories will be treated as a library entry point.

### 🔓 TypeScript Peer Dependency (shipped in 0.17)

TypeDoc has kept TypeScript as a dependency for a long time. The cited reason for this is that the project uses internal compiler APIs that any new version of TypeScript could break.

However, in practice, the internal APIs that TypeDoc uses have remained stable since I first started working on the project over a year ago. In addition to this, many TypeDoc users hacked around TypeDoc's strict version limitations by removing TypeScript from TypeDoc's node_modules folder.

TypeDoc still uses internal APIs. I expect it will break with *some* future version of TypeScript, but giving users the ability to use a newer TS version is a good thing. For now, the peer dependency will require at least the minor version of TypeScript kept in TypeDoc's dev dependencies.

### ➡️ Renamed External Modules (shipped in 0.17)

"External Modules" have been renamed to be properly called "Modules", and the legacy "Modules" have been renamed to "Namespaces". This is a major breaking change for plugins and themes with custom CSS, but has been overdue for a long time.

### ❓ Unresolved Questions

- For library mode, if there is only one entry point, should the module be removed and all reflections be present in the root reflection? (This would look like `--mode file` but with exports from the entry point only)
- The way library mode works makes it behave as if `--excludeNotExported` was set, does this make sense? I think it does... Should library mode remove the only exported checkbox in the generated docs?

Closes #109
Closes #639 
Closes #880


src/lib/utils/options/sources/typedoc.ts
```diff
@@ -26,8 +26,9 @@ export function addTypeDocOptions(options: Options) {
         help: "Specifies the output mode the project is used to be compiled with: 'file' or 'modules'",
```
socsieng: Update help text?

socsieng: Should also probably update readme file as well:

> `--mode <file|modules>`

Gerrit0: Good catch, thanks!

Gerrit0: Good catch, thanks!

Gerrit0: Good catch, thanks!

Gerrit0: Website docs will also be updated by https://github.com/TypeStrong/typedoc-site/pull/26

danmarshall:
Hi @Gerrit0 , I very much appreciate this work 🥇 I have been wanting this capability for a long time.
I am trying it out but I do not get any output at all, other than blank pages for global and my index module. 

My minimal sample repo is https://github.com/danmarshall/test-typedoc-library and I have generated the docs here: https://danmarshall.github.io/test-typedoc-library/
Thanks!

rolftimmermans:
Love this, thanks for your work!

Is re-exporting already supposed to work? I'm trying this:

**internal.ts**
```
/**
 * Re-exported function.
 **/
export function foo() {

}
```

**index.ts**
```
export {foo} from "./internal"

/**
 * Exported function.
 **/
export function bar() {

}
```

When running `yarn run typedoc --mode library index.ts` I only see the `bar()` function documented.

Gerrit0:
Oops! Sorry about that @danmarshall and @rolftimmermans. Both of those should have worked. I forgot to resolve symbol aliases before trying to document the node.

Will be fixed in 0.17.0-2 shortly.

rolftimmermans:
Perfect! The re-exports now work very well.

The only downside I see at the moment is that the entry points are a bit hard to find. For libraries with a single (main) entry point it would be good to just treat all symbols as "globals", so that they are visible immediately in the table of contents when opening the main/readme page.

See the screenshot: all API documentation is "hidden" behind the unfortunately named `"index"` link.

Screenshot 2020-01-21 at 13 38 02

Gerrit0:
I guess there's a couple votes for not creating a module if only one entry point is provided. I think it should be relatively simple to do.

Weakky:
Hey, thanks for your amazing work 🙏. I tried running the library mode against https://github.com/prisma-labs/graphql-santa, but ran into an issue. We are exposing ts types from some underlying libraries.

Is there any way to get the type information from them? I tried adding `--includeDeclarations`, but that doesn't seem to work
Instead, most the the type informations contain just the name of the types, but are not "clickable"/discoverable

danmarshall:
@Weakky - I have a similar scenario. It does make me wonder how far the crawling should go. All the way to the base TypeScript libs?

TheNeuralBit:
This is awesome, thanks so much @Gerrit0. I tried this 0.17.0-2 out on the arrow docs and it looks great!

FYI @trxcllnt and @lmeyerov I pushed up a preview. gh-pages doesn't seem to work right but you can browse it with githack.com: https://raw.githack.com/TheNeuralBit/arrow/typedoc-library/js/doc/index.html

Weakky:
@danmarshall Yeah, that makes sense to me, unless specified otherwise with a flag. I see two ways to filter out deps:

- Either by depth
- Or by a list of include (or exclude)

Geo1088:
Thanks so much for working on this! I'm running into an error when trying to run this on my project, any help? I'm working with [this repository](https://github.com/Geo1088/yuuko).

<details>
<summary>Console session output</summary>

```
george ~/yuuko $ yarn run typedoc --version
yarn run v1.21.1
$ /home/george/yuuko/node_modules/.bin/typedoc --version

TypeDoc 0.17.0-2
Using TypeScript 3.7.5 from /home/george/yuuko/node_modules/typescript/lib
Done in 0.68s.
george ~/yuuko $ yarn run typedoc --mode library --json test.json src/Yuuko.ts 
yarn run v1.21.1
$ /home/george/yuuko/node_modules/.bin/typedoc --mode library --json test.json src/Yuuko.ts

Using TypeScript 3.7.5 from /home/george/yuuko/node_modules/typescript/lib
/home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:84
            for (const declaration of resolved.declarations) {
                                               ^

TypeError: resolved.declarations is not iterable
    at BlockConverter.convertVisibleDeclarations (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:84:48)
    at /home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:50:26
    at Context.withScope (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/context.js:101:9)
    at /home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:49:25
    at Context.withSourceFile (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/context.js:85:9)
    at BlockConverter.convertSourceFile (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:39:17)
    at BlockConverter.convert (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/nodes/block.js:30:18)
    at Converter.convertNode (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/converter.js:116:53)
    at Converter.compile (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/converter.js:159:26)
    at Converter.convert (/home/george/yuuko/node_modules/typedoc/dist/lib/converter/converter.js:90:29)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

</details>

Regarding the questions posed in the OP:

> For library mode, if there is only one entry point, should the module be removed and all reflections be present in the root reflection?

This would be my use case, +1.

> The way library mode works makes it behave as if `--excludeNotExported` was set, does this make sense? I think it does...

I think this makes sense, but I wouldn't be against requiring both flags for this behavior either in case someone has a valid use case for not inferring `--excludeNotExported`.

> Should library mode remove the only exported checkbox in the generated docs?

This behavior should probably be separated from library mode and be implemented as part of `--excludeNotExported` if it's not already present. If it's decided that library mode will imply that flag as well, then the checkbox should go away. (Is the checkbox not already hidden from the docs when using that flag? If not, this is probably a question for a separate PR entirely.)

suchipi:
Hello! I am really excited for this feature. I pulled `next` and tried it out, but found that some of the stuff in my package was not documented; specifically, functions that were inside objects, like follows:

```ts
import Foo from "./Foo";
import Bar from "./Bar";

export default { Foo, Bar };
```

I have created a repro repo for this here: https://github.com/suchipi/typedoc-object-repro

aciccarello:
This looks great! Glad to see library mode added. What do you think about making this a 1.0 release?

Gerrit0:
> I guess there's a couple votes for not creating a module if only one entry point is provided. I think it should be relatively simple to do.

This will happen.

> I tried running the library mode against https://github.com/prisma-labs/graphql-santa, but ran into an issue.... 

How are you running TypeDoc? I tried cloning the project and running with `npx typedoc --out tddocs --mode library .\src\index.ts` (also with `--ignoreCompilerErrors`).

Is `src/index.ts` your entry point? If so, it seems to export a single `const app`. You aren't actually exporting any of the types contained in that variable so TypeDoc doesn't include them. (That said, it looks like TypeDoc bugs out a bit, I'd expect the docs to show `app: App`, not `app: { a whole bunch of stuff }`)

If you can't do `import { Type } from 'lib'`, I don't think library mode should include it...

> Thanks so much for working on this! I'm running into an error when trying to run this on my project, any help? I'm working with this repository on WSL.

Thanks for the reproduction! TypeScript lies :/ apparently `import * as Eris` isn't a declaration, so that symbol doesn't have any declarations and TypeDoc gets confused...  Looking into it.

> If it's decided that library mode will imply that flag as well

Right now, it does. Only the exports from the entry points are documented.

> functions that were inside objects, like follows:

This is... horrible. Why are you default exporting an object with names? Just directly export the names... In my opinion default export is a bad idea in general. You lose the ability to associate a meaningful name with the export. I'm mostly in agreement with [Basarat](https://basarat.gitbook.io/typescript/main-1/defaultisbad), and in complete agreement with [Nicholas](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/)

I'm not really sure how TypeDoc should handle this. It seems wrong to create a top level declaration for both `Foo` and `Bar` as if they were exported... It should at least create a reflection with `{ Foo: Foo, Bar: Bar }`... even if `Foo` and `Bar` aren't linked.

Right now we don't have a good way to describe "reflections that are indirectly exported". To solve the specific issue this needs to be solved first.

> This looks great! Glad to see library mode added. What do you think about making this a 1.0 release?

Not yet! A 1.0 release implies that TypeDoc's API is stable, and that is still far from the truth. I think the options API can now be considered stable along with the JSON serialization API (but that still needs docs, I've started a [development](https://typedoc.org/guides/development/) page on the TypeDoc site but it isn't even close to being finished)

Gerrit0:
> Thanks for the reproduction! TypeScript lies :/ apparently import * as Eris isn't a declaration, so that symbol doesn't have any declarations and TypeDoc gets confused... Looking into it.

This wasn't quite right. Eris (or relevant `@types` package) isn't installed, so TypeScript is unable to find a declaration file for it... We have three options when this happens:

1. Log some kind of error and skip the declaration
2. Pretend it is just an unknown value and document it as `unknown`
3. Don't care because this causes a TypeScript error. I'm leaning toward this option. Telling TypeDoc to ignore compiler errors should only be done if docs are built after a successful build.

Interestingly, I can't build your project. TypeScript gives 12 errors when trying to compile. One of these is:

```
src/Yuuko.ts:8:23 - error TS2307: Cannot find module 'eris'.

8 import * as Eris from 'eris';
                        ~~~~~~
```

This wasn't reported before TypeDoc crashes because apparently we convert the project *before* checking for errors... This is bad. Fixing.

As soon as you install Eris, this problem goes away completely! The project builds and TypeDoc no longer crashes. The downside (?) is that Eris is now included in your documentation. This is unavoidable currently since `@hidden` would have to be placed within the Eris declaration file.

Gerrit0:
Inspired by the report above caused by Eris, I spent some time considering how library mode should work with TypeDoc's options for including / excluding content... open to feedback here on what we should do.

---

Most of TypeDoc's options for including / excluding content from the documentation are heavily influenced by the existing file and module modes. This doesn't translate well to the library mode where nodes to document are discovered by asking the TypeScript compiler for what is exported.

- `includeDeclarations` - Currently ignored. Exported symbols from declaration files are always included. Most of the time, I think libraries would want this turned on so that their documentation accurately describes the exported api surface... Is this accurate?

- `exclude` - Currently mostly ignored. It will result in `expandInputFiles` ignoring a file when being expanded (if a directory is passed) but it won't prevent declarations in said files from being included in the documentation, which is inconsistent with the other modes... I think this should be consistent..?

- `externalPattern` - Currently mostly ignored. It could cause whole modules being removed from the documentation, but will have no affect on declarations from those files being exported (`export * as F from 'f'` will be removed but not `export { f } from 'f'`)... this is inconsistent in a bad way.

- `excludeNotExported` - Currently effectively forced to be true. I think this is correct.

- `excludePrivate` / `excludeProtected` - Work as intended.

suchipi:
> This is... horrible. Why are you default exporting an object with names?

@Gerrit0 That was a contrived example for the sake of repro; my actual code is more like:
```ts
// Physics.ts
function PhysicsBody() {}
function PhysicsEngine() {}
function PhysicsConstraint() {}

const Physics = {
  Body: PhysicsBody,
  Engine: PhysicsEngine,
  Constraint: PhysicsConstraint,
};
```
```ts
// Components.ts

import Mouse from "./Mouse";
import Physics from "./Physics";

export { Mouse, Physics }
```
```ts
// index.ts

export * from "./Components";
```

The idea is that for the end-user, related functions are stored in a namespace together.

Gerrit0:
Is there a reason you don't do something like this?

```ts
// physics.ts
export function Body() {}
export function Engine() {}

// components.ts
import * as Physics from './physics'
import * as Mouse from './mouse'
export { Physics, Mouse }
// or if you have TS 3.8
export * as Physics from './physics'

// index.ts
export * from './components.ts'
```

This produces an identical API surface (well, ignoring internal paths), but because you don't create the object yourself, TypeDoc is able to recognize that everything is a re-export and handle it correctly in library mode.

suchipi:
When I do it that way, TypeDoc in library mode lists the contents of Physics as a separate module, and doesn't indicate that Physics it an exported property from the "index" module:

Screen Shot 2020-01-24 at 2 38 59 AM
Screen Shot 2020-01-24 at 2 39 03 AM

(this third screenshot is just demonstrating the lack of a Physics variable)
Screen Shot 2020-01-24 at 2 39 28 AM

Geo1088:
> Eris (or relevant @types package) isn't installed, so TypeScript is unable to find a declaration file for it...

Aha, that explains it. My project has Eris as a peer dependency, so I have to have it installed with `--no-save` to build my project. I was testing on a fresh clone and forgot to do that. Sorry for making you debug that, all on me lol.

Gerrit0:
> Aha, that explains it. My project has Eris as a peer dependency, so I have to have it installed with --no-save to build my project

This is what dev dependencies are for! `npm i --save-dev eris`

> TypeDoc in library mode lists the contents of Physics as a separate module, and doesn't indicate that Physics it an exported property from the "index" module:

This is a bug, TypeDoc should be exporting it under the name it is exported from the root module.

paztis:
It seems there's a lot of remaining problems in the library mode
I just create 3 simple flies:

export-a.ts
```
export const moduleA = 'a';
export const moduleB = 'b';
export const moduleC = 'c';

const useless = 'useless';
export default useless;
```

export-b.ts
```
import * as startImport from './export-a';

interface ITest {
    attr1?: string;
    attr2?: number;
    attr3?: ITest2;
}

interface ITest2 {
    attr1: string;
}

export {startImport};
export const normalName = 'a';

const b: ITest = {};
export {b as renamedB};

const internalNameDefault = 'default';
export default internalNameDefault;
```

index.ts
```
export * from './export-b'
export {default} from './export-b'
```

In the documentation I imagine to have in index.ts file

- **starImport** namespace => instead I've got **export-a** module that is totally different:
  - name is bad: the public name is **starImport** not **export-a**
  - contnet is also bad because `import * as X from ...` create a namespace with **only the named exports**, when your module pointer contains all the exports, **default** included.
- **normalName**: this one is good a the export name is not transforming the variable name
- **renamedB** => instead I've got the **b** name that is not the public name of the library
- **default** => instead I've got **internalNameDefault**

The problem of the **starImport** namespace is that you're pointing on the original module instead of creating a new namespace / module

All the naming problems came from **block.ts**:
```
    private convertVisibleDeclarations(context: Context, node: ts.SourceFile) {
        const moduleSymbol = context.getSymbolAtLocation(node);
        if (!moduleSymbol) {
            this.application.logger.warn(`File ${node.fileName} is not a module and cannot be converted in library mode`);
            return;
        }

        for (const symbol of context.checker.getExportsOfModule(moduleSymbol)) {
            const resolved = context.resolveAliasedSymbol(symbol);
            for (const declaration of resolved.declarations) {
                this.owner.convertNode(context, declaration);
            }
        }
    }
```
**resolveAliasedSymbol(symbol)** returns you the internal symbol with the internal name, when **symbol** well have the external name.

You might preserve intermediates reflexions for this.
Reflexion also have a **name** and an **originalName** field. originalName well point to the internal name but the other might point to the public one.

It is not possible to correctly document a library until start imports, default exports and renemed exports are not correctly managed.
This is standard cases that are really important to treat.

NB: even in module or files mode, these case are never been well treated.

paztis:
I just try this fix:
```
private convertVisibleDeclarations(context: Context, node: ts.SourceFile) {
        const moduleSymbol = context.getSymbolAtLocation(node);
        if (!moduleSymbol) {
            this.application.logger.warn(`File ${node.fileName} is not a module and cannot be converted in library mode`);
            return;
        }

        for (const symbol of context.checker.getExportsOfModule(moduleSymbol)) {
            const resolved = context.resolveAliasedSymbol(symbol);


            if (symbol.getName() !== resolved.getName()) {
                const reflection = createReferenceReflection(context, symbol, resolved);
                    for (const declaration of resolved.declarations) {
                        this.owner.convertNode(context, declaration);
                    }
                });
            } else {
                for (const declaration of resolved.declarations) {
                    this.owner.convertNode(context, declaration);
                }
            }
        }
    }
```

and it resolved all the namings problems (but not the namespace and the interface definition)
image


I will create a PR for this in your branch

paztis:
PR created: #1196

paztis:
this modification in types/reference.ts allow us to see the interface documentation
```
    convertNode(context: Context, node: ts.TypeReferenceNode, type: ts.TypeReference): Type | undefined {
        if (!type.symbol) {
            return new IntrinsicType('Object');
        } else if (type.symbol.flags & (ts.SymbolFlags.TypeLiteral | ts.SymbolFlags.ObjectLiteral)) {
            return this.convertLiteral(context, type.symbol, node);
        }

        const result = createReferenceType(context, type.symbol);
        if (result) {
            if (node.typeArguments) {
                result.typeArguments = this.owner.convertTypes(context, node.typeArguments);
            }
            for (const declaration of type.symbol.declarations) {
                this.owner.convertNode(context, declaration);
            }
        }

        return result;
    }
```

But it is nested inside the parent export definition and hard to read
image


I can manage the **context.withScope** to move up until the root module context (project), but it merge all the project interfaces.
If I use the scope of the 1st module, it mix it inside the exported vars.

Perhaps it is at the templating level we need to mark the exported vars or separate them inside another graphic group to avoid this mix ?

Gerrit0:
> starImport, renamedB, internalNameDefault

I've found a fix for this, I think the right solution is just to rename the converted node (with a few other tweaks because other stuff breaks when doing this)

> it is well defined with the type ITest but there no way to know what is the definition of ITest. This interface is not exported, but it doesn't prevent to document it.

I'm not sure I agree with this. In modules or file mode, TypeDoc would document non-exported members. I've never liked this as it makes it very easy for the generated documentation to lie about what is available to consumers.

Library mode tries to generate documentation that accurately describes what a user would actually be able to use in a module.

If the documentation says there is an interface `ITest` in the package, I expect to be able to do `import { ITest } from 'package'` and use it. If I can't do this then the documentation arguably shouldn't include it.

---

I spent a couple hours tonight working on these issues, and have mostly fixed them locally, but I've run into a couple more problems that I don't have time to work out right now.

paztis:
> I've found a fix for this, I think the right solution is just to rename the converted node (with a few other tweaks because other stuff breaks when doing this)`

This generates other issues: if you do a défaut export of a class, in all your class dedicated pages, the name of this class will be default instead of its original name.
Default extends default is understandable.

> I'm not sure I agree with this. In modules or file mode, TypeDoc would document non-exported members. I've never liked this as it makes it very easy for the generated documentation to lie about what is available to consumers.

This type is exported in typescript. You can do `typeof` over a function return type to get back the type.

Here we're not just talking about a esthetic doc, we talk about a correct doc, that really explain all the library to the users

Gerrit0:
> This generates other issues: if you do a défaut export of a class, in all your class dedicated pages, the name of this class will be default instead of its original name.
Default extends default is understandable.

Fair point, a reference also needs to be created because nodes might declare circular dependencies, which this doesn't handle correctly right now...

> This type is exported in typescript. You can do typeof over a function return type to get back the type.

This isn't the same as being exported, and doesn't always work.

```ts
type Nominal<T> = { __nominal__: never } & T
export const Arr: Nominal<string[]> = ["str"] as any
```

Here, while I can do `type TArr = typeof Arr`, I cannot get a reference to `Nominal` so that I can provide my own `T`.

Libraries which choose not to export interfaces that they expect external users to conform to are asking for trouble...  In fact, there was some discussion (#1177) asking for TypeDoc to take a more extreme approach by changing `ITest` to `unknown`.

The correct thing to do for libraries is to actually export the types that are indirectly exported. Sure, I can do `type ShouldHaveExportedThis = ReturnType<Parameters<import('lib').Func>[3]>`, but that shouldn't be required.

paztis:
Understand for the type export.
So for the export naming we need a intermediate solution between your and mine ?

Gerrit0:
Yes, for export naming we need to find some solution that correctly handles exports under a different name, and doesn't completely break when some symbol is exported twice. #1197 is a very strange artifact that I am guessing is related to this problem.

It would be nice to be able to tell if a given symbol is exported under its original name, and if it is to generate a reference to the original name, and otherwise create a new declaration, but I'm not sure if this solves all the problems... I should have time to play around with it more this weekend.

Gerrit0:
After some more thought... This is the best I've come up with. TypeDoc doesn't have a way to describe a symbol that exists but is only exported under a different name, and I'm loath to add one. It doesn't fit with the rest of the architecture, and I'm not ready to rebuild the models yet (though I really want to! The inheritance doesn't make sense in many cases)

If a name is exported more than once, one export should be chosen as canonical and all other references to that symbol should be created as references. The canonical name should be taken from the first point that applies:

1. The symbol is directly exported from an entry point under its local name (defined in that file and exported there)
1. The symbol is re-exported from an entry point under its local name. If it is re-exported under the same name from multiple entry points, arbitrarily pick one (I'm not sure there is value in defining some process here, maybe earlier entry points should be preferred?)
1. The symbol is directly exported from an entry point under a different name (`export { a as b }`).
1. The symbol is re-exported from an entry point under a different name (`export { a as Canonical } from 'f'` or `export { default as Canonical } from 'f'`
1. The symbol is exported as `default` - only in this case will there be no documentation available with a different name for the symbol than `default`.

Thoughts?

paztis:
If at the end the final exported names are correct in each file it is ok.
It will reduce the number of references.

Le dim. 2 févr. 2020 à 20:58, Gerrit Birkeland <notifications@github.com> a
écrit :

LeviticusMB:
This does not seem to work if I specify `--theme minimal`? *(TypeDoc 0.17.0-3)*

paztis:
I'd continue to work on my PR #1196 and it seems correct now

My fork module is also available in @paztis/typedoc over npm (https://www.npmjs.com/package/@paztis/typedoc).

You can try if you want.
On my side it covers all my needs now. It may covers also yours.

Can you check at it ?

ggoodman:
@Gerrit0 I'm a huge fan of the naming adjustments for namespaces but have another nuance that might be interesting to consider:

- Exported symbols that are both namespaces and interfaces

Suppose you want to define an RPC Transport interface, you might have an interface like:

```typescript
/**
 * The interface for rpc-compatible transports.
 */
interface Transport {
    /**
     * Dispose of this transport and free any allocated resources.
     */
    dispose(): void;
    /**
     * Register a RPC message listener with the transport
     *
     * @param handler A handler function to be called with each RPC message received from a peer
     */
    onMessage(handler: (msg: unknown[]) => void): {
        dispose(): void;
    };
    /**
     * Send an RPC message to the peer over this transport
     *
     * @param msg The array-encoded message that should be sent to the peer over the transport
     * @param transfer An optional array of objects that should be marked as transferrable when the transport supports it
     */
    sendMessage(msg: unknown[], transfer?: unknown[]): void;
}
```

However, your library already provides ready-made factories for common `Transport`s, so you throw those in a namespace sharing the name with the interface:

```typescript
declare namespace Transport {
    function fromNodeMessagePort(port: import("worker_threads").MessagePort): Transport;
    function fromDomWorker(worker: Worker): Transport;
    function fromDomMessagePort(port: MessagePort): Transport;
}
```

Now, `typedoc` will consider `Transport` a namespace that happens to have both functions and methods and will totally drop the fact that there is an exported `interface`:

image

I hope that this is something that you might consider addressing as well! ❤️

MoeLava:
Hello! I appreciate this PR as it could be more suitable for my use case. However, I have a request for the navigation sidebar.
Let's say I have this test library at https://github.com/MoeLava/typedoc-lib-test to generate docs for. Currently, it looks like
image
but would it be possible to make it look something like 
image
instead?

Gerrit0:
Quick note on progress here - There are still several issues with the implementation here that I want to address. Some have been mostly addressed by @paztis's work in #1196, but not all the edge cases that I want to cover before a full release. I hope to have some time in a couple weeks to spend a few days on this and get it wrapped up. This PR is being rescheduled for 0.18.

Some of the changes in this PR will be cherry picked to master for a 0.17 release that upgrades TS to 3.8. These are:
- `--disableSources`
- `--gitRemote`
- TS as a peer dependency
- Check compilation errors before converting
- Support for ``[[`link`]]`` style links

Library mode is coming eventually... I just need to find a few days to concentrate on it.

---

@ggoodman TypeDoc's declaration merging has some issues, I think that's a bit out of scope for what I'm trying to do here, but please feel free to open an issue to discuss it. I think it would be a good change (The only merging that makes sense to me is merging multiple interfaces or namespaces of the same name or a namespace + class of the same name. Merging interfaces and namespaces seems like a disaster waiting to happen)

@MoeLava that's a tricky change. How would we tell how far to expand containers? In this case the expanded docs are clearly more useful, but what if each of those namespace had 50 members instead of 2? I don't really want to define some arbitrary nesting level that effects the output... (apparently there is some hard coded limit with 9 members in the TOC already that I didn't know about until recently)

I think a better solution might be a collapsible TOC. There could be an option `--tocExpandThreshold` to govern how many children should cause the parent to be collapsed or open by default... that said, I think that is also probably a separate issue.

PS: You might want to try changing your docs command to `typedoc --out docs src/index.ts --mode library`

rubengmurray:
First off, this has been a great thread to read through - thanks for all the hard work on this project.

I'm just getting started with TypeDoc and trying out library mode but I'm running into an issue with the display of imported types. The interface declaration is missing, but the annotations on those types _are_ being documented. The types are imported from a private TypeScript package shared between internal projects.

I'm not sure if it exactly relates to any of the issues discussed here thus far but here goes:

Here's a structural example of the project I want documenting and the dependency:

Project I want documenting
```
// (myFile.ts)
import { TypeGroupA } from 'sharedTypes'
export { TypeGroupA }

export class MyClass {
  public myFunc(): TypeGroupA.Details { 
    // do something that returns TypeGroupA.Details
  }
}

```

In 'sharedTypes' dependency
```
// (index.ts)
export { TypeGroupA, TypeGroupB, TypeGroupC } from './typegroup1';
export { TypeGroupD } from './typegroup2';

// (typegroup1.ts)
import * as TypeGroupA from './typegroupa';
import * as TypeGroupB from './typegroupb';
import * as TypeGroupC from './typegroupc';

export {
  TypeGroupA,
  TypeGroupB,
  TypeGroupC
};

// (typegroupa.ts)
/**
* Interface description
**/
export interface Details {
  name: string;
  iq: number;
}

```

Now, strangely in the generated TSDoc I see the comment of the interface, but not the interface itself...

Screenshot 2020-04-02 at 13 35 12

Seems like if it gets that far it should be displaying the interface types... a bug something I'm doing?

rubengmurray:
I should add that navigating to `file:///.../docs/interfaces/` does display all of the expected interfaces as separate html files. Indeed they are even displayed with a hierarchy when clicked (if they are extensions of other interfaces), but no details for any.

marcofugaro:
Hey, great PR! I've tried this mode with `tsdoc@next` and can give you some feedback on it!

Here is my `typedoc.json`
```json
{
  "inputFiles": ["./src/index.ts"],
  "out": "docs/",
  "mode": "library",
  "theme": "minimal",
  "excludeNotExported": true,
  "excludeExternals": true
}

```

And my structure is like this:

```ts
// index.ts

export * from './collision/ObjectCollisionMatrix'
export * from './collision/AABB'
export * from './collision/ArrayCollisionMatrix'
export * from './collision/Broadphase'
export * from './collision/GridBroadphase'
// ...
```

```ts
// ./collision/AABB.ts

/**
 * Axis aligned bounding box class.
 * @class AABB
 * @constructor
 * @param {Object} [options]
 * ...
 */
export class AABB {
  // ...
}
```

And so on.

However in the index, I still see the modules:
image

And I would like to see just the classes, something like this
image

If you want to see the full implementation see https://github.com/react-spring/cannon-es/pull/14

Hope this helps wit this PR, thanks for your work @Gerrit0!

Gerrit0:
@rubengmurray Definitely a bug! I'll add this to my test cases.
@marcofugaro That's really weird, I'll take a look at that too.

Unfortunately I haven't had much time to work on this since the last update, there are major issues with the existing implementation that seem to get worse the more I look at them... I'm going to try another approach which will hopefully fix them.

Berkmann18:
@Gerrit0 Do you nee help in getting this through?
It would amazing to have this on `typedoc`.

Gerrit0:
It's coming along, unfortunately I'm really busy finishing up last semester of college right now so I don't have much time to work on it. The implementation in this PR is also not going to work - it gives okay results, but has major issues that can't be fixed with the current converter architecture. I've made a list in https://github.com/TypeStrong/typedoc/issues/1266 on what needs to be done for this to be released.

Unfortunately I haven't spent much time separating out discrete tasks, so I can't really point you toward something that can be done independently.

WandersonAlves:
Hey, thanks for this PR :smile: 

I'm using `typedoc@^0.17.0-3` and `typescript@^3.9.2`, and no problems so far.

I run `typedoc` throught `"docs": "typedoc --out docs --mode library src/index.ts"` script in `package.json`

And it's working as expected
image

christopherthielen:
> The implementation in this PR is also not going to work - it gives okay results, but has major issues that can't be fixed with the current converter architecture.

The output from library mode is so much better for my project that I intend to switch my docgen to use it.  I understand that this implementation is dead, but it just makes so much more sense!  My plugins were doing a bunch of stuff to manage modules and internal/external but it's mostly not necessary any more.  I will probably be implementing a much simpler plugin that shows only whitelisted symbols in the `groups` and `toc` on the index page.

paztis:
Have you try my version (https://www.npmjs.com/package/@paztis/typedoc) ? I've start from the feat/639 branch to improve the library mode. I've also an open PR over the project to merge it.

kossnocorp:
Hey everyone, I just wanted to thank everyone involved, you're doing fantastic and much-needed work!

I'm migrating date-fns to TypeScript and TypeDoc (I use it to generate JSON and then load to the database) is a lifesaver!

telamonian:
> > The implementation in this PR is also not going to work - it gives okay results, but has major issues that can't be fixed with the current converter architecture.
> 
> The output from library mode is so much better for my project that I intend to switch my docgen to use it. I understand that this implementation is dead, but it just makes so much more sense!

:+1: to this. I help maintain [a huge monorepo project](https://github.com/jupyterlab/jupyterlab/) (with [85](https://github.com/jupyterlab/jupyterlab/tree/master/packages) packages, if you can believe it), and our existing [typedoc-based API docs](https://jupyterlab.github.io/jupyterlab/) have been somewhere in between bad and useless. This is largely due to the fact that we've essentially been stuck with generating the docs for each package independently.

I've been playing around with library mode, and it has been a breath of fresh air. Now we get a unified API doc build, one in which the monorepo packages actually cross-reference each other (with working hyperlinks and everything!), and in which I don't have to handroll a janky static `index.html` to serve as a TOC.

So I think we too will be switching our docgen over to library mode as implemented in this PR (we already have a PR started for this at jupyterlab/jupyterlab#8320). @Gerrit0 If there's anything I can do to help pitch in and push library mode out the door (eg testing, "good first issue" PRs, etc), please let me know

Gerrit0:
That's awesome to hear @telamonian! I didn't know Jupyterlab used TypeDoc.

The main issues with this implementation are:
1. Models are built up piecemeal. TypeDoc historically has built docs by visiting nodes on the AST provided by TS. This causes several problems. First, there are quite a few hacked together methods to determine what is exported, from where, and to try to emulate TypeScript's declaration merging. This sometimes results in documentation which is just... wrong, which is worse than no documentation to begin with.
2. Partially due to 1, TypeDoc cannot even begin to deal with complex import/exports. If you have circularity in imports, we will completely miss exported members.

The main issue with my attempt at solving this... is that I am really bad at saying no to scope creep. I'm slowly getting better here...

To fix the first issue we really need to revisit the models themselves. After discussion in #1244, I've come to the conclusion that TypeDoc shouldn't do it at all, instead producing 2+ declarations (with the exception of namespace + namespace, interface + interface, and interface + class)

Today, basically all the models are instances of [`DeclarationReflection`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/models/reflections/declaration.ts#L36), which contains a ton of extra properties that don't make sense on all members (For example, why can an enum member have call signatures)... giving each logical part of a program its own model will force fixing the first issue, so I decided to do it...

Once each part of a program had its own model, the converters had to be updated to use them. This *mostly* wasn't too bad. I've rebuilt almost all of the converters.

I also had to update the serializers for JSON output to work with the new models, but that wasn't too bad... I increased coupling a bit by moving serialization partially back into the models, but it's easier to follow what's going on there now.

What I hadn't foreseen and what's become the biggest time sink here is that the current themes are incredibly tied to the old models. They simply don't work with the new models. I tried updating the handlebars themes to work with this, but soon realized that missing out on the type info on what is available for each model was going to make this take ages... so I decided to start from scratch, building the rendered HTML with Preact (Didn't use React because Preact is smaller, and I'm not using anything fancy anyways)... did I say I'm bad with scope creep? Yea... I also decided to pull #1292 in and have spent a weekend and a half working on getting that working with support for specifying both light and dark themes (The [Rust docs](https://doc.rust-lang.org/std/vec/struct.Vec.html) made me jealous)...

TLDR, this attempt basically turned into a full rewrite of TypeDoc, which like all rewrites doesn't support everything the original did. My current todo list (partially collected from scattered todo comments):

- [ ] Support JS users using `module.exports`, global files [source](https://github.com/TypeStrong/typedoc/blob/library-mode/src/lib/converter/converter.ts#L219) (smallish, shouldn't be too hard if familiar with the TS api)
- [ ] Finish Object reflections in the converter (medium)
- [ ] Fix converter for destructured exports. (smallish)
- [ ] Take a look at existing plugins to see what events they use, if the existing emitted events meet those needs, or if we need other events. (medium)
- [ ] Add indexed signatures to models that can have them. I forgot they existed when rebuilding the models... (medium)
- [ ] Get highlighting working. This means let the user pick a dark and light theme to highlight code with, and generate HTML which uses CSS variables to set the color, so that it switches correctly. (I thought this was going to be easy, turned out to be hard, what I'm currently working on)
- [ ] Correctly handle multiple entry points, producing re-exports when appropriate to avoid duplicating documentation. (I have the process this should follow written out somewhere... have thought about it a fair bit)
- [ ] Fix `{@link}` and `[[link]]` (medium)
- [ ] Merge the class and interface converters. There's a fair bit of duplicate code (fairly small)
- [ ] Figure out how `@event` should work now that the models accurately represent the source and we can't just set the `kind` to `ReflectionKind.Event`
- [ ] Fix the source reference plugin to get links relative to the root dir (small)
- [ ] Add a log level flag so that verbose messages aren't printed by default (small)
- [ ] Finish rebuilding the theme (much larger than I expected... and I am quite out of practice with this frontend work...)
    - [ ] Support hideGenerator flag (small)
    - [ ] Make overriding templates easy for users (???)
    - [ ] Write docs on how to do this (small-medium)
    - [ ] Talk with the people that built the markdown plugin about how to integrate. (not sure how much this will take, I don't like breaking people... but I have REALLY done it this time)
    - [ ] Make `@category` group reflections on the page by category (medium)
    - [ ] Display a page index for documented items on that page (smallish, dependent on previous)
    - [ ] Support for the `media://file` and `[[include:file]]` (small)
- [ ] Review the existing tests. They're currently broken on the branch. I think the serialization tests can mostly just be rebuilt, but want to look more closely at them.
- [ ] Create a rendered test with the new theme + minimal theme. The rendered tests might be something that can be simplified with Cypress or some such tool (???)
- [ ] Write "how to build a TypeDoc plugin" (medium, going to use the sources plugin as an example most likely)
- [ ] Update the [development page](https://typedoc.org/guides/development/) on the website (medium)
- [ ] Update plugins that were broken when rebuilding converters
    - [ ] Decorator plugin (medium)
    - [ ] Implements plugin (medium)
    - [ ] Deep comment plugin (small, probably merge with the existing comment plugin)
- [ ] Test on a bunch of different projects, particularly large ones. (Apache Arrow, Jupyterlab) (??? depends on what issues are found)
- [ ] Upgrade fs-extra, bump minimum node version to node 10 (tiny)
- [ ] Rename `inputFiles` to `entryPoints` (breaking everything anyways, might as well make names make sense, small)
- [ ] Review existing options to see which ones still make sense.

Stretch goals I want but am not going to delay a release over:

- [ ] Highlight headings according to the color scheme chosen by users, not just code snippets (medium? Tricky. I spent a few hours on it, failed)
- [ ] Switch to ESLint. TSLint is unsupported. (smallish? I haven't done it before)
- [ ] Add a reflection for "union enums" https://github.com/microsoft/tsdoc/issues/164

---

So.. yea. Lots to do yet. If you (or anyone!) wants to help, please feel free to take any of the items on the above list and send a PR towards the [library-mode](https://github.com/TypeStrong/typedoc/tree/library-mode) branch. It definitely isn't ready as a beta, or even an alpha on npm... but I think the hardest parts have been done.

Gerrit0:
After spending an hour or so going through issues today I realized this probably can't be the next minor release. I'm going to keep working on it, but there are other bugs that I don't want to force people to wait until this is done to have fixed.... in fact, I'm going to call this a goal for 1.0. With the architecture changes that have been made, I think a stable API is much more possible.

brandones:
Library mode was working magnificently for me. Is there an npm package or version that I can freeze to until this is merged? Typedoc isn't really useful for my libraries without it, but is *extremely* useful with it.

Gerrit0:
Pinning the version to 0.17.0-3 will let you keep using the existing implementation.

telamonian:
@Gerrit0 Just to let you know, jupyterlab has pinned to 0.17.0-3 and is still using library mode for our API docs. And [they are spectacular](https://jupyterlab.github.io/jupyterlab/) (or at least much better than they were previously). Thanks for all the great work on this so far!

codycuellar:
I just want to +1 this effort. It's the only mode that makes any sense to me. There's still a few minor issues - but overall it documents things in a way that is clear for consumers of our library, AND for devs of the library, and doesn't really require any extra plugins and doc tags to get it working as expected.

Gerrit0:
Thanks for the feedback! I'm glad it has been useful.

I just realized today that typedoc.org has been pointing people to this PR, which won't be merged. There is a beta up for TypeDoc 0.20 which acts as described in this PR (the `mode` option was removed, so I can't really call it library mode) See #1364 for updates on that.

Issue #1136, opened 2019-11-15
Only generate legend for items that are displayed on the page


Raiondesu:
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem
Currently, the legend at the bottom is cluttered with ALL possible types of exports including some particularly rare ones. This can often lead to a feeling of being lost in some readers, when first looking at the bottom of the page.

For example, if viewing a class that only contains some methods and properties, it's pretty useless for the reader to be reminded about icons for enums, enumeration members or variables...

## Suggested Solution
It should be possible to only render legend for the items that are currently present on the page.

This would greatly benefit documentation simplicity of the small packages with short API-s.

johansten:
Is this really working the way it is supposed to? I'm getting a legend that contains far fewer entries in them, but they have virtually no overlap with the types I'm expecting it to display:

<img width="1039" alt="Screen Shot 2020-04-06 at 16 49 15" src="https://user-images.githubusercontent.com/151766/78540362-a32dce80-7826-11ea-89f3-632b4ce9627c.png">
<img width="44" alt="Screen Shot 2020-04-06 at 16 48 37" src="https://user-images.githubusercontent.com/151766/78540373-a628bf00-7826-11ea-89c5-8d728550a51a.png">

Gerrit0:
That definitely looks like a bug, looks like we are missing `Enum`, `Interface`, `Class`, `Function` and... whatever that other purple dot is.

There was a bug in a previous version caused by the rename from "External Modules" to "Modules", are you on the latest version?

johansten:
I'm using 0.17.4

Gerrit0:
Could you open a new issue with a reproduction? Seems to be working correctly based on a quick test with the symbols I noted are missing that I just did.

mcmartincerny:
I have the same problem. Enum and other types are missing. I am using mode "file".
@johansten : have you found a solution to the problem?

kamtschatka:
I am having the same issue with the enums. Interestingly the "Enumeration" shows up in my real project and the "Enumeration member" is missing, but in my reproducer both of them are missing.
Is there a way to include all of them again, so we at least have all of them shown properly?

PR #1187, opened 2020-01-22
feat: Only generate legend for items that are displayed on the page

socsieng:
For #1136

Will require another change to typedoc-default-themes (https://github.com/TypeStrong/typedoc-default-themes/pull/94)


Gerrit0:
Marking this as for 0.17, thanks!

Gerrit0:
Thanks! Will be released later today.

PR #1194, opened 2020-01-26
fix: Plugin resolution for relative paths

### LINKED ISSUE #1188: Passing --plugin with a relative path is unintuitive
## Expected Behavior

When running TypeDoc with `--plugin ./test.js` I expect `./test.js` relative to the cwd to be loaded.

## Actual Behavior

TypeDoc fails to load the file since it tries to load `./test.js` relative to an internal path.

## Environment
  - Typedoc version: 0.16/0.17

socsieng:
Resolve plugin modules relative to the current path instead of the internal path.

Fixes #1188

### DIFF
--- a/src/lib/utils/plugins.ts
+++ b/src/lib/utils/plugins.ts
@@ -20,7 +20,7 @@ export class PluginHost extends AbstractComponent<Application> {
      */
     load(): boolean {
         const logger = this.application.logger;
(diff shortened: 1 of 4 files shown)


Gerrit0:
Looks good :) This needs a minor release since it has reasonable potential to break existing setups. Thanks!

Gerrit0:
Thanks! Will be released later today.

PR #1196, opened 2020-01-30
Feat/639

paztis:
correct the default exports and star imported namespaces in library mode


src/lib/converter/nodes/block.ts
```diff
@@ -112,8 +112,18 @@ export class BlockConverter extends ConverterNodeComponent<ts.SourceFile|ts.Bloc
...
+
+            const declarationsProcess = () => {
+                for (const declaration of resolved.declarations) {
+                    this.owner.convertNode(context, declaration);
+                }
+            };
+
+            if (symbol.getName() !== resolved.getName()) {
+                const reflection = createReferenceReflection(context, symbol, resolved);
+                context.withScope(reflection, declarationsProcess);
```
Gerrit0: This is kind of weird, conceptually a reference shouldn't have children, it should point to something that has children... Not necessarily wrong, but unexpected. It think might be better to just set the name of the converted reflection.

src/lib/converter/nodes/block.ts
```diff
@@ -114,7 +114,7 @@ export class BlockConverter extends ConverterNodeComponent<ts.SourceFile|ts.Bloc
             const resolved = context.resolveAliasedSymbol(symbol);
 
             // export declaration is always unique: no need of loop
-            const declaration = resolved?.declarations?.[0];
+            const declaration = resolved.declarations?.[0];
```
Gerrit0: This isn't correct. This is valid TS:

```ts
export interface A {
    a: number
}

export interface A {
    b: number
}
```

src/lib/converter/context.ts
```diff
@@ -398,6 +398,28 @@ export class Context {
 
         return typeParameters;
     }
+
+    /**
+     * Typscript getFullyQualifiedName method don't always return unique string
```
Gerrit0: Yes, that's correct. If a declaration is not exported, it won't contain the source file path... It seems wrong to me to add it since it might conflict with a FQN created by the TS version. That said, I haven't read through everything yet so there might be a good reason for it that I'm not seeing...

src/lib/converter/nodes/block.ts
Gerrit0: You can do the same with namespaces

src/lib/converter/plugins/DynamicModulePlugin.ts
```diff
@@ -72,8 +78,28 @@ export class DynamicModulePlugin extends ConverterComponent {
...
+            this.hiddenPathPart.forEach((pathPart) => {
+                if (name.endsWith(pathPart)) {
+                    name = name.substr(0, name.length - pathPart.length);
+                }
+            });
+
+            reflection.name = this.basePath.trim(name);
+
+            // in case of mono-repo (node_modules outside the project), trunk the module name after the node_modules folder
+            const nodeModulesRegexp = new RegExp('^(.*)node_modules/');
```
Gerrit0: Better style to use the literal regex syntax:
```suggestion
            const nodeModulesRegexp = /^(.*)node_modules\//;
```

Gerrit0: I'm kind of worried about this though, it seems like it could lead to paths appearing to be contained within the project that are external.

src/lib/converter/plugins/DynamicModulePlugin.ts
```diff
@@ -23,6 +23,9 @@ export class DynamicModulePlugin extends ConverterComponent {
      */
     private reflections!: Reflection[];
 
+    private hiddenExtension = ['.ts', '.d'];
```
Gerrit0: Did .ts extensions show up somewhere? I've never seen them...

Gerrit0: Ah, I see this is because it now overwrites the name... not sure about this. Will have to play with it.

src/lib/converter/nodes/block.ts
paztis: I know we can declare 2 times the same interface, that result in a merges one, but can we really export 2 times over the same name something ?
In classic case (const) the second export erase the 1st one.

It means here we need to declare a unique merged interface ?

src/lib/converter/context.ts
paztis: In 2 different files I was having the same named export.
And typescript wasn't providing me a unique identifier (no file path in front)
The result was bad reference link in the documentation.

Here in case the fqn doesn't contain a doc, I prefix it with the file path and it has resolved my problem

src/lib/converter/plugins/DynamicModulePlugin.ts
paztis: Ok will fix

src/lib/converter/plugins/DynamicModulePlugin.ts
paztis: The goal is to avoid the remaining.d at the end of the file.
It supports the .ts, the .d.ts and classic cases (.js, .jsx, .tsx,...) Thanks to the extension.
But extension alone is not working for .d.ts

src/lib/converter/nodes/block.ts
Gerrit0: Doh, that's what I get for looking at code with antihistamine in my system. You are correct.

paztis:
Any news for htis branch ? Is is still in progress on your side ?

paztis:
new updates here: I resolve all the problematic problems we were having

paztis:
Hi 

what is the status of this PR ?
Do you want to merge it ? or not ?
It start to be a long time now...

paztis:
Hello

Any news ?

Gerrit0:
Hey, sorry for the delay in response here.

The library mode implementation used in this PR + the feat/639 branch was essentially abandoned. Trying to make library mode fit with the other modes caused problems with maintainability (what was valid in different places changed depending on the mode, which options made sense changed...) I originally thought the way to fix this was going to be rewriting the converters, and started doing that on the library-mode branch... it took too long to realize that would practically never be finished so I took another look at the feat/639 branch + this branch, this resulted in the work on the library2 branch, which I'll be putting a beta up for this weekend.

I don't plan on merging either of these branches, but it has been very useful in working on the library2 branch, so thanks for that!

paztis:
I will take a look at library2 branch to see.
One thing is I actively use this PR at work and I hope I can reach the same result.
In case not, is it ok for you if I officially fork you repo to release my version ?

Gerrit0:
If you run into issues with the library2 branch - by all means open an issue!

The beauty of open source is that forking is possible - so long as you abide by the license I have absolutely no problems with forks, though of course I'd like to make the implementation in library2 so good that nobody feels the need to fork to use another implementation :)

Gerrit0:
0.20 has been released, thank you for your work here :)

PR #1206, opened 2020-02-14
Moved @types/minimatch dependency to devDepencencies

alexandercerutti:
I moved @types/minimatch to devDependencies because, right now any project that uses typedoc, installs along with the main package, also minimatch's types. They shouldn't be installed if not for development.

Or is there a specific reason for which these types have been inserted in dependencies? If any, can you please explain it?

😊

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "node": ">= 6.0.0"
   },
   "dependencies": {
-    "@types/minimatch": "3.0.3",
     "fs-extra": "^8.1.0",
     "handlebars": "^4.7.2",
     "highlight.js": "^9.17.1",
@@ -43,6 +42,7 @@
     "typescript": "3.7.x"
   },
   "devDependencies": {
+    "@types/minimatch": "3.0.3",
     "@types/fs-extra": "^8.0.1",
     "@types/lodash": "^4.14.149",
     "@types/marked": "^0.7.2",
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,6 @@ export { Application } from './lib/application';
 export { CliApplication } from './lib/cli';
 
 export { EventDispatcher, Event } from './lib/utils/events';
-export { createMinimatch } from './lib/utils/paths';
 export { resetReflectionID } from './lib/models/reflections/abstract';
 export { normalizePath } from './lib/utils/fs';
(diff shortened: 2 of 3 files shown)


Gerrit0:
Minimatch is included because we re-export some of its types - see #1035.

I'd like to move it to devDependencies, but haven't had the time to look at what is exported and if it can be removed from the public API. I think if `createMinimatch` is no longer exported from the root file, it could be removed. If you want to go ahead and do that and make sure the tests still pass, I think `@types/minimatch` can be moved to devDependencies :)

https://github.com/TypeStrong/typedoc/blob/c75df1c6a0d19aca71946a732ed834b43e69a761/src/index.ts#L5

alexandercerutti:
I'll investigate more and let you know! 😊 Thank you for the explanation.

alexandercerutti:
I think that removing this, would mean a possible breaking change for plugins... what do you think about @Gerrit0?

Gerrit0:
Yes, it would. I am okay with that, it will have to wait for the 0.17 release, but that already has several breaking changes slated for it, one more to include this fix is well worth it.

alexandercerutti:
I've run the tests before and after the edits.
I've removed, as you showed to me, the export of  `createMinimatch`  from index.ts and changed tests path.

There's only one test that fails due to a timeout error (over 10000 seconds), so it should definitely be related to the changes.
I'm gonna report it here below, probably you already know that:

```
 1) TypeDoc
       Application
         Honors the exclude option even if a module is imported:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/c/Users/alexa/repos/typedoc/dist/test/typedoc.test.js)
      at processImmediate (internal/timers.js:445:21)
```

Gerrit0:
The failing test you have locally is due to the tests just running slowly on your machine for some reason (slow disk access probably). They don't fail on mine or on the CI.

alexandercerutti:
That's weird, but I don't know. ¯\_(ツ)_/¯
Anyway, I corrected the message, I meant the problem wasn't related to the PR changes.

Thank you!

Gerrit0:
Thanks! Will be in 0.17 later today.

PR #1205, opened 2020-02-14
Logger: also count the warning messages

PissedCapslock:
If you want to make your build fail on error messages, you currently can do
```
const result = app.generateDocs(project, OUTPUT_DIR);
if (!result || app.logger.hasErrors() ) {
  console.log("Errors were logged during the generation of the documentation. Check the log for more information");
  process.exit(1);
}
```

However, warning messages typically indicate a problem as well. If you want to make your build fail on those as well, you need a custom logger to keep track of the warning messages yourself.

This PR makes this easier by counting the warning messages in a similar fashion as the error messages.

PR #1210, opened 2020-02-18
Logger extensions now also count the warnings

PissedCapslock:
I cannot believe I messed up that previous PR (https://github.com/TypeStrong/typedoc/pull/1205), but the extensions of the `Logger` class did not count the warnings.
Solved this by removing the error counting code from the child classes, and called the super method instead.
This ensures that both the error count and the warning count increases.


Gerrit0:
My bad as well - I assumed child classes were well behaved, which is not an assumption that can be made with TypeDoc yet unfortunately!

Thanks for the fix!

PissedCapslock:
That was my assumption as well. Great was my surprise after upgrading that the logger said zero warnings while the log indicated otherwise.
Thanks for merging (again).

Gerrit0:
Released  in 0.16.11 :)

Issue #1180, opened 2020-01-16
Generate documentation from JSON output


jonsage:
# JSON file could also be used as the input for Typedoc

Typedoc can output a JSON file containing all the reflection data instead of creating the HTML (or other format) documentation. It would good if this JSON file could also be used as the input for Typedoc and the HTML (or other format) documentation generated from that JSON.

## Problem
I work with a documentation portal, and I do not have access to the TypeScript code itself, just the exported reflection data (JSON). I'd like to process this into documentation and manage the templates/theming within my side of things.

## Suggested Solution
Allow Typedoc JSON format refelection data to be used as input for the Typedoc tool.

Gerrit0:
This would be a neat bit of functionality to support. I think we should eventually be able to do this... it's a ways off though. Some information in the reflections isn't reflected in the JSON output. (Mainly cross references, I think all the type information is present)

raygesualdo:
> Some information in the reflections isn't reflected in the JSON output.

Can you expound on this @Gerrit0? I was planning on using the JSON output to build a simple markdown doc for some internal Typescript libraries we use.

Gerrit0:
I think with 0.20 we're a lot closer to being able to do this in TypeDoc itself. Missing items I can think of initially:

- `files` / `directories` on `ProjectReflection` - I want to remove these anyways. We build them up but don't really use it for anything, and AFAIK no plugin does either.
- `url`, `anchor`, `hasOwnDocument`, `cssClasses` on `Reflection` - these are all inappropriately placed anyways, they shouldn't live on the reflections.

... and that's all I can think of right now, which means this is actually a lot closer to being possible than I thought it was. It might be one of the goals of 0.22 since 0.21 is going to clean up both of these.

This shouldn't stop you from using the JSON output today. This request is for TypeDoc being able to be re-run on JSON produced from a prior run. It's certainly possible to generate docs with a different tool from the json already.

raygesualdo:
Thanks for the clarification!

jankuca:
This would be a killer feature.

sancarn:
TypeDoc sites look amazing too, the way I see it, this feature could see TypeDoc site used in other typed languages (as long as they can convert the AST type data to JSON)

divya-gariga:
Now using Typedoc, can we generate documentation in HTML format using JSON output?

Gerrit0:
[...]

Issue #910, opened 2018-12-01
Json deserialization


hlehmann:
In order to easily use the json file for documentation rendering, could it be possible to "deserialize" and get a ProjectReflection object ?

Gerrit0:
Neat idea! Currently not possible, and this would probably require a fair bit of work since serialization is spread out over so many different classes right now. If you want a `ProjectReflection` for now it's probably best to use the api to generate one.

```ts
const td = require('typedoc');
const app = new td.Application({
  // options
});
const inputFiles = app.expandInputFiles('src');
const projectReflection = app.convert(inputFiles);
```

mootari:
From what I (believe to) recall this is practically impossible at the moment because:
- not all the required information is retained in the JSON output,
- there's a lot of cross-references between reflections that cannot be restored,
- there are references to TypeScript type instances that cannot be skipped.

A better way might be to stop accessing reflection instances in the theme and instead rely on a reduced dataset (like from toObject()).

Gerrit0:
I recently spent quite a bit of time in the serialization code and agree with mootari. There isn't any nice way to resolve this. If you need a `ProjectReflection` when rendering a theme, the best way to do that is to use the code I provided above.

Gerrit0:
So... this seems to be a recurring theme with TypeDoc for me. I think something is impossible, then 6 months or a year later I come back and realize - this isn't that difficult. There are a few things that need to happen first (and some of these are being done for library mode), but this is definitely doable, and I really like the idea.

- [ ] Semi-revert the serialization work done to separate models and serailizers. We still need the serialization module so that plugins can add their own data to the serialized models, but much of it can be restored to the models so that the code lives together and is easier to reason about.
- [ ] Create a new method on `Application`:
    ```ts
    interface Application {
        restoreFromJson(json: object): ProjectReflection
    }
    ````
- [ ] Create the `Deserializer` module.
    ```ts
    interface Deserailizer {
        addReflectionDeserailizer(kind: ReflectionKind, Extract<SomeReflection, { kind }>)
        addReflectionDeserailizer(kind: TypeKind, Extract<SomeType, { kind }>)
        deserializeReflection(json: object): SomeReflection
        deserializeType(json: object): SomeType
    }
    ```

The deserializer should work in much the same way as the serializer used in library mode (not pushed to GH yet, still have another 200 or so type errors to fix...)

Gerrit0:
#1180 is the more common user facing side of this feature, so I'm going to track it over there.

PR #1216, opened 2020-02-22
feat: extract & expose `toString` for stringifying types

### LINKED ISSUE #1160: [RFC] Consider extracting `toString` functionality to standalone module
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem

It's not currently possible to use typedoc's type to string rendering alone &mdash; you get the CLI, the file/filesystem handling, markdown rendering & syntax highlighting, etc. Extracting string rendering to a separate module/package would enable:

* using typedoc in a more modular fashion, i.e. to parse and render independent of one another
* use of parts of typedoc's capabilities in more areas, such as browser-like environments or other site generators
* rendering of type classes as is possible now, but also of serialized objects like the ones in typedoc JSON outputs

My driving use case is to render types to documentation in [Docusaurus v2](https://github.com/facebook/docusaurus), which uses React & [MDX](https://github.com/mdx-js/mdx). While there is an existing [plugin](https://github.com/tom-grey/typedoc-plugin-markdown/blob/42c49e9b6ae47602f2d1f6c43a102da4fb5a6682/THEMES.md#docusaurus--docusaurus2) for outputting markdown and targeting Docusaurus v2, it doesn't allow using the full power of the React & MDX ecosystems (it just outputs static markdown).

## Suggested Solution

I'd be willing to work on this if accepted. I recommend extracting the `toString` methods that currently live in each of typedoc's various type classes into a separate module that can then be used internally as well as exposed to allow the above mentioned features. The new API would act on objects provided as arguments which should allow it to support both the existing functionality and also operate on data such as from typedoc's serialized JSON output.

I imagine at a high level that the API could look something like this:

```ts
// specialized functions for each type
export const stringifyArray = (node: ArrayType): string => {
    // implementation
};

export const stringifyUnion = (node: UnionType): string => {
    // implementation
};

const stringifiers = {
    array: stringifyArray,
    union: stringifyUnion
  // ...
};

// generic type-to-string renderer that defers to
// the appropriate specialized functions above
export const stringifyType = (node: Type | SerializedType): string => {
    return stringifiers[node.type](node);
};
```

Then the existing classes could likely be refactored to use these by implementing `toString` in the base class instead of overriding it in every subclass:

```ts
// src/lib/models/types/abstract.ts

import { stringifyType } from '../path/to/stringify-module'

export abstract class Type {
	// ...

    /**
     * Return a string representation of this type.
     */
    toString(): string {
        return stringifyType(this);
    }
}
```

haltcase:
Closes #1160

This essentially moves all the type model `toString` implementations into a separate module and exposes them for users, as spec'd in the above issue. `toString` is implemented on the base model now, so that method is still kept around for backward compatibility currently.

Currently it should have parity, and has a start for stringifying reflections in a similar manner (just call signature for now).

I've also added tests, which don't cover everything, but these weren't actually tested before at all. Progress!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 export { Application } from './lib/application';
 export { CliApplication } from './lib/cli';
-export * from './lib/stringifiers';
+export { stringifyType } from './lib/stringifiers';
 
 export { EventDispatcher, Event } from './lib/utils/events';
 export { createMinimatch } from './lib/utils/paths';
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,6 @@
 export { Application } from './lib/application';
 export { CliApplication } from './lib/cli';
+export * from './lib/stringifiers';
 
 export { EventDispatcher, Event } from './lib/utils/events';
 export { createMinimatch } from './lib/utils/paths';
--- a/src/lib/models/types/array.ts
+++ b/src/lib/models/types/array.ts
@@ -1,4 +1,4 @@
-import { Type, UnionType, IntersectionType } from './index';
+import { Type } from './index';
 
 /**
  * Represents an array type.
@@ -49,17 +49,5 @@ export class ArrayType extends Type {
             return false;
         }
         return type.elementType.equals(this.elementType);
-    }
-
-    /**
-     * Return a string representation of this type.
-     */
-    toString() {
-        const elementTypeStr = this.elementType.toString();
-        if (this.elementType instanceof UnionType || this.elementType instanceof IntersectionType) {
-            return '(' + elementTypeStr + ')[]';
-        } else {
-            return elementTypeStr + '[]';
-        }
     }
 }
--- a/src/lib/stringifiers.ts
+++ b/src/lib/stringifiers.ts
@@ -1,0 +1,151 @@
+import * as t from './models/types';
+import * as r from './models/reflections';
+
+const stringifyList = (list: t.Type[], separator: string): string => {
+    return list.map(t => stringifyType(t)).join(separator);
+};
+
+export const stringifyArray = (node: t.ArrayType): string => {
+    const elementTypeStr = stringifyType(node.elementType);
+    if (node.elementType instanceof t.UnionType || node.elementType instanceof t.IntersectionType) {
+        return `(${elementTypeStr})[]`;
+    } else {
+        return `${elementTypeStr}[]`;
+    }
+};
+
+export const stringifyConditional = (node: t.ConditionalType): string => {
+    return `${node.checkType} extends ${node.extendsType} ? ${node.trueType} : ${node.falseType}`;
+};
+
+export const stringifyIndexedAccess = (node: t.IndexedAccessType): string => {
+    return `${stringifyType(node.objectType)}[${stringifyType(node.indexType)}]`;
+};
+
+export const stringifyInferred = (node: t.InferredType): string => {
+    return `infer ${node.name}`;
+};
+
+export const stringifyIntersection = (node: t.IntersectionType): string => {
+    return stringifyList(node.types, ' & ');
+};
+
+export const stringifyIntrinsic = (node: t.IntrinsicType): string => {
+    return node.name;
+};
+
+export const stringifyPredicate = (node: t.PredicateType): string => {
+    const out = node.asserts ? ['asserts', node.name] : [node.name];
+    if (node.targetType) {
+        out.push('is', stringifyType(node.targetType));
+    }
+
+    return out.join(' ');
+};
+
+export const stringifyQuery = (node: t.QueryType): string => {
+    return `typeof ${stringifyType(node.queryType)}`;
+};
+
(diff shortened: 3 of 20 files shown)


src/lib/stringifiers.ts
```diff
@@ -0,0 +1,151 @@
...
+    union: stringifyUnion,
+    unknown: stringifyUnknown,
+    void: stringifyVoid
+};
+
+/**
+ * Return a string representation of the given type.
+ */
+export const stringifyType = (node: t.Type): string => {
+    if (!node || !node.type || !{}.hasOwnProperty.call(stringifiers, node.type)) {
```
Gerrit0: We control `stringifiers`, no need for the paranoid version of hasOwnProperty. Also, I'd rather not worry about not being provided node/node.type. This is a TS only library with strictNullChecks on, if someone breaks that contract, that's on them.

src/test/stringifiers.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    new t.IntrinsicType('number'),
+    new t.TupleType([
+        new t.IntrinsicType('boolean'),
+        new t.IntrinsicType('string')
+    ])
+]);
+
+describe('stringifiers', function() {
+    it('maintains the `toString` method on type models', function() {
+        Assert.equal(union.toString(), stringifyType(union));
```
Gerrit0: Prefer `strictEqual`.

Also looks like we only use `throws` and `equal`, can we just import what we need?

```ts
import { throws, strictEqual as equal } from 'assert'
```

src/lib/stringifiers.ts
```diff
@@ -0,0 +1,151 @@
+import * as t from './models/types';
```
Gerrit0: So this isn't your fault, it was a problem with the existing code but nobody ran into it since it wasn't used externally... But if we are going to more publicly expose this functionality we need to fix this. The serializers can generate incorrect types. Consider this:

```
const broken = new t.IntersectionType([
  new t.ReferenceType('T', t.ReferenceType.SYMBOL_FQN_RESOLVED),
  new t.UnionType([
    new t.ReferenceType('U', t.ReferenceType.SYMBOL_FQN_RESOLVED),
    new t.ReferenceType('V', t.ReferenceType.SYMBOL_FQN_RESOLVED)
  ])
])
```

This will currently stringify to `T & U | V` while the actual type is `T & (U | V)`

There's also a problem with function types and conditional types. `() => any extends 1 ? 1 : 2` is different than `(() => any) extends 1 ? 1 : 2`.

src/index.ts
```diff
@@ -1,5 +1,6 @@
 export { Application } from './lib/application';
 export { CliApplication } from './lib/cli';
+export * from './lib/stringifiers';
```
Gerrit0: Does it make sense to export everything here? It seems to make more sense to me to only export `stringifyType` and call the correct stringifier internally.

src/lib/models/types/array.ts
```diff
@@ -1,4 +1,4 @@
-import { Type, UnionType, IntersectionType } from './index';
```
Gerrit0: :+1: One less circular dependency.

src/lib/stringifiers.ts
citycide: Agreed, I did notice there could be some precedence issues with complex union/intersection types. We'll have to add some parenthesizing logic.

src/lib/stringifiers.ts
citycide: Mostly I did this so that we could better support passing in an object pulled from typedoc JSON, given the lack of deserializing (related: #1180) and my use cases in the RFC. In that case it would probably be better to use an alternative type that ensures we only use properties that exist on both the JSON objects and the live models.

src/index.ts
citycide: True, I can make that change.

Gerrit0:
Heads up - as a part of library mode work I think I've resolved the issues with wrapping. https://github.com/TypeStrong/typedoc/blob/library-mode/src/lib/models/types/abstract.ts#L59

This still won't quite work for your JSON use case, but library mode also brings us much closer to being able to reinflate JSON into the original models.

citycide:
@Gerrit0 cool, so your `library-mode` branch subsumes the functionality of this PR? If that's the case, I'd be up for working on JSON deserialization once the dust settles. Is that something you consider in scope for the type models in typedoc?

Gerrit0:
It doesn't contain everything that this does insofar as it won't work for JSON models... I think deserialization will meet all the requirements from here.

I absolutely want to support deserialization of the full project. Tracking in #910 :)

Josegto1 (approved):

Gerrit0:
At this point, I don't think I'm going to ever merge this -- it's a nice idea, but until TypeDoc itself can be run without Node APIs, I don't want to maintain multiple entry points to typedoc.

TypeDoc 0.24 did add support for deserializing a JSON project, so at this point I'm recommending using TypeDoc's deserialization API, and then calling toString on the created type..

PR #1221, opened 2020-02-27
fix: Ensure child comment tags get set

### LINKED ISSUE #1220: Functions do no output unsupported tags

## Expected Behavior

According to [the docs](https://typedoc.org/guides/doccomments/#supported-tags): `TypeDoc renders any unsupported tags in a list in the documentation, so they are not lost`.
## Actual Behavior

This seems to work fine with class methods defined like this:
```typescript
// ... inside a class

/**
 * Test method
 * @deprecated
 */
 someMethod() { ...
```
... but when I try to add an unsupported tag to a regular function, arrow function, or even to a class method that's defined with arrow-function syntax:

```typescript
// ... inside a class

/**
 * Test method
 * @deprecated
 */
 someMethod = () => { ...

// ... or in a module instead of a class...

/**
 * Test method
 * @deprecated
 */
 export const someMethod = () => { ...

/**
 * Another test method
 * @deprecated
 */
 export function someMethod1() { ...
```
The tag doesn't show up in the generated documentation for any of those. When I output the JSON, I see that the comment has that tag, but it isn't being rendered in the output. This may very well be that I'm just doing something wrong! But I'd love to know what that is. 😄 

## Steps to reproduce the bug

- Define a class withe method, and add an `@deprecated` tag to the comment for that method
- Define another method on the class, but declare it with arrow function syntax (or if you like, create some other module that just has a regular exported function. In most of mine I have them as `const` declarations with arrow-function syntax)
- Generate the output
- Observe that the `@deprecated` tag shows up in the documentation for the first one, but not the second.

## Environment
  - Typedoc version: 0.16.10
  - Node.js version: 12.13.0
  - OS: Windows 10 / Mac OS X

epzilla:
Fix issue where child comment tags are not getting output in the docs/json.

Fixes #1220

### DIFF
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -199,6 +199,7 @@ export class CommentPlugin extends ConverterComponent {
             const comment = parseComment(rawComment, reflection.comment);
             this.applyModifiers(reflection, comment);
             this.removeExcludedTags(comment);
+            reflection.comment = comment;
         } else if (reflection.kindOf(ReflectionKind.Module)) {
             this.storeModuleComment(rawComment, reflection);
         } else {
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -303,7 +303,11 @@ export class CommentPlugin extends ConverterComponent {
                     if (!childComment) {
                         childComment = signature.comment = new Comment();
                     }
-                    childComment.copyFrom(comment);
+
+                    childComment.shortText = childComment.shortText || comment.shortText;
+                    childComment.text      = childComment.text      || comment.text;
+                    childComment.returns   = childComment.returns   || comment.returns;
+                    childComment.tags      = childComment.tags      || comment.tags;
                 }
 
                 if (signature.parameters) {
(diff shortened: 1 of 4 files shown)


src/lib/converter/plugins/CommentPlugin.ts
```diff
@@ -303,7 +303,11 @@ export class CommentPlugin extends ConverterComponent {
                     if (!childComment) {
                         childComment = signature.comment = new Comment();
                     }
-                    childComment.copyFrom(comment);
+
+                    childComment.shortText = childComment.shortText || comment.shortText;
+                    childComment.text      = childComment.text      || comment.text;
+                    childComment.returns   = childComment.returns   || comment.returns;
+                    childComment.tags      = childComment.tags      || comment.tags;
```
Gerrit0: The first fix using `copyFrom` looks like the right solution to me here. Am I missing something for why it won't work and we need to potentially share references to tags?

Thanks for doing the digging to figure out why this wasn't working!

epzilla: Umm there _was_ a reason, but I’ve forgotten what it was 😂. It was late last night. If I get a chance tonight I’ll try the other way again and run the tests, and report back what happens. Thanks!

epzilla: Ok I remember what it was now. `copyFrom` overwrites everything in the child comment that may already be there, whereas doing it the other way preserves what's in the child comment unless those fields are falsy, in which case it takes whatever is in the parent.

epzilla: Not having as much in-depth knowledge about the ramifications, I wonder if modifying the `copyFrom` method to do this "conditional copying" instead might be a clean way to do it? It looks like it's only used in 2 places in the code.

Gerrit0: Ah, that makes sense, thanks! Looks good to me.

Gerrit0:
Released  in 0.16.11 :)

Issue #1249, opened 2020-03-26
Declaration option of type ParameterType.Map can't use a defaultValue that is not part of the value map


krisztianb:
- [X] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

## Expected Behavior

When adding a declaration option for a plugin the following call should work:

```
typedoc.options.addDeclaration({
    type: ParameterType.Map,
    name: "fontStyle",
    map: new Map([
        ["normal", FontStyle.Normal],
        ["plain", FontStyle.Plain],
        ["italic", FontStyle.Italic],
        ["bold", FontStyle.Bold],
   ],
   defaultValue: FontStyle.Undefined,
});
```

A default value not listed in the map of possible values enables the plugin to verify whether or not a value was provided for the option at all. Otherwise the plugin can't know if the default value was a fallback or specified by the user. This might make a difference in the logic of the plugin.

## Actual Behavior

Upon loading the plugin the following error is thrown:

```
Error: Failed to validate default value for fontStyle
    at Result.expect (C:\MyProjects\typedoc\class-proj\node_modules\typedoc\dist\lib\utils\result.js:17:15)
    at Options.addDeclaration (C:\MyProjects\typedoc\class-proj\node_modules\typedoc\dist\lib\utils\options\options.js:59:18)
    at PluginEnumOption.addToApplication (C:\MyProjects\typedoc\class-proj\node_modules\typedoc-plugin-base\dist\plugin_enum_option.js:43:25)
    at PlantUmlPluginOptions.addToApplication (C:\MyProjects\typedoc\class-proj\node_modules\typedoc-plantuml\dist\plantuml_plugin_options.js:138:51)
    at PlantUmlPlugin.addOptionsToApplication (C:\MyProjects\typedoc\class-proj\node_modules\typedoc-plantuml\dist\plantuml_plugin.js:62:22)
    at PlantUmlPlugin.PluginBase.initialize (C:\MyProjects\typedoc\class-proj\node_modules\typedoc-plugin-base\dist\plugin_base.js:14:14)
    at load (C:\MyProjects\typedoc\class-proj\node_modules\typedoc-plantuml\dist\index.js:9:44)
    at PluginHost.load (C:\MyProjects\typedoc\class-proj\node_modules\typedoc\dist\lib\utils\plugins.js:28:21)
[...]

PR #1250, opened 2020-03-27
No need to convert defaultValue for options of type map

krisztianb:
Fixes: https://github.com/TypeStrong/typedoc/issues/1249

The convert function was used to "validate" the default value for all options. This is not necessary for options of type map, since the type of the default value is provided via the template class MapDeclarationOption and must be of the right type. Moreover the conversion throws an error if the default value for the option is of the right type, but not listed in the map of possible values.

### DIFF
--- a/src/lib/utils/options/options.ts
+++ b/src/lib/utils/options/options.ts
@@ -1,7 +1,7 @@
 import * as _ from 'lodash';
 import * as ts from 'typescript';
 
-import { DeclarationOption, ParameterScope, convert, TypeDocOptions, KeyToDeclaration, TypeDocAndTSOptions, TypeDocOptionMap } from './declaration';
+import { DeclarationOption, ParameterScope, ParameterType, convert, TypeDocOptions, KeyToDeclaration, TypeDocAndTSOptions, TypeDocOptionMap } from './declaration';
 import { Logger } from '../loggers';
 import { Result, Ok, Err } from '../result';
 import { insertPrioritySorted } from '../array';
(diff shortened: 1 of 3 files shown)


src/lib/utils/options/options.ts
```diff
@@ -320,6 +314,22 @@ export class Options {
...
     }
+
+    /**
+     * Sets the value of a given option to its default value.
+     * @param declaration The option whoes value should be reset.
+     */
+    private setOptionValueToDefault(declaration: Readonly<DeclarationOption>): void {
+        if (declaration.scope !== ParameterScope.TypeScript) {
+            // No nead to convert the defaultValue for a map type as it has to be of a specific type
+            if (declaration.type === ParameterType.Map) {
```
Gerrit0: Could you add a test for your specific usage that illustrates why this is important? The one you posted in gitter should be fine, or even something simpler.

src/lib/utils/options/options.ts
```diff
@@ -320,6 +314,22 @@ export class Options {
         }
         return errors.length ? Err(errors) : Ok(void 0);
     }
+
+    /**
+     * Sets the value of a given option to its default value.
+     * @param declaration The option whoes value should be reset.
+     */
+    private setOptionValueToDefault(declaration: Readonly<DeclarationOption>): void {
```
Gerrit0: :+1: nice to remove duplicate logic.

src/lib/utils/options/options.ts
krisztianb: Good idea. Test added. I ran it through. Previously all but the new test passed. Now all tests pass.

Gerrit0:
Thanks! I'm not sure if I'll get a release out today, next Saturday at the latest. Going to be busy...

krisztianb:
No hurry on this one. 😄

Issue #1251, opened 2020-03-29
Use json file for index search


sgrishchenko:
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem

Now `typedoc-default-themes` package use `serach.js` file for index search providing. `eval` function is called to parse data for search. I think, this is potential vulnerability.

## Suggested Solution

I think, it will be more safe to use JSON format for search data passing.

Gerrit0:
>  I think, this is potential vulnerability.

Not in this case. We are only sending JSON data. Sure, someone could put extra stuff in there via a plugin, but at that point they already have code running on someone's machine via node... which is much worse that on a site. Since documentation hosts can modify files before publishing anyways, there isn't a vulnerability here.

I do like the switch to json though.

Gerrit0:
Resolved by #1252, I hope to have time for a release tonight :)

Gerrit0:
Released in 0.17.4 :)

PR #1252, opened 2020-03-29
Generate json file for index search

sgrishchenko:
This is implementation of #1251.
Here I added additional file in JSON format and wrote warning for custom themes, which use JS format. JS file was not removed for backward compatibility. Next step: to support this changes in `typedoc-default-themes` package.

### DIFF
--- a/src/lib/output/plugins/JavascriptIndexPlugin.ts
+++ b/src/lib/output/plugins/JavascriptIndexPlugin.ts
@@ -66,12 +66,18 @@ export class JavascriptIndexPlugin extends RendererComponent {
             rows.push(row);
         }
 
-        const fileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.js');
-        const data =
-            `var typedoc = typedoc || {};
+        const jsFileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.js');
+        const jsData =
+            `console.warn('[typedoc] search.js file is deprecated. Please, use search.json file in your custom theme instead.');
+            var typedoc = typedoc || {};
             typedoc.search = typedoc.search || {};
             typedoc.search.data = ${JSON.stringify({kinds: kinds, rows: rows})};`;
 
-        writeFile(fileName, data, false);
+        writeFile(jsFileName, jsData, false);
+
+        const jsonFileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.json');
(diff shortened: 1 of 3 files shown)


src/lib/output/plugins/JavascriptIndexPlugin.ts
```diff
@@ -66,12 +66,18 @@ export class JavascriptIndexPlugin extends RendererComponent {
             rows.push(row);
         }
 
-        const fileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.js');
-        const data =
-            `var typedoc = typedoc || {};
+        const jsFileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.js');
+        const jsData =
+            `console.warn('[typedoc] search.js file is deprecated. Please, use search.json file in your custom theme instead.');
```
Gerrit0: I'd rather just remove this with the switch to `.json`. The search file is usually the biggest part of the deployed docs, so this change might otherwise nearly double the size of our output.

src/lib/output/plugins/JavascriptIndexPlugin.ts
```diff
@@ -66,12 +66,18 @@ export class JavascriptIndexPlugin extends RendererComponent {
...
+            var typedoc = typedoc || {};
             typedoc.search = typedoc.search || {};
             typedoc.search.data = ${JSON.stringify({kinds: kinds, rows: rows})};`;
 
-        writeFile(fileName, data, false);
+        writeFile(jsFileName, jsData, false);
+
+        const jsonFileName = Path.join(event.outputDirectory, 'assets', 'js', 'search.json');
```
Gerrit0: Related to this, I'd really like to switch to serializing a pre-built index https://lunrjs.com/guides/index_prebuilding.html and sending that instead. There's no reason we can't do as much work for clients before rendering as possible... probably fits best in a different PR though.

src/lib/output/plugins/JavascriptIndexPlugin.ts
sgrishchenko: I have removed js file generation.

src/lib/output/plugins/JavascriptIndexPlugin.ts
sgrishchenko: I think it is good idea, I have implemented this approach for typedoc-default-themes [here](https://github.com/TypeStrong/typedoc-default-themes/pull/101)

PR #1265, opened 2020-04-12
fix: Only set inputFiles from tsconfig if not already set

### LINKED ISSUE #1263: typedoc.json inputFiles property not working as expected
## Search terms
typedoc.json inputFiles

## Expected Behavior
When I specified the inputFiles option in the typedoc.json file to have an array (or single) of directories such as ["./src/foo/lib", "./src/bar/lib"]. It pickup all the files in the ./src file. I expect it to only search for files in the two lib folders.

## Actual Behavior

All files in the ./src folder are parsed and the generated doc contains files in other folder.

## Steps to reproduce the bug

typedoc.json
    {
        "mode": "modules",
        "out": "./docs",
        "inputFiles": ["./src/foo/lib", "./src/bar/lib"]
    }

Case #1: doesn't work
typedoc --options typedoc.json --tsconfig tsconfig.json

Case #2: work (with or without the inputFiles property in the typedoc.json file)
typedoc --options typedoc.json --tsconfig tsconfigjson "./src/foo/lib" "./src/bar/lib"

## Environment
- Typedoc version: 0.17.4
- TypeScript version: 3.8.3
- Node.js version: 12.16.1
- OS: Windows10

Gerrit0:
Note: This effectively sets the minimum Node version to 9.0.0 since it uses [util.isDeepStrictEqual](https://nodejs.org/api/util.html#util_util_isdeepstrictequal_val1_val2) and as such can't really be merged until 0.18... I didn't want to add a dependency just for this.

fix: Options.isDefault was always false when passed non-literals.

Closes #1263

### DIFF
--- a/src/lib/utils/options/options.ts
+++ b/src/lib/utils/options/options.ts
@@ -1,3 +1,4 @@
+import { isDeepStrictEqual } from 'util';
 import * as _ from 'lodash';
 import * as ts from 'typescript';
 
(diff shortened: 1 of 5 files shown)


Gerrit0:
Incidentally, Travis fails because of this. We still test on Node 8.

benmccann:
@Gerrit0 this one will need a rebase. also, you can make travis pass by removing node 8 in the `.travis.yml`: https://github.com/TypeStrong/typedoc/blob/bc20444c6b92f6d5946875a066f277346a7e437b/.travis.yml#L5

PR #1268, opened 2020-04-13
feat: Add detecting read-only properties

### LINKED ISSUE #585: Show in the generated documentation which properties are readonly.

Lhoerion:
Resolves #585

Lack of this feature surprised me, I thought that such an important feature would be here from the beginning. Also I saw previous pull request that wasn't only addressing this issue and is a bit old, so I created* new one from the foundations to make sure it's compatible.

This is my first pull request here, I've left the ability for maintainers to edit this pull request, so I'm open to any suggestions, critic, help etc.

### DIFF
--- a/src/lib/converter/factories/declaration.ts
+++ b/src/lib/converter/factories/declaration.ts
@@ -179,6 +179,7 @@ function setupDeclaration(context: Context, reflection: DeclarationReflection, n
     reflection.setFlag(ReflectionFlag.Protected, !!(modifiers & ts.ModifierFlags.Protected));
     reflection.setFlag(ReflectionFlag.Public,    !!(modifiers & ts.ModifierFlags.Public));
     reflection.setFlag(ReflectionFlag.Optional,  !!(node['questionToken']));
+    reflection.setFlag(ReflectionFlag.Readonly,  !!(modifiers & ts.ModifierFlags.Readonly));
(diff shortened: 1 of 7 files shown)


Lhoerion:
This pull request fixes #585

Gerrit0:
I'll have to pull this down this weekend and make sure it works, but it looks good at a first pass! Thanks!

Gerrit0:
Thanks! I should be able to get a release out this weekend.

Issue #1272, opened 2020-04-19
"extends" option not working in TypeDoc config


WilliamJns:
## Search terms
[In this conversation](https://github.com/TypeStrong/typedoc/issues/493), the idea of using `extends` option in the config file was raised and in [this pull request](https://github.com/TypeStrong/typedoc/commit/07fb1ce879bfcbea0c3d75e2b1a1dee8575c15ae) was implemented, I wanted to use this feature, but unfortunately, I'm having trouble using this option.

I will present the solution that I have found to solve this problem in a new pull request.
## Expected Behavior
Merge all files and options together.

## Steps to reproduce the bug
1. Create a base config file to use in all typedoc configs calling `base-typedoc.json`.

```json
{
  "name": "Sample Documentation",
  "readme": "./docs/index.md",
  "out": "./build-docs",
  "target": "ES5",
  "mode": "file",
  "media": "./docs/assets",
  "includeDeclarations": false,
  "ignoreCompilerErrors": true,
  "disableOutputCheck": true,
  "logger": "none",
  "theme": "sample"
}
```

2. Extend the created `base-typedoc.json` in original `typedoc.json`

```json
{
  "extends": "base-typedoc.json",
  "out": "./build-docs"
}
```

3.  Build the document using `typedoc --options typedoc.json`.

## Environment
- Typedoc version: 0.17.3
- TypeScript version: 3.8.3
- Node.js version: 12.16.1
- OS: MacOS Catalina 10.15.4

PR #1273, opened 2020-04-19
fix: Fix loading options twice in bootstrap to keep 'extends'.

WilliamJns:
Fix https://github.com/TypeStrong/typedoc/issues/1272

Every time we open the option files and after reading the value of extends in each one, this property is removed from that file, so the next time the value of extends is no longer available to us ([according to this code logics](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/readers/typedoc.ts#L67)).

Therefore, reading the options file more than once will cause #1272 issue.

### DIFF
--- a/src/lib/utils/options/readers/typedoc.ts
+++ b/src/lib/utils/options/readers/typedoc.ts
@@ -1,6 +1,6 @@
 import * as Path from 'path';
 import * as FS from 'fs';
-import * as _ from 'lodash';
+import { cloneDeep } from 'lodash';
 
 import { OptionsReader } from '..';
 import { Logger } from '../../loggers';
@@ -51,12 +51,14 @@ export class TypeDocReader implements OptionsReader {
         }
         seen.add(file);
 
-        const data: unknown = require(file);
+        const fileContent: unknown = require(file);
 
-        if (typeof data !== 'object' || !data) {
+        if (typeof fileContent !== 'object' || !fileContent) {
             logger.error(`The file ${file} is not an object.`);
             return;
         }
+
+        const data: object = cloneDeep(fileContent);
 
         if ('extends' in data) {
             const extended: string[] = getStringArray(data['extends']);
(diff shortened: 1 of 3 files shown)


Gerrit0:
This will break wrappers around TypeDoc. Both the Grunt and Gulp wrappers around TypeDoc use the option passing in bootstrap to let users specify config files or options directly.

A better solution would be to clone the options object before mutating it.

WilliamJns:
@Gerrit0 Thanks for the tip. Please review the new changes.

src/lib/utils/options/sources/typedoc.ts
```diff
@@ -21,6 +21,11 @@ export function addTypeDocOptions(options: Options) {
         type: ParameterType.Array
     });
 
+    options.addDeclaration({
```
Gerrit0, line 24: This implies that users can specify `--extends` on the command line, or in typedocOptions in tsconfig.json, which isn't true.

src/lib/utils/options/readers/typedoc.ts
```diff
@@ -51,13 +51,16 @@ export class TypeDocReader implements OptionsReader {
...
+        const fileContent: unknown = require(file);
 
-        if (typeof data !== 'object' || !data) {
+        if (typeof fileContent !== 'object' || !fileContent) {
             logger.error(`The file ${file} is not an object.`);
             return;
         }
 
+        // clone option object to avoid of property changes in re-calling this file
+        const data: object = cloneDeep(fileContent);
```
Gerrit0, line 62: :+1:

src/lib/utils/options/sources/typedoc.ts
Gerrit0: Best to just remove it, I still need to update the website docs to mention this, but it shouldn't show up in --help.

Gerrit0:

PR #1277, opened 2020-04-23
Do not silently swallow the missed file errors - issue warnings

SamuraiJack:
Currently if typedoc can not find the file, specified with [[include:]] "pragma" it basically ignores the error and does not give any feedback to the user. This makes hard to debug the relative paths, which starts with many directory switches, like `../../../`. A warning is very helpful for debugging this problem.


Gerrit0:

PR #1278, opened 2020-04-24
Added possibility to define min and max values for a number option

krisztianb:
Also added a couple of tests for the new feature.

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -258,27 +258,38 @@ export function convert<T extends DeclarationOption>(value: unknown, option: T):
             return Result.Ok([]);
         case ParameterType.Map:
             const optionMap = option as MapDeclarationOption<unknown>;
-            if (getTypeOf(value) === getTypeOf(optionMap.defaultValue)) {
-                return Result.Ok(value);
+            const key = String(value).toLowerCase();
+            if (optionMap.map instanceof Map) {
+                if (optionMap.map.has(key)) {
+                    return Result.Ok(optionMap.map.get(key));
+                }
+                if ([...optionMap.map.values()].includes(value)) {
+                    return Result.Ok(value);
+                }
+            } else {
+                if (optionMap.map.hasOwnProperty(key)) {
+                    return Result.Ok(optionMap.map[key]);
+                }
+                if (Object.values(optionMap.map).includes(value)) {
+                    return Result.Ok(value);
+                }
             }
-            return Result.Err(optionMap.mapError ?? `Invalid type for value of ${optionMap.name}`);
+            return Result.Err(optionMap.mapError ?? getMapError(optionMap.map, optionMap.name));
         case ParameterType.Mixed:
             return Result.Ok(value);
     }
 }
 
-/**
- * Returns the type of a value.
- * @param value The value whoes type is wanted.
- * @returns The type of the value;
- */
-function getTypeOf(value: unknown): unknown {
-    const typeOfValue = typeof value;
+function getMapError(map: MapDeclarationOption<unknown>['map'], name: string) {
+    let keys = map instanceof Map ? [...map.keys()] : Object.keys(map);
+    const getString = (key: string) => String(map instanceof Map ? map.get(key) : map[key]);
 
-    // null is also of type 'object'
-    if (typeOfValue === 'object' && value) {
-        return Object.getPrototypeOf(value) as unknown;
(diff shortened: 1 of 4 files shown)


src/test/utils/options/options.test.ts
```diff
@@ -25,6 +26,43 @@ describe('Options', () => {
...
     });
 
+    it('Does not throw if number declaration has no min and max values', () => {
+        const declaration: NumberDeclarationOption = {
+            name: 'test-number-declaration',
+            help: '',
+            type: ParameterType.Number,
+            defaultValue: 1
+        };
+        doesNotThrow(() => options.addDeclaration(declaration));
```
Gerrit0, line 36: doesNotThrow is redundant. All it does is rethrow an error if one occurs.

src/lib/utils/options/declaration.ts
```diff
@@ -246,7 +256,14 @@ export function convert<T extends DeclarationOption>(value: unknown, option: T):
         case ParameterType.String:
             return Result.Ok(value == null ? '' : String(value));
         case ParameterType.Number:
-            return Result.Ok(parseInt(String(value), 10) || 0);
+            const numberOption = option as NumberDeclarationOption;
+            const numValue = parseInt(String(value), 10) || 0;
+            if (typeof numberOption.minValue === 'number' &&  numValue < numberOption.minValue) {
+                return Result.Err(`${numberOption.name} must be >= ${numberOption.minValue}`);
```
Gerrit0, line 262: Is there a reason you don't include both bounds in the error message? It seems like it would be more useful:

```bash
$ typedoc --numOption 10
Error: numOption must be <= 5
$ typedoc --numOption 2
Error: numOption must be >= 3
# vs
$ typedoc --numOption 10
Error: numOption must be between 2 and 10
```

Probably best to extract the error message creation to a helper function, since there are several possible combinations.

src/test/utils/options/options.test.ts
krisztianb: I'm confused. What would be the right solution? I guess you understand what it is trying to assert.

src/lib/utils/options/declaration.ts
krisztianb: I agree that if both are defined it would be more useful to include both bounds in the error message.

src/lib/utils/options/declaration.ts
```diff
@@ -293,3 +315,59 @@ function getMapError(map: MapDeclarationOption<unknown>['map'], name: string) {
...
+    return isFiniteNumber(numberOption.minValue) || isFiniteNumber(numberOption.maxValue);
+}
+
+/**
+ * Checks if the given value is a finite number.
+ * @param value The value being checked.
+ * @returns True, if the value is a finite number, otherwise false.
+ */
+function isFiniteNumber(value?: unknown): value is number {
+    return typeof value === 'number' && isFinite(value);
```
Gerrit0, line 353: Well this sent me on a wild goose chase. This is equivalent to `Number.isFinite`, but the types for that function say they only accept `number`, so TS would yell... this was fixed, but the fix changed generated files instead of the source files and so was thrown out in the next release.

src/test/utils/options/options.test.ts
Gerrit0: It doesn't hurt anything, I guess there is some semantic value, but that's provided by the test name. It's fine as is.

src/lib/utils/options/declaration.ts
```diff
@@ -293,3 +315,59 @@ function getMapError(map: MapDeclarationOption<unknown>['map'], name: string) {
...
+ * @param value The value being checked.
+ * @param minValue The lower bound of the range of allowed values.
+ * @param maxValue The upper bound of the range of allowed values.
+ * @returns True, if the value is within the given bounds, otherwise false.
+ */
+function valueIsWithinBounds(value: number, minValue?: number, maxValue?: number): boolean {
+    if (isFiniteNumber(minValue) && isFiniteNumber(maxValue)) {
+        return minValue <= value && value <= maxValue;
+    } else if (isFiniteNumber(minValue)) {
+        return minValue >= value;
```
Gerrit0, line 367: Didn't see this until after I merged, I'll fix it. Should be `minValue <= value`

krisztianb: Sorry. You're right. I've probably should have written a couple of more tests and this would have popped up.

PR #1282, opened 2020-04-26
Fixed test for map declaration type

krisztianb:
I messed up the test in the original commit. Sorry about that. This should fix the test.


Gerrit0:
Thanks!

Issue #1291, opened 2020-05-01
Allow every possible number as a defaultValue for a number option


krisztianb:
## Search Terms

plugin number declaration option

## Problem

Since we allow every possible value for a map option it makes sense to do the same for numbers.

Example usage:

Given a plugin with a number option (minValue = 0 and maxValue = 10). If no value is provided the plugin should completely ignore the option.
Currently defaultValue would default to 0 and therefore there is no way for the plugin to check if the option was provided by the user or not.

## Suggested Solution

If we allow every possible number the plugin could define the defaultValue as -1 for example and then proceed accordingly.

Another solution would be not to set the defaultValue to 0 if it is not defined when calling addDeclaration.

What do you think?

Gerrit0:
This sounds reasonable to me, originally I was going to say that we should instead provide an `isSet` function, but that complicates the class considerably. This seems like a good compromise.

Gerrit0:
v0.17.7 released with this :)

PR #1296, opened 2020-05-06
Allow every possible number as a defaultValue for a number option

krisztianb:
Fixes: https://github.com/TypeStrong/typedoc/issues/1291

PS: Removed some of the tests which are no longer needed.


src/lib/utils/options/options.ts
```diff
@@ -303,6 +303,9 @@ export class Options {
             // No need to convert the defaultValue for a map type as it has to be of a specific type
             if (declaration.type === ParameterType.Map) {
                 this._values[declaration.name] = declaration.defaultValue;
+            } else if (declaration.type === ParameterType.Number) {
+                // Don't use convert for number options to allow every possible number as a default value
+                this._values[declaration.name] = declaration.defaultValue || 0;
```
Gerrit0, line 308: So, "every possible number" doesn't include NaN? I think I'm okay with this... will sleep on it.

src/test/utils/options/options.test.ts
```diff
@@ -41,53 +41,16 @@ describe('Options', () => {
...
 
-    it('Does not throw if default value is within range for number declaration', () => {
+    it('Does not throw if default value is out of range for number declaration', () => {
         const declaration: NumberDeclarationOption = {
             name: 'test-number-declaration',
             help: '',
             type: ParameterType.Number,
             minValue: 1,
             maxValue: 10,
-            defaultValue: 5
```
Gerrit0, line 51: Let's make the default value explicit so that the test is obvious and doesn't break in the (admittedly very unlikely) situation that the default value is changed.

src/lib/utils/options/options.ts
krisztianb: Yes, I was thinking about that one too. Maybe allowing `NaN` is the better solution though. Otherwise I would wonder why the plugin resets it to `0`.

src/test/utils/options/options.test.ts
krisztianb: Sorry, I don't understand. What do you mean by "make it explicit"? Do you mean to create a separate variable for it and use that in the object definition?

Gerrit0: I meant specifying `defaultValue: 0` instead of letting it default.

src/lib/utils/options/options.ts
Gerrit0: I think disallowing `NaN` is a good idea. It is such a common source of bugs that I'd rather just disallow it entirely. If someone *really* wants an obviously bad value, they can use `-Infinity` or something.

src/test/utils/options/options.test.ts
krisztianb: Oh, sure. Somehow my brain ignored that the line has a red background (meaning it is no longer there).

krisztianb: Well, it turns out it is already specified just that way: `defaultValue: 0`. Don't know why it does not show up as such here.

src/lib/utils/options/options.ts
krisztianb: @Gerrit0 Yeah, that's probably the better solution for now. That way you can be sure to have "a number" as a value.

src/test/utils/options/options.test.ts
Gerrit0: Doh! So it is, diff is great until it isn't. Looks good to me.

PR #1301, opened 2020-05-12
fix: Module declaration parsed as namespace

### LINKED ISSUE #1284: `declare module` parsed as `namespace`
This input:
```typescript
declare module "m" {
  export function f();
}
```

generates this JSON output:
```json
{
	"id": 0,
	"name": "",
	"kind": 0,
	"flags": {},
	"originalName": "",
	"children": [
		{
			"id": 1,
			"name": "\"m\"",
			"kind": 2,
			"kindString": "Namespace",
			"flags": {
				"isExported": true
			},
			"children": [
				{
					"id": 2,
					"name": "f",
					"kind": 64,
					"kindString": "Function",
					"flags": {
						"isExported": true
					},
					"signatures": [
						{
							"id": 3,
							"name": "f",
							"kind": 4096,
							"kindString": "Call signature",
							"flags": {
								"isExported": true
							},
							"type": {
								"type": "intrinsic",
								"name": "any"
							}
						}
					],
				}
			],
			"groups": [
				{
					"title": "Functions",
					"kind": 64,
					"children": [
						2
					]
				}
			],
		}
	],
	"groups": [
		{
			"title": "Namespaces",
			"kind": 2,
			"children": [
				1
			]
		}
	]
}
```

I was expecting reflection with id = 1 to be of kind = 1 (module), not of kind = 2 (namespace)

Lhoerion:
Fixes #1284

I stumbled upon this particular issue and noticed that module declaration lacks `ts.NodeFlags.Namespace` in AST, so it was very easy to "fix". I've checked if it works, but it may need further investigation.


Gerrit0:
Thanks for the PR! It looks like there's a problem with this.

```ts
module Foo { }
```

and

```ts
namespace Foo { }
```

Should result in the same output (both are namespaces according to TD).
Eventually, we'll be able to use the cleaner check that you propose, but since the former code block is still valid, we can't use these flags.

---

The TS docs have apparently removed any useful reference to this legacy syntax. It's still on archive.org though. https://web.archive.org/web/20190317062641/https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html

Lhoerion:
The latest commit should address this "misbehaviour".
I had to use casting to `any` for `ts.NodeFlags.Ambient` because of being marked as internal and in result missing only in types declaration.

Gerrit0:
I'm afraid that doesn't work either. Ambient doesn't mean module.

```ts
declare namespace Foo {} // Has Ambient flag
```

I think the best we can do for now is check `ts.isStringLiteral(node.name)`

Lhoerion:
What you posted is actually the topic of the issue, so everything is taken into the consideration. Ambient declaration is the only exception that have distinguished namespace and module and in result should have separate output. I don't think if double quotes decide about being a module or not

src/lib/converter/nodes/module.ts
```diff
@@ -24,7 +24,9 @@ export class ModuleConverter extends ConverterNodeComponent<ts.ModuleDeclaration
     convert(context: Context, node: ts.ModuleDeclaration): Reflection | undefined {
         const reflection = context.isInherit && context.inheritParent === node
             ? <DeclarationReflection> context.scope
-            : createDeclaration(context, node, ReflectionKind.Namespace);
+            : createDeclaration(context, node, node.name.kind === 10
```
Gerrit0, line 27: Using the kind is fine (that's all the function does) but the inline constant could break across different TS versions. We should instead use the enum.

```suggestion
            : createDeclaration(context, node, node.name.kind === ts.SyntaxKind.StringLiteral
```

Gerrit0: I should have time later today or tomorrow to pull this down and fix if you don't have time.

Gerrit0:

PR #1303, opened 2020-05-13
fix: copy inherited parameter descriptions

### LINKED ISSUE #787: @inheritDoc is not copying parameter documentation

StephanBijzitter:
Original change by @kayahr in
https://github.com/TypeStrong/typedoc/pull/788 but was deleted
before the pull request could be merged

Closes #788, closes #787

### DIFF
--- a/src/lib/converter/plugins/ImplementsPlugin.ts
+++ b/src/lib/converter/plugins/ImplementsPlugin.ts
@@ -3,6 +3,7 @@ import { Type, ReferenceType } from '../../models/types/index';
 import { Component, ConverterComponent } from '../components';
 import { Converter } from '../converter';
 import { Context } from '../context';
+import { Comment } from '../../models/comments/comment';
 
 /**
  * A plugin that detects interface implementations of functions and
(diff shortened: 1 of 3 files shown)


Gerrit0:
Thanks! This seems to work perfectly :)

PR #1304, opened 2020-05-15
fix: Minimatch pattern assertion on Windows OS

Lhoerion:
I'm not entirely sure if that does the job (doesn't come up with some repercussions), but I don't have experience with minimatch package.
You can experience an error on Windows, when you've your project on other drive than system drive. With the things how `absolutePath` expression works, you won't land magically on system drive like the comparison pattern expects you to, so I removed the `C:` from the beginning of the pattern.


Gerrit0:
Thanks!

Issue #1256, opened 2020-03-30
@link to field with link text does not render as hyperlink


pluma:
- [x] I have checked [issues with bug label](https://github.com/TypeStrong/typedoc/labels/bug) and found no duplicates

The implementation seems to be incomplete.

The following works as expected:

`{@link http://example.com | Something neat}`

As does this:

`{@link AcmeFactory.doStuff}`

But this does not:

`{@link AcmeFactory.doStuff | Link text here}`

It doesn't seem to be recognised at all and is output as-is.

This syntax is used in the playground: https://microsoft.github.io/tsdoc/

Gerrit0:
TypeDoc doesn't yet support TSDoc - it existed for years before TSDoc did and has its own (inferior) system for parsing comments.

That said, this should be a relatively easy fix. It looks like the intent to support this was there, we just do it badly.  PR welcome. https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/plugins/MarkedLinksPlugin.ts#L147

Gerrit0:
Fixed in 0.17.7 - 935e10a469a54282ee79911f1a82eddccef7739a

pluma:
Not sure where this belongs but I just noticed VSCode seems to think the `|` (pipe) is part of the link text rather than part of the link syntax.

Gerrit0:
Yea... that's a VSCode issue. I'm not sure what it uses for parsing comments, but it isn't TypeDoc.

PR #1306, opened 2020-05-17
[fix] link caption issue

jun-sheaf:
See reference #1256.


Gerrit0:
I believe this was fixed with 935e10a

Gerrit0:
I just checked and it seems my initial check was correct, this is already fixed on master, but a release with the fix hasn't been published yet (possibly tonight, definitely tomorrow if not tonight). Thanks for the PR anyways!

jun-sheaf:
Alright then 👍 That's great!

PR #1308, opened 2020-05-18
Exclude not documented tweaks

canonic-epicure:
This PR improves the `--excludeNotDocumented` behavior - it always includes all properties of the object type, if it has documentation (even if the properties do not have docs). 

This seems to be an expected behavior for objects, which are used to document various structures in code, for example:

```ts
/**
 * Result type of `someMethod` (Object type has docs)
 */
export type SomeMethodResult = { noErrors : boolean, results : [] }
```
This behavior does not apply for classes/interfaces - for those if the property is not documented it is not included.


Gerrit0:
Thanks! Should have a release later today.

PR #1313, opened 2020-05-24
Fix #1294

### LINKED ISSUE #1294: baseUrl does not affect module names
## Search terms
baseUrl, module, name, directory

## Folder Structure
```
src/
    a/
        a.ts
        b.ts
        c.ts
```

```baseUrl``` = ```src```

## Actual Behavior
Modules are displayed as:
```
a
b
c
```

## Expected Behavior
Using ```baseUrl```:
```
a/a
a/b
a/c
```

## Steps to reproduce the bug
Create similar file structure with ```baseUrl``` and run TypeDoc.

## Environment
- Typedoc version: Latest
- TypeScript version: Builtin To TypeDoc
- Node.js version: 10.19.0
- OS: Ubuntu 20.04

### PR DESCRIPTION
**Pull request by TheBrokenRail**
Fixes #1294

### DIFF
--- a/src/lib/converter/plugins/DynamicModulePlugin.ts
+++ b/src/lib/converter/plugins/DynamicModulePlugin.ts
@@ -61,6 +61,9 @@ export class DynamicModulePlugin extends ConverterComponent {
             name = name.replace(/"/g, '');
             this.reflections.push(reflection);
             this.basePath.add(name);
+            if (context.getCompilerOptions().baseUrl) {
+                this.basePath.add(context.getCompilerOptions().baseUrl as string + '/file');
+            }
         }
     }
 
--- a/src/lib/converter/plugins/DynamicModulePlugin.ts
+++ b/src/lib/converter/plugins/DynamicModulePlugin.ts
@@ -42,6 +42,9 @@ export class DynamicModulePlugin extends ConverterComponent {
     private onBegin(context: Context) {
         this.basePath.reset();
         this.reflections = [];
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/lib/converter/plugins/DynamicModulePlugin.ts [hunk 1]
```diff
@@ -61,6 +61,9 @@ export class DynamicModulePlugin extends ConverterComponent {
             name = name.replace(/"/g, '');
             this.reflections.push(reflection);
             this.basePath.add(name);
+            if (context.getCompilerOptions().baseUrl) {
+                this.basePath.add(context.getCompilerOptions().baseUrl as string + '/file');
```
TheBrokenRail, line 65: This is needed because ```BasePath``` expects a file and then get the dirname, turning ```a/b/c``` into ```a/b``` this counteracts that.

TheBrokenRail:
Any updates?

Gerrit0:
Can we add the baseUrl in `onBegin` instead of `onDeclaration`? The efficiency gain should be minor, but I'd rather not be slow when it's easy to use a faster option.

TheBrokenRail:
Done!

Gerrit0:
Can we do that for the SourcePlugin as well?

TheBrokenRail:
```SourcePlugin``` already has it in ```onBeginDocument```, but it can't be in ```onBegin``` because ```onBegin``` is not passed an instance of ```Context```.

Gerrit0:
Makes sense, thanks! Putting together a release now.

PR #1318, opened 2020-06-04
fix: GithubPlugin: read correct remote when multiple github repos exist

christopherthielen:
When a typedoc project includes multiple github repositories, the remote URL is always set to the root repository's url:

Say you have sources like this with repo1 including a git submodule of repo2:

```
src - https://github.com/user/repo1.git
+- sourcefile.ts
+- submodule - https://github.com/user/repo2.git
  +- src
    +- sourcefile_in_other_repo.ts
```

Today, the plugin will:

- pushd into `/src/submodule/src`
- get the path to the repo root
- popd back to `/src` _(whoops)_
- get the remote of `/src` _when it should be getting the remote of `/src/submodule/src` instead_

This PR ensures the remote is fetched before the popd happens.


Gerrit0:
Thanks!

PR #1348, opened 2020-08-15
Reflect template parameter default values and use them as type arguments

### LINKED ISSUE #1341: Default values for type parameters are not used as type arguments
## Search terms

type-parameters, type-arguments

## Expected Behavior

Default values for type parameters are used as type arguments.

## Actual Behavior

When a class is extending the binding of a template class with some type parameters getting no type arguments, the default values for those type parameters are not used and instead the type parameters are shown on the page of the class, which makes the class look generic while it isn't.

Example:

````typescript
class TemplateClass<A = {}, B = any> {
}

class ConreteClass extends TemplateClass<string> {
}
````

Results in a typedoc page for `ConreteClass` where `ConreteClass` has a type parameter `B`, instead of it being derived from `TemplateClass<string, any>`.

## Steps to reproduce the bug

Create the documentation for the code above.

## Environment
- Typedoc version: 0.17.8
- TypeScript version: 3.9.6
- Node.js version: 12.13.0
- OS: Windows 10 (64-bit)

### LINKED ISSUE #1345: Type parameters of template classes should not be inherited
## Search terms

type-parameters, React.Component

## Expected Behavior

When a class is extending a binding of the template class `React.Component` there should be no type parameters visible, because the extending class is not generic, but its base class is. Moreover the type parameters are: P, S and SS.

Example code:

````typescript
class AlarmGrid extends React.Component<PropsFromRedux> {
  // *snip*
}
````

## Actual Behavior

When a class is extending a binding of the template class `React.Component` the following type parameters are shown: S, SS and S. Which means that P is missing and S is displayed twice. But there should be none at all, as described above.

typedoc screenshot

This was originally reported by by @Kerry-Cho on Gitter [here](https://gitter.im/TypeStrong/typedoc?at=5ee0d19d035dfa126110c976).

## Steps to reproduce the bug

1. Create a simple project
2. Do `npm install react` and `npm install --save-dev @types/react`
3. Create a class extending `React.Component` (see example code above)
4. Create the documentation using typedoc

## Environment
- Typedoc version: 0.17.8
- TypeScript version: 3.9.6
- Node.js version: 12.13.0
- OS: Windows 10 (64-bit)

### PR DESCRIPTION
**Pull request by krisztianb**
Fixes #1341 and Fixes #1345 

When reviewing please take a closer look at the changes here: https://github.com/TypeStrong/typedoc/compare/master...krisztianb:fix-react-template-parameters?expand=1#diff-6fc223c798b8302651796c1b41fa064a

There is a strange code regarding mixins there. I'm not entirely sure that my logic is correct.

### DIFF
--- a/src/lib/converter/utils/types.ts
+++ b/src/lib/converter/utils/types.ts
@@ -7,7 +7,8 @@ import * as ts from 'typescript';
  */
 export function getTypeParametersOfType(type: ts.Type): ts.TypeParameterDeclaration[] {
     for (const declaration of type.symbol.declarations) {
-        if (ts.isClassDeclaration(declaration) && declaration.typeParameters) {
+        if ((ts.isClassDeclaration(declaration) || ts.isInterfaceDeclaration(declaration)) &&
+             declaration.typeParameters) {
             return declaration.typeParameters.map(tp => tp);
         }
     }
--- a/src/lib/converter/utils/types.ts
+++ b/src/lib/converter/utils/types.ts
@@ -5,11 +5,13 @@ import * as ts from 'typescript';
  * @param type The type whos type parameters are wanted.
  * @returns The type parameters of the type. An empty array if the type has no type parameters.
  */
-export function getTypeParametersOfType(type: ts.Type): ts.TypeParameterDeclaration[] {
-    for (const declaration of type.symbol.declarations) {
+export function getTypeParametersOfType(type: ts.Type): ReadonlyArray<ts.TypeParameterDeclaration> {
+    const declarations = type.getSymbol()?.getDeclarations() ?? [];
+
+    for (const declaration of declarations) {
         if ((ts.isClassDeclaration(declaration) || ts.isInterfaceDeclaration(declaration)) &&
              declaration.typeParameters) {
-            return declaration.typeParameters.map(tp => tp);
+            return declaration.typeParameters;
         }
     }
 
@@ -25,20 +27,24 @@ export function getTypeParametersOfType(type: ts.Type): ts.TypeParameterDeclarat
  */
 export function getTypeArgumentsWithDefaults(
     typeParams: ts.TypeParameterDeclaration[],
-    typeArguments?: ts.NodeArray<ts.TypeNode>
-): ts.NodeArray<ts.TypeNode> {
+    typeArguments?: ReadonlyArray<ts.TypeNode>
+): ReadonlyArray<ts.TypeNode> {
     if (!typeArguments || typeParams.length > typeArguments.length) {
         const typeArgumentsWithDefaults = new Array<ts.TypeNode>();
 
         for (let i = 0; i < typeParams.length; ++i) {
             if (typeArguments && typeArguments[i]) {
                 typeArgumentsWithDefaults.push(typeArguments[i]);
-            } else if (typeParams[i].default) {
-                typeArgumentsWithDefaults.push(typeParams[i].default!);
+            } else {
+                const defaultType = typeParams[i].default;
+
+                if (defaultType) {
+                    typeArgumentsWithDefaults.push(defaultType);
+                }
             }
         }
 
-        return ts.createNodeArray<ts.TypeNode>(typeArgumentsWithDefaults);
+        return typeArgumentsWithDefaults;
     }
 
     return typeArguments;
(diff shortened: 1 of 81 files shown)

### CODE REVIEW

File: src/lib/converter/utils/types.ts [hunk 1]
```diff
@@ -0,0 +1,45 @@
+import * as ts from 'typescript';
+
+/**
+ * Returns the type parameters of a given type.
+ * @param type The type whos type parameters are wanted.
+ * @returns The type parameters of the type. An empty array if the type has no type parameters.
+ */
+export function getTypeParametersOfType(type: ts.Type): ts.TypeParameterDeclaration[] {
+    for (const declaration of type.symbol.declarations) {
```
Gerrit0, line 9: `symbol` might be undefined, and `symbol.declarations` might be undefined, leading to a crash, prefer `type.getSymbol()?.getDeclarations() ?? []`

src/lib/converter/utils/types.ts
```diff
@@ -0,0 +1,45 @@
...
+        if ((ts.isClassDeclaration(declaration) || ts.isInterfaceDeclaration(declaration)) &&
+             declaration.typeParameters) {
+            return declaration.typeParameters.map(tp => tp);
```
Gerrit0, line 12: Why the map? To remove `readonly` from the array type? I'd rather just annotate this as returning a readonly array.

src/lib/converter/utils/types.ts
```diff
@@ -0,0 +1,45 @@
...
+    typeArguments?: ts.NodeArray<ts.TypeNode>
+): ts.NodeArray<ts.TypeNode> {
+    if (!typeArguments || typeParams.length > typeArguments.length) {
+        const typeArgumentsWithDefaults = new Array<ts.TypeNode>();
+
+        for (let i = 0; i < typeParams.length; ++i) {
+            if (typeArguments && typeArguments[i]) {
+                typeArgumentsWithDefaults.push(typeArguments[i]);
+            } else if (typeParams[i].default) {
+                typeArgumentsWithDefaults.push(typeParams[i].default!);
```
Gerrit0, line 37: You shouldn't need a non-null assertion here.

src/lib/converter/utils/types.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @returns The complete list of type arguments with possible default values if type arguments are missing.
+ */
+export function getTypeArgumentsWithDefaults(
+    typeParams: ts.TypeParameterDeclaration[],
+    typeArguments?: ts.NodeArray<ts.TypeNode>
+): ts.NodeArray<ts.TypeNode> {
+    if (!typeArguments || typeParams.length > typeArguments.length) {
+        const typeArgumentsWithDefaults = new Array<ts.TypeNode>();
+
+        for (let i = 0; i < typeParams.length; ++i) {
```
Gerrit0, line 33: This has the potential to bite us if ignoreCompilerErrors is set... But if that's set there are many other things that will go wrong, so this is fine.

src/lib/converter/utils/types.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Returns a list of type arguments. If a type parameter has no corresponding type argument, the default type
+ * for that type parameter is used as the type argument.
+ * @param typeParams The type parameters for which the type arguments are wanted.
+ * @param typeArguments The type arguments as provided in the declaration.
+ * @returns The complete list of type arguments with possible default values if type arguments are missing.
+ */
+export function getTypeArgumentsWithDefaults(
+    typeParams: ts.TypeParameterDeclaration[],
+    typeArguments?: ts.NodeArray<ts.TypeNode>
+): ts.NodeArray<ts.TypeNode> {
```
Gerrit0, line 29: Why the node array? Is there some reason we can't just use `ts.TypeNode[]`?

src/lib/converter/utils/types.ts
krisztianb: That's what I thought too. But for some strange reason it is needed.

src/lib/converter/utils/types.ts
krisztianb: I did it, because I couldn't use the `NodeArray` type as input to `Array.concat` later on. I find these TS types a big confusing. But `ReadonlyArray` is the better solution.

src/lib/converter/utils/types.ts
krisztianb: Unfortunately a `NodeArray` is required by `context.inherit` later on. But I changed it a bit.

Gerrit0:
Thanks! Hoping to get a release out tonight.

Gerrit0:
Released in 0.19 :)

PR #1352, opened 2020-08-23
fix: Address typo in log statement "directory has be set as the..."

MathBunny:
Before: 

> "The './docs' directory has be set as the output location by default" 

After: 

> "The './docs' directory has **been** set as the output location by default"


Gerrit0:
Thanks!

PR #1354, opened 2020-08-23
fix: Support for type operators `readonly` & `unique` (#1336)

### LINKED ISSUE #1336: `readonly T[]` is exported as `keyof T[]`
## Search terms

readonly, keyof

## Expected Behavior

For the following code:
```ts
export const a: readonly string[] = [];
```

typedoc should export:

```
a: readonly string[] = []
```

## Actual Behavior

typedoc exports:

```
a: keyof string[] = []
```

image

## Environment
- Typedoc version: 0.17.8
- TypeScript version: 3.9.6
- Node.js version: 14.4.0
- OS: macOS 10.15.4

ericzhangjx:
Fixes #1336 

BTW although all tests passed, I found no test files for `converter/types/type-operator.ts` or `models/type/type-operator.ts`. I can only use the built package to generate docs for my working project (and confirmed the correctness). Maybe a revision of tests should also be involved.

### DIFF
--- a/src/lib/cli.ts
+++ b/src/lib/cli.ts
@@ -61,7 +61,7 @@ export class CliApplication extends Application {
                 }
                 if (!this.out && !this.json) {
                     this.logger.log("No 'out' or 'json' option has been set");
-                    this.logger.log("The './docs' directory has be set as the output location by default");
+                    this.logger.log("The './docs' directory has been set as the output location by default");
                     this.generateDocs(project, './docs');
                 }
                 if (this.logger.hasErrors()) {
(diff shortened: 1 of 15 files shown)


Gerrit0:
Looks good! I would like to have a test
- Add a new file in `src/test/converter/types` for type operators
- Export some types which use these operators
- Rebuild the specs with `npm run rebuild_specs converter types`

ericzhangjx:
Test cases for type operators `unique`, `readonly` and `keyof` added (much easier than I thought!)

ericzhangjx:
Okay this looks weird... Some other commits are involved.
I'm thinking if I wrongly executed `git pull typedoc master` before adding test cases. Should I close this pr and open another one?

PR #1361, opened 2020-09-01
Check for `isNamedTupleMember` and `isIdentifierOrPrivateIdentifier` before calling to restore backward compatibility with typescript 3.7.5 and 3.9.7 respectively

Dergash:
Fix backward compatibility for typescript 3.7.5 and 3.9.7 by checking the existence of missing ts methods before calling them.
See https://github.com/TypeStrong/typedoc/issues/1362

### DIFF
--- a/src/lib/converter/context.ts
+++ b/src/lib/converter/context.ts
@@ -410,7 +410,7 @@ export class Context {
 
 function isNamedNode(node: ts.Node): node is ts.Node & { name: ts.Identifier | ts.PrivateIdentifier | ts.ComputedPropertyName } {
     return node['name'] && (
-        ts.isIdentifierOrPrivateIdentifier(node['name']) ||
+        ts.isIdentifierOrPrivateIdentifier && ts.isIdentifierOrPrivateIdentifier(node['name']) ||
         ts.isComputedPropertyName(node['name'])
     );
 }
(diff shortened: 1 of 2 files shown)


Gerrit0:
I'm not against backwards compatibility in theory, but in practice it is a nightmare that I don't really have time to maintain.

You fixed the two points that generally cause crashes when running with an unsupported version... but there are at least three more that I remember off the top of my head.

- `ts.TupleTypeNode` changed the property that contained elements are within, so it still doesn't work in 3.x. 
- `ts.KeywordTypeNode` doesn't include the same set of types across 3.x and 4.x.
- `ts.isPrivateIdentifier` is used within the class (property?) converter, and will crash that.

... and that's just what I remember. There are probably another half dozen breaking changes that I don't remember in the past couple of TS version upgrades.

Another problem with trying to continue supporting old versions is that if we do that, we also need to test with old versions and have some process for deciding when to stop supporting them.

Dergash:
Your points are perfectly valid and I probably wouldn't even bother with creating an issue if I've had noticed the peer dependency warning before I went to fix it, so I haven't any arguments for the change aside of 'it's really two lines and support for popular create-react-template tool is back.

Feel free to close.

Gerrit0:
Major + 1 is a *long* time. 3.0 was more than 2 years ago. I can see supporting 3.9 until 4.1 comes out, but I really don't want to go further back than one minor version. The TS team only releases one every 3 months or so... create-react-app should upgrade their TypeScript version. There's already enough cruft in this codebase without trying to support multiple versions. It's fine for ts-node to support back to 2.7... they just use the transpilation API, which hasn't really changed... but nearly every version of TypeScript has some change that breaks TypeDoc.

I've made the necessary changes to work with TypeScript 3.9 in 833ebb81127c240e09cbe6367ebaf335ce0a0ef4 + added you as an author. 3.9 support is included in version 0.19.1

PR #1378, opened 2020-10-16
Grammar fix in readme.md

malienist:
Edit in readme.md:

Changed - 
`Specifies the location and file name a json file describing the project is written to.`

to - 
`Specifies the location and file name of a json file describing the project is written to.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ For a complete list of the command line arguments run `typedoc --help` or visit 
 -   `--options`<br>
     Specify a json option file that should be loaded. If not specified TypeDoc will look for 'typedoc.json' in the current directory.
 -   `--json <path/to/output.json>`<br>
-    Specifies the location and file name a json file describing the project is written to. When specified no documentation will be generated.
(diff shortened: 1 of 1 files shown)


Gerrit0:
I believe the original wording is correct, if less clear than it could be. It is saying that `--json` can be used to specify where the JSON-serialized documentation can be written to. The reworded statement is a sentence fragment now. I think we can do better... perhaps something like:

> Specifies the location and file to write json documentation to.

README.md
```diff
@@ -56,7 +56,7 @@ For a complete list of the command line arguments run `typedoc --help` or visit
 -   `--options`<br>
     Specify a json option file that should be loaded. If not specified TypeDoc will look for 'typedoc.json' in the current directory.
 -   `--json <path/to/output.json>`<br>
-    Specifies the location and file name a json file describing the project is written to. When specified no documentation will be generated.
+    Specifies the location and file name of a json file describing the project is written to. When specified no documentation will be generated.
```
papb, line 59:
```suggestion
    Specifies the path to write the json file describing the project. When specified, no documentation will be generated - only this json file will be.
```

PR #1384, opened 2020-10-28
fix: Fix bug in ReferenceType equality check

### LINKED ISSUE #1383: ReferenceType.equals() returns incorrectly returns true when symbolFullyQualifiedName fields are different
## Search terms

ReferenceType, equals, symbolFullyQualifiedName

## Expected Behavior

`ReferenceType.equals()` should return false when `symbolFullyQualifiedName` isn't equal.

Example:

```ts
const type1 = new ReferenceType("Type1", "Type1");
const type2 = new ReferenceType("Type2", "Type2");

equal(type1.equals(type2), false); // should return false, but returns true instead
```

## Actual Behavior

`ReferenceType.equals()` returns true because `reflection` on both types are `undefined` and satisfies the `|| other.reflection === this.reflection` condition.

## Steps to reproduce the bug

```ts
const type1 = new ReferenceType("Type1", "Type1");
const type2 = new ReferenceType("Type2", "Type2");

equal(type1.equals(type2), false); // should return false, but returns true instead
```

## Environment

-   Typedoc version: 0.19.2
-   TypeScript version: 4.0.3
-   Node.js version: v10.19.0
-   OS: macOS

socsieng:
fixes #1383

### DIFF
--- a/src/lib/models/types/reference.ts
+++ b/src/lib/models/types/reference.ts
@@ -92,7 +92,9 @@ export class ReferenceType extends Type {
         return (
             other instanceof ReferenceType &&
             (other.symbolFullyQualifiedName === this.symbolFullyQualifiedName ||
-                other.reflection === this.reflection)
+                (other.reflection === this.reflection &&
+                    other.reflection != null &&
+                    this.reflection != null))
         );
     }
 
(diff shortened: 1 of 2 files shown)


Gerrit0:
If you can fix the lint issues - looks good to me.

socsieng:
> If you can fix the lint issues - looks good to me.

Apologies. Should have checked first.

Done.

Gerrit0:
Perfect, thanks!

PR #1398, opened 2020-11-21
Add validation function to declaration options

### LINKED ISSUE #1316: Validation function for string declaration options
## Search Terms
plugin string declaration option

## Problem
If a plugin has a string parameter it is pretty common that the provided value of this parameter has to be of a specific form. For example in [this plugin](https://github.com/krisztianb/typedoc-umlclass) there are several options to define a color value in this form `<transparent|#RGBHEX>`.

It would be nice if the plugin didn't have to validate the option values itself. Instead it could define a validation logic when declaring the option and typedoc would do the validation. After all we already have min and max values for number declaration options.

## Suggested Solution
Two solutions spring to my mind for this:

1. Define an optional regular expression with the string declaration option which is used in the convert function to validate the provided value.
2. Define a validation function with the string declaration option which is called by typedoc. The function could return just a boolean or a possible error message.

krisztianb:
Fixes #1316 

Overview:

* New `validate` property for declaration options
* New tests for the new properties

Please pay attention to line **279** in `src/lib/utils/options/declaration.ts`. I removed one of the signatures. This enables TS's type refinement and thence no more type assertion is required when accessing the option. I don't know if this breaks anything else.

### DIFF
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -174,6 +174,7 @@ export interface StringDeclarationOption extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: string) => void;
 }
@@ -198,6 +199,7 @@ export interface NumberDeclarationOption extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: number) => void;
 }
@@ -212,6 +214,7 @@ export interface BooleanDeclarationOption extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: boolean) => void;
 }
@@ -226,6 +229,7 @@ export interface ArrayDeclarationOption extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: string[]) => void;
 }
@@ -240,6 +244,7 @@ export interface MixedDeclarationOption extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: unknown) => void;
 }
@@ -267,6 +272,7 @@ export interface MapDeclarationOption<T> extends DeclarationOptionBase {
 
     /**
      * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
      */
     validate?: (value: T) => void;
 }
--- a/src/lib/utils/options/declaration.ts
+++ b/src/lib/utils/options/declaration.ts
@@ -171,6 +171,11 @@ export interface StringDeclarationOption extends DeclarationOptionBase {
      * An optional hint for the type of input expected, will be displayed in the help output.
      */
     hint?: ParameterHint;
+
+    /**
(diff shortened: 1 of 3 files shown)


src/lib/utils/options/declaration.ts
```diff
@@ -199,6 +211,12 @@ export interface BooleanDeclarationOption extends DeclarationOptionBase {
      * If not specified defaults to false.
      */
     defaultValue?: boolean;
+
+    /**
+     * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
+     */
+    validate?: (value: boolean) => void;
```
Gerrit0, line 219: I don't think we want a validate function here. It doesn't make sense to me to validate a true/false value.... If only one is allowed, it should not be an option.

src/lib/utils/options/declaration.ts
```diff
@@ -238,6 +269,12 @@ export interface MapDeclarationOption<T> extends DeclarationOptionBase {
      * Optional override for the error reported when an invalid key is provided.
      */
     mapError?: string;
+
+    /**
+     * An optional validation function that validates a potential value of this option.
+     * The function must throw an Error if the validation fails and should do nothing otherwise.
+     */
+    validate?: (value: T) => void;
```
Gerrit0, line 277: I also don't think validate makes sense here... mapped option types are already constrained to a specific set of values.

src/lib/utils/options/declaration.ts
```diff
@@ -276,66 +313,86 @@ export function convert<T extends DeclarationOption>(
...
-            return [];
+            return strArrValue;
+        }
         case ParameterType.Map: {
-            const optionMap = option as MapDeclarationOption<unknown>;
             const key = String(value).toLowerCase();
-            if (optionMap.map instanceof Map) {
-                if (optionMap.map.has(key)) {
-                    return optionMap.map.get(key);
+            let mapValue;
```
Gerrit0, line 366: There's an implicit any here,  `let mapValue: unknown`

src/lib/utils/options/declaration.ts
```diff
@@ -276,66 +313,86 @@ export function convert<T extends DeclarationOption>(
...
-                if (Object.values(optionMap.map).includes(value)) {
-                    return value;
+                if (Object.values(option.map).includes(value)) {
+                    mapValue = value;
                 }
             }
-            throw new Error(
-                optionMap.mapError ?? getMapError(optionMap.map, optionMap.name)
-            );
+            if (mapValue == undefined) {
```
Gerrit0, line 382: This isn't sufficient. Someone could conceivably have a map that looks like this: `{ null: null, undefined: undefined }` to change how missing values are output in JSON.

Gerrit0 (changes requested):
I need to pull it down and play with it, I suspect that the signature changes are okay

src/lib/utils/options/declaration.ts
krisztianb: I was thinking about this too. I figured that one might want to validate based on some other value(s), eg: the value of another option, the current date :wink: or whatever. What do you think?

src/lib/utils/options/declaration.ts
krisztianb: Sorry, I don't get it.

src/lib/utils/options/declaration.ts
krisztianb: Same as for boolean option?

src/lib/utils/options/declaration.ts
Gerrit0: Yep, that looks good!

src/lib/utils/options/declaration.ts
Gerrit0: Hmm... at that point, I think plugin authors would be better off doing validation when they are going to use the options. There's no guarantee about the order which options are validated in, so doing this would likely result in duplicating code.

krisztianb: Okay, you have the last word.

Gerrit0:
Travis is entirely too slow... will merge as soon as it passes, want to pull it into 0.20.

PR #1406, opened 2020-11-29
Fix eslint no-unused-locals errors

krisztianb:
Enable ESLint "no-unused-locals" rule in the project.

I'm open to other suggestions on how to handle unneeded parameters of methods:

* Add an `eslint-disable` statement instead
* Remove them, but by commenting them out instead of simply deleting them


Gerrit0:
Thanks! I'm fine with disabling eslint as you did.

PR #1412, opened 2020-12-06
New TypeDoc option "markedOptions"

### LINKED ISSUE #1395: Disable email address escaping with HTML character references
## Search Terms

Markedjs, Marked, email, mangle

## Problem

Autolinked email addresses in the generated files are escaped with HTML character references.
On every run of TypeDoc, the escape sequence is randomly generated.
The result is, that the generated HTML files differ from the previous version, even if the meaningful content doesn't change. A version management system detects that as change then.

## Suggested Solution

There should be a way to disable atleast the `mangle` option of Marked (see https://marked.js.org/using_advanced).

Even better, if all Marked options would be configurable in the `typedoc.json` like:
```
"markedOptions": {
  "mangle": false
}
```

krisztianb:
Closes #1395

### DIFF
--- a/src/lib/output/plugins/MarkedPlugin.ts
+++ b/src/lib/output/plugins/MarkedPlugin.ts
@@ -93,39 +93,6 @@ export class MarkedPlugin extends ContextAwareRendererComponent {
         Handlebars.registerHelper("relativeURL", (url: string) =>
             url ? this.getRelativeUrl(url) : url
         );
-
-        Marked.setOptions(this.createMarkedOptions());
-    }
-
-    /**
-     * Creates an object with options that are passed to the markdown parser.
-     *
-     * @returns The options object for the markdown parser.
-     */
-    private createMarkedOptions(): Marked.MarkedOptions {
-        const markedOptionsViaOption = this.application.options.getValue(
-            "markedOptions"
-        );
-
-        if (
-            markedOptionsViaOption != undefined &&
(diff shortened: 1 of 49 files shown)


src/lib/output/plugins/MarkedPlugin.ts
```diff
@@ -219,6 +215,34 @@ export class MarkedPlugin extends ContextAwareRendererComponent {
...
+     *
+     * @returns The options object for the markdown parser.
+     */
+    private createMarkedOptions(): Marked.MarkedOptions {
+        const markedOptions = (this.application.options.getValue(
+            "markedOptions"
+        ) ?? {}) as Marked.MarkedOptions;
+
+        if (
+            typeof markedOptions === "object" &&
```
Gerrit0, line 229: We also need a check here for `null`

Gerrit0 (changes requested):
I *really* dislike the theme tests... need to find a better solution. One small comment

src/lib/output/plugins/MarkedPlugin.ts
krisztianb: I think you missed the null coalescing a few lines above.

Gerrit0: Oops! Looks like you're right, should be able to merge tonight.

Gerrit0:
Thanks!

PR #1413, opened 2020-12-06
Enable ESLint rule "no-empty-function"

krisztianb:
Instead of disabling the rule I opted for adding a comment that makes it clear that the function intentionally does nothing.


Gerrit0:
Looks good, thanks!

PR #1451, opened 2021-01-04
ReferenceType.toString: Don't document type arguments if there are none

### LINKED ISSUE #1450: toString of reference types as type arguments include unnecessary angle brackets
## Search terms

type-arguments, angle brackets, toString

## Expected Behavior

The call of `toString` for the simple class `Animal` should return `Animal`.

## Actual Behavior

The call of `toString` for the simple class `Animal` returns `Animal<>`.

## Steps to reproduce the bug

Create a documentation for the following module:

```typescript
class Animal {
    public name: string;
}

export class TemplateTester<A, B = Animal> {
    public aaa: A;
    public bbb: B;
    public aOrb: A | B;
    public unionAB: string | A | Readonly<A | B>

    public BaseMethod(param1: A, param2: B): void {}
    public StaticMethodWithTypeParameters<A, B>(p1: A, p2: B): void {}
}

export class TemplateTesterSubClass2 extends TemplateTester<number> {
}
```

The documentation page for `TemplateTesterSubClass2` seems to be fine. But when traversing the properties and their types via my plugin, calling `toString` for properties with `Animal` result in `Animal<>` as you can see in the resulting diagram below:

class-diagram

## Environment

-   Typedoc version: 0.20.11
-   TypeScript version: 4.1.2
-   Node.js version: 12.13.0
-   OS: Windows 10

krisztianb:
Fixes #1450

### DIFF
--- a/src/lib/models/types/reference.ts
+++ b/src/lib/models/types/reference.ts
@@ -103,7 +103,8 @@ export class ReferenceType extends Type {
     toString() {
         const name = this.reflection ? this.reflection.name : this.name;
         let typeArgs = "";
-        if (this.typeArguments) {
+
+        if (this.typeArguments && this.typeArguments.length > 0) {
             typeArgs += "<";
             typeArgs += this.typeArguments
                 .map((arg) => arg.toString())


Gerrit0:
Thanks!

PR #1461, opened 2021-01-07
Update bin script to terminate gracefully

krisztianb:
As discussed here: https://gitter.im/TypeStrong/typedoc?at=5ff617b3d5f4bf2965cdd079

### DIFF
--- a/bin/typedoc
+++ b/bin/typedoc
@@ -30,7 +30,7 @@ run(app)
         console.error(error);
         return ExitCodes.ExceptionThrown;
     })
-    .then(process.exit);
+    .then((exitCode) => (process.exitCode = exitCode));
 
 /** @param {td.Application} app */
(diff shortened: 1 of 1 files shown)


Gerrit0:
Thanks!

Gerrit0:
I am reverting this change -- it previously allowed working around a lack of places to properly place async creation/cleanup work, but I believe the use case which required this is now covered by [`preRenderAsyncJobs`](https://typedoc.org/api/classes/Renderer.html#preRenderAsyncJobs) and [`postRenderAsyncJobs`](https://typedoc.org/api/classes/Renderer.html#postRenderAsyncJobs), if that doesn't cover it, some other hook needs creating

krisztianb:
Thanks for letting me know. I will have a look.

PR #1467, opened 2021-01-14
feat(category): add option to read more categories from doc

### LINKED ISSUE #1158: Support for multiple categories
I would also love to see multiple `@category` tags handled. The current implementation only keeps the first tag value around. Example:
```
    /**
     * @category Foo
     * @category Bar
     */
```
```
  "tags": [
    {
      "tag": "category",
      "text": "Foo"
    },
    {
      "tag": "category",
      "text": "Bar\n"
    }
  ]
```
becomes:
```
  "categories": [
    {
      "title": "Foo",
      "children": [
        371
      ]
    }
  ]
```
_Originally posted by @mootari in https://github.com/TypeStrong/typedoc/issues/564#issuecomment-359137258_

brunozoric:
Closes #1158 

Added a possibility to put more `@category` in the doc.

It is extracting the categories but the output is wrong. I added `AnotherTest` category to `TestClass` `publicMethod`, `protectedMethod` and `privateMethod` and changed the `spec.json` and `specs-with-lump-categories.json` according to my understanding what the output should be.
It always outputs the method to a category if its the first category in the comments. But for the second category (and the consecutive ones) it outputs a method only if its the first time output of the category.
Here are classes I was testing with:
```ts
/**
 * @category MyAnotherCategory
 * @category MyCategory
 */
export class MyClass {
    /**
     * @category MyMethodCategory
     * @category MyAnotherMethodCategory
     */
    public someMethod() {
        return 1;
    }
}
/**
 * @category MyCategory
 * @category MyAnotherCategory
 */
export class MyAnotherClass {
    public constructor() {
    }
}
/**
 * @category MyAnotherCategory
 * @category MyCategory
 */
export class MyThirdClass {
    public constructor() {
    }
}
/**
 * @category MyCategory
 * @category MyAnotherCategory
 */
export class MyFourthClass {
    public constructor() {
    }
}
```

And here is the output:
Screenshot 2021-01-14 at 10 25 02

@Gerrit0 Could you point me to where can I fix the output? I was searching through but could not find anything.

### DIFF
--- a/src/lib/converter/plugins/CategoryPlugin.ts
+++ b/src/lib/converter/plugins/CategoryPlugin.ts
@@ -161,7 +161,7 @@ export class CategoryPlugin extends ConverterComponent {
                 let category = categories.find((cat) => cat.title === childCat);
                 if (category) {
                     category.children.push(child);
-                    return;
+                    continue;
                 }
                 category = new ReflectionCategory(childCat);
                 category.children.push(child);
--- a/src/lib/converter/plugins/CategoryPlugin.ts
+++ b/src/lib/converter/plugins/CategoryPlugin.ts
@@ -187,7 +187,11 @@ export class CategoryPlugin extends ConverterComponent {
                     commentTags.push(tag);
                     return;
                 }
-                categories.push(tag.text.trim());
+                const text = tag.text.trim();
+                if (!text) {
+                    return;
+                }
+                categories.push(text);
             });
             comment.tags = commentTags;
             return categories;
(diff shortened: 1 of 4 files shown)


src/lib/converter/plugins/CategoryPlugin.ts
```diff
@@ -156,14 +157,16 @@ export class CategoryPlugin extends ConverterComponent {
                 defaultCat.children.push(child);
                 return;
             }
-            let category = categories.find((cat) => cat.title === childCat);
-            if (category) {
+            for (const childCat of childCategories) {
+                let category = categories.find((cat) => cat.title === childCat);
+                if (category) {
+                    category.children.push(child);
+                    return;
```
Gerrit0, line 164: I suspect here's your bug - you meant to `continue` instead of `return`

brunozoric: Damn! I missed that :/ Fixed it now.

Gerrit0 (approved):
Looks great! Thanks for the contribution!

Issue #304, opened 2016-10-11
Support @copydoc or enhance @inheritdoc


samchon:
I want some features to be added.
## @copydoc

Copy definition from another one.

If I need to define same content on another object, then it's very annoying. Look at the below codes, then you may understand how I am suffering from duplicated definitions
- [ParallelSystemArray](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelSystemArray.ts#L16-L70)
  - [ParallelServerArray](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelServerArray.ts#L17-L63)
  - [ParallelClientArray](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelClientArray.ts#L17-L63)
  - [ParallelServerClientArray](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelServerClientArray.ts#L22-L68)
- [ParallelSystemArrayMediator](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelSystemArrayMediator.ts#L10-L84)
  - [ParallelServerArrayMediator](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelServerArrayMediator.ts#L17-L92)
  - [ParallelClientArrayMediator](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelClientArrayMediator.ts#L18-L84)
  - [ParallelServerClientArrayMediator](https://github.com/samchon/framework/blob/master/ts/src/samchon/templates/parallel/ParallelServerClientArrayMediator.ts#L22-L97)
## @inheritdoc

The **@inheritdoc** feature in TypeDoc copies definition from its parent class. However, it's standalone. It doesn't allow using the **@inheritdoc** as subsidiary one.

Overriding a method means some logics are changed and some explanation is required. I want to combine **@inheritdoc** with customized comments explaining what is changed.

``` typescript
/**
 * This is the parent class.
 * 
 * Blah blah blah...
 */
class ParentClass
{
    /**
     * Do something.
     */
    public domSomething(): void
    {
    }
}

/**
 * This is a child class derived from the {@link ParentClass}
 *
 * #### [Inherited] {@link ParentClass}
 * @copy8doc ParentClass
 */
class ChildClass
{
    /**
[...]

Issue #494, opened 2017-05-03
Comment splitting leads to problems


jonchardy:
As it stands, TypeDoc splits comments based on the first empty line within the docstring:
```
/**
 * Here's some documentation.
 *
 * Here's more detail.
 */
```
On our output, we get some useful splitting.
- ` shortText`: "Here's some documentation."
- `text`: "Here's more detail."

Unfortunately, this can cause serious headaches when it's not wanted. Take this example:
```
/**
 * Here's some documentation with inline code:
 * ```js
 * var a = 0;
 *
 * console.log(a);
 * ```
 */
```
Now we end up with something much worse.
- `shortText`: "Here's some documentation with inline code:\n```js\nvar a = 0;"
- `text`: "console.log(a);\n```"

Of course, the parser would have to be way smarter to handle cases like this where the code markdown spans multiple lines, and I don't know how feasible that is. Maybe there could be an option where the `text` field would always contain the full comment text. That would make custom theming much simpler for those of us who want to handle things differently. Right now, it's hard to concatenate the comments since the lines end up trimmed.

I'm open for helping with this if someone can point me to an example where the JSON output is built differently based on some command-line option.

PR #1468, opened 2021-01-14
feat(inheritDoc): Add support for copying item’s documentation by copying it from another API item

Dergash:
Hello! Thanks for your work, I really appreciate it.

Recently our team started to use Typedoc and we've come to the conclusion that we really do miss the support for `@inheritDoc`, which I believe is not currently implemented (see open issue #304) when we talk about inheriting from **unrelated** entities which is allowed in TSDoc, so I hope I can do something to make this feature a bit close.

I've prepared a really quick PoC and a demo 
```ts
/**
  * Original description
  *
  * @param arg1 Numeric argument 1
  * @param arg2 Numeric argument 2
  * @returns Stringified sum of arg1 and arg2
  */
export function sourceFunction(arg1: number, arg2: number): string {
    return `${arg1 + arg2}`
}

/**
 * @inheritDoc sourceFunction
 * @param argOne Numeric argument One
 * @param argTwo Numeric argument Two
 * @param argThree Numeric argument Three
 * @deprecated TODO: Remove
 */
export function targetFunction(argOne: number, argTwo: number, argThree: number): string {
    return `${argOne + argTwo + argThree}`
}
```
Demo for this snippet is here: https://dergash.github.io/typedoc/modules.html#targetfunction

When preparing this PoC I've come to some questions which I hope we can discuss before going further.

So this draft is majorly based around [ImplementsPlugin](https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/plugins/ImplementsPlugin.ts) which I believe contains initial but undocumented support for [TSDoc/inheritDoc](https://tsdoc.org/pages/tags/inheritdoc/).  
My understanding is that when some class implements an interface, methods from that interface can inherit some parts of documentation (shortText, text, `@returns` and  any tags which TypeDocs leaves without processing), to enable this feature `inheritDoc` must be specified:

```ts
export interface SomeInterface {
  /**
   * Some method
   *
   * @param someArg1 Some argument
   * @returns Some results
   */
  someMethod(someArg1: number): string
}

export class SomeInterfaceImplementation implements SomeInterface {
  /**
   * @inheritDoc
   */
  someMethod(someArg2: number) {
    return String(someArg2)
  }
}
```

Apparently [it works](https://dergash.github.io/typedoc/classes/someinterfaceimplementation.html#somemethod). Docs will not be inherited without the tag.

So thinking about the differences between this implementation and the one expected in #304 / TSDoc:

### Missing curly brackets
Apparently TSDoc treats `inhericDoc` as an `inline tag`, hence they [wrap it in curly bracers](https://github.com/microsoft/tsdoc/pull/76/files#diff-9ffacd4b38ec593eecd8cacce59be5405f282f0f7a63231016610f390d54b87aR109): 
```ts
// https://tsdoc.org/pages/tags/inheritdoc/
/**
* {@inheritDoc example-library#Serializer.writeFile}
* @deprecated Use {@link example-library#Serializer.writeFile} instead.
*/
public save(): void {
    . . .
}
```
`ImplementsPlugin` (and `InheritDocPlugin` from my PoC) are expecting no curly braces; the only tags moment stored in [MarkedLinksPlugin](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/plugins/MarkedLinksPlugin.ts#L16) and probably MarkedPlugin

### What gets copied
According to TSDoc page following things are copied:

* summary section
* `@remarks` block
* `@params` blocks
* `@typeParam` blocks
* `@returns` block

`summary` and `@remarks` confused me a bit; after reading both TSDoc and Typedoc docs I believe TSDoc `summary` equals to Typedoc [shortText](https://github.com/TypeStrong/typedoc/blob/84c649619b43dc8092a24c6c39f34e57d9de60fd/src/lib/models/comments/comment.ts#L15), but I might be wrong: Typedoc emphasize on shortText "being the first paragraph of text" and the `text` is the rest of the comment, while from TSDoc description and [API Exractor Overview](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/) I've got the impression that summary is all text until the first block tag.  
So it probably works like this: Typedoc takes first paragraph into `shortText` and the rest of the text to `text`, when TSDoc tooks short text to `summary` and rest of the text shoul be in `remarks` tag.

So at the moment either `summary` and `@remarks` are already copied as per TSDoc, or they don't and can't be copied due to inconsistency in comment structure.

`@params` and `@typeParams` are not copied, but according to the TSDoc the should be.

`@returns` is copied.

Custom tags which are not processed by Typedoc are copied.

Should this be implemented both for inheritDoc and implements plugin?

### How exactly copying should be implemented?

At the moment inherited values override matching values on the target entity. Which seems consistent with [JSDoc](https://jsdoc.app/tags-inheritdoc.html) where `@inheritDoc` leads to ignoring all other tags. From TSDoc site and github I didn't quite get if it's override or merge. But if it is indeed override then `@inheritDoc` won't work for #304 where there is a need to inherit a comment by concating it with existing comment.

If this is the case, will you consider support for `@copyDoc`?

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -209,7 +209,8 @@ export function parseComment(
         if (
             tagName === "param" ||
             tagName === "typeparam" ||
-            tagName === "template"
+            tagName === "template" ||
+            tagName === "inheritdoc"
         ) {
             line = consumeTypeData(line);
             const param = /[^\s]+/.exec(line);
--- a/src/lib/converter/plugins/ImplementsPlugin.ts
+++ b/src/lib/converter/plugins/ImplementsPlugin.ts
@@ -1,5 +1,4 @@
 import {
-    Reflection,
     ReflectionKind,
     DeclarationReflection,
     SignatureReflection,
@@ -8,8 +7,8 @@ import { Type, ReferenceType } from "../../models/types/index";
 import { Component, ConverterComponent } from "../components";
 import { Converter } from "../converter";
 import { Context } from "../context";
-import { Comment } from "../../models/comments/comment";
 import { zip } from "../../utils/array";
+import { copyComment } from "../utils/reflections";
 
 /**
  * A plugin that detects interface implementations of functions and
@@ -82,7 +81,7 @@ export class ImplementsPlugin extends ConverterComponent {
                     interfaceMember,
                     context.project
                 );
-                this.copyComment(classMember, interfaceMember);
+                copyComment(classMember, interfaceMember);
 
                 if (
                     interfaceMember.kindOf(ReflectionKind.FunctionOrMethod) &&
@@ -105,7 +104,7 @@ export class ImplementsPlugin extends ConverterComponent {
                                             interfaceSignature,
                                             context.project
                                         );
-                                        this.copyComment(
+                                        copyComment(
                                             classSignature,
                                             interfaceSignature
                                         );
@@ -117,46 +116,6 @@ export class ImplementsPlugin extends ConverterComponent {
                 }
             }
         );
-    }
-
-    /**
-     * Copy the comment of the source reflection to the target reflection.
-     *
-     * @param target
-     * @param source
-     */
-    private copyComment(target: Reflection, source: Reflection) {
-        if (
-            target.comment &&
-            source.comment &&
-            target.comment.hasTag("inheritdoc")
-        ) {
-            target.comment.copyFrom(source.comment);
-
-            if (
-                target instanceof SignatureReflection &&
-                target.parameters &&
-                source instanceof SignatureReflection &&
-                source.parameters
-            ) {
-                for (
-                    let index = 0, count = target.parameters.length;
-                    index < count;
-                    index++
-                ) {
-                    const sourceParameter = source.parameters[index];
-                    if (sourceParameter && sourceParameter.comment) {
-                        const targetParameter = target.parameters[index];
-                        if (!targetParameter.comment) {
-                            targetParameter.comment = new Comment();
-                            targetParameter.comment.copyFrom(
-                                sourceParameter.comment
-                            );
-                        }
-                    }
-                }
-            }
-        }
     }
 
     private analyzeInheritance(
@@ -201,7 +160,7 @@ export class ImplementsPlugin extends ConverterComponent {
                         parentMember,
                         context.project
                     );
-                    this.copyComment(child, parentMember);
+                    copyComment(child, parentMember);
                 }
             }
         }
--- a/src/lib/converter/plugins/InheritDocPlugin.ts
+++ b/src/lib/converter/plugins/InheritDocPlugin.ts
@@ -1,0 +1,88 @@
+import {
+    ContainerReflection,
+    DeclarationReflection,
+    ReflectionKind,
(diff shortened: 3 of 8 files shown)


Gerrit0:
Thanks for the PR!

Your analysis is right - TypeDoc's behavior doesn't match TSDoc's, which is somewhat expected since it predates the TSDoc specification by several years. The goal is for TypeDoc to move towards aligning with the TSDoc standard, so any improvements in that area are welcomed!

> Missing curly brackets

I never really understood why inheritdoc was an inline tag... For now, the POC you provided is consistent with other TypeDoc tags, so I like it as is. TypeDoc doesn't really have support for inline tags currently (`{@link}` is rather hackily supported), I've been putting off adding support until I have time to spend some real time getting significantly closer to TSDoc.

> `summary` and `@remarks`

This is another difference between TSDoc & TypeDoc - TSDoc's `summary` would be `${comment.shortText}\n\n${comment.text}`.  This behavior does need to be revisited as it can cause issues (#494), so I'd say we already copy `summary`, but not `@remarks` since TypeDoc doesn't have special handling for it.

> Should this be implemented both for inheritDoc and implements plugin?

I think - yes. The described behavior makes sense to me.

> override or merge

Tossing this into the TSDoc playground: https://tsdoc.org/play

```ts
/**
 * summary
 * @remarks X
 * {@inheritdoc x}
 */
```

Shows that TSDoc is even stricter - not an override, but an error if summary or remarks is specified (but not for `@param`, `@returns`, `@typeParam`). This is kind of weird; I would have expected it to copy if not present in the inheriting comment... For parity with TSDoc, we probably shouldn't extend this behavior, and instead use `@copyDoc` for providing merge support.

Dergash:
[Extended demo here ](https://dergash.github.io/typedoc/modules/inheritance.html)
[Demo source code here](https://github.com/Dergash/typedoc/commit/d85f3bf35a28d47f9507962ee3e9909358f19e1d)

Some details on implementation:

1) `copyComment` is extracted from `ImplementsPlugin` to `utils/reflections`

2) `copyComment` is modified to conform with TSDoc standard of replacing @param block instead of merging; to avoid breaking existing behaviour of `@inheritDoc` the current implementation is moved to `legacyCopyImplementation` function and is called when `@inheritDoc` is used without parameter. 

3) `copyFrom` now only copies tags specified in TSDoc: `@remarks` and `@returns`, and leaves other tags not used by Typedoc without change

4) `InheritDocPlugin` introduced to traverse through reflections with `@inheritDoc` and call `copyComment`. There is a white spot there concerning matching methods signatures: `InheritDocPlugin` relies on [isTypeListEqual](https://github.com/TypeStrong/typedoc/blob/00ae3c759033ec389b594044ef21991bac88ab5d/src/lib/models/types/abstract.ts#L66)to check templated arguments and it works for entities with established inheritance relationship (extends, implements). When referencing unrelated entity there is no obvious way to know that `source<T>(arg: T)` matches `target<T>(arg: T)`, so there is a fallback to first available which probably not very correct.

If it works I can move to @copyDos as per initial issue #304

Gerrit0:
At first glance, this looks great! I will need to look at it in more depth this weekend.

Gerrit0 (approved):
This is great! Thanks :)

PR #1482, opened 2021-01-24
feat: Support for --watch, --preserveWatchOutput, --emit

### LINKED ISSUE #403: #FeatureRequest Support Watch mode
When working on a complex codebase, it would be useful to have a watch mode, so you would be able to see changes faster.

Gerrit0:
Closes #403

### DIFF
--- a/bin/typedoc
+++ b/bin/typedoc
@@ -58,6 +58,24 @@ async function run(app) {
         return ExitCodes.NoEntryPoints;
     }
 
+    if (app.options.getValue("watch")) {
+        app.convertAndWatch(async (project) => {
+            const out = app.options.getValue("out");
+            if (out) {
+                await app.generateDocs(project, out);
+            }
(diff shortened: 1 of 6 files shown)


cspotcode:
Looks good, though I'm no typedoc expert.  Does `--emit` emit all the same things as `tsc`?  That's the only bit that's confusing to me.  Is the goal to allow running `typedoc` instead of `tsc`?  Or is the goal to emit `tsbuildinfo` files by simply allowing all compiler emit to reach the filesystem?

cspotcode:
Also, thanks for implementing this so quickly.

Gerrit0:
Yes, `--emit` makes it possible to use typedoc instead of tsc to build. Once `--watch` was added, it made sense to add it so that people running typedoc in watch mode wouldn't also need to run tsc in watch to build their code.

PR #1486, opened 2021-01-26
feat: add highlight theme option

refineddigital:
adds support for changing the default code highlighter theme

### DIFF
--- a/src/lib/utils/highlighter.ts
+++ b/src/lib/utils/highlighter.ts
@@ -1,6 +1,6 @@
 import { ok as assert } from "assert";
 import * as shiki from "shiki";
-import { Theme } from "shiki-themes";
+import { Theme as ShikiTheme } from "shiki-themes";
 import { Highlighter } from "shiki/dist/highlighter";
 import { unique } from "./array";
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/highlighter.ts
```diff
@@ -22,10 +23,10 @@ const supportedLanguages = unique([
 
 let highlighter: Highlighter | undefined;
 
-export async function loadHighlighter() {
+export async function loadHighlighter(theme: Theme) {
     if (highlighter) return;
     highlighter = await shiki.getHighlighter({
-        theme: "light-plus",
+        theme: theme || "light-plus",
```
Gerrit0, line 29: I'd rather specify this as the default in the option declaration - then it can be picked up and shown in the typedoc.json schema file.

src/lib/utils/options/sources/typedoc.ts
```diff
@@ -243,4 +244,9 @@ export function addTypeDocOptions(options: Pick<Options, "addDeclaration">) {
             "Specify the options passed to Marked, the Markdown parser used by TypeDoc",
         type: ParameterType.Mixed,
     });
+    options.addDeclaration({
+        name: "highlightTheme",
+        help: "Specifies the code highlighting theme.",
+        type: ParameterType.String,
+    });
```
Gerrit0, line 251: Can you add a validation function to this to ensure that the user provided is valid? I believe Shiki exports an array of themes.

Gerrit0 (changes requested):
Looking good, just a couple comments :)

refineddigital:
@Gerrit0 ok, I think I have now done as you have requested.

Gerrit0:
Looks great, thanks! Will be published either today or tomorrow :)

refineddigital:
You are a champion, thanks @Gerrit0

PR #1507, opened 2021-02-10
Add support for non .com gh enterprise domains

TUNER88:
We have an instance of GitHub Enterprise on domain non `.com` domain (`github.*.*.tools`) so file links could not be generated due to fixed `.com` TLD in the regex pattern.

This PR will fix this fixed TLD limitation.


Gerrit0:
Thanks!

PR #1512, opened 2021-02-20
Preserve intendation in code blocks

MichaelFroeschen:
The generated code blocks do not have any intendation which makes it hard to read.
With this change, the intendation is preserved in the case that the code block is defined with `


Gerrit0:
Thanks!

Gerrit0:
Published in 0.20.27

PR #1534, opened 2021-03-13
fix: readonly tuples were recognized as arrays

Minhir:
Readonly tuples as method arguments were wrongly recognized as arrays. 

So `get(arg: readonly [number, string])` was converted to `readonly number[]`. I fixed it. I hope someone can help to find more elegant way to handle this.


Gerrit0:
Thanks for the PR! I think a cleaner fix will to be to resolve the type reference and check the resolved type rather than relying on the internal method to do that.

Issue #1557, opened 2021-03-29
Request : Add file path and line number to make MarkedPlugin error messages more friendly.


MasatoMakino:
## Search Terms

error message, getHighlighted, markedOptions, highlight

## Problem

Currently, the error messages of the `getHighlighted` function of `MarkedPlugin.ts` do not show the file path and line number of the file where the error occurred. So users need a lot of time to identify the cause of the error message.

## Suggested Solution

Currently, `getHighlighted` generates the following error messages

https://github.com/TypeStrong/typedoc/blob/315bd2501360c9e3738c512a850f37cf314b8627/src/lib/output/plugins/MarkedPlugin.ts#L109-L121

```
Unsupported highlight language "{language type}" will not be highlighted. Run typedoc --help for a list of supported languages.
```

The following information will be added to this message.

```
This code block is in line n of "path/to/file".
```

Gerrit0:
This is a good idea 👍

Gerrit0:
Resolved by #1561

PR #1561, opened 2021-04-02
Add output file path and source file path to MarkedPlugin error message

MasatoMakino:
related : #1557 

In this PR, the error message does not show the line number. If there are code blocks in the included external files, it is not possible to get the correct number of lines. Instead of the line number, I have added information about target code block and output file path.

The error messages are formatted with tabs. I have searched in the repository but could not find any example of error message formatting.


Gerrit0:
Thanks!

This isn't a perfect fix, since the `sources` property won't be set if someone uses the `--disableSources` option, but it's better than what we have today.

PR #1565, opened 2021-04-07
Create `pretty` option

cAttte:
this PR adds a `pretty` option, to specify whether the output of `--json` should be formatted/prettified or not (e.g., current output vs. `{"x":8}`). it defaults to `true` as to match the current behavior and not break existing setups (not that it would matter much anyway, but).

this is *kind of* useful as, usually, JSON output doesn't need to be human-readable, and if not formatted, it can save some bandwidth (i.e. if you're serving the JSON through a server) and storage, but it probably makes no difference with small projects. i tested it with a Very Big interface; the pretty JSON was ~868 KB, and the ugly JSON was ~317 KB.

i'm pretty sure i covered everything and this is a very tiny change, but let me know if i missed anything! thanks for this project, btw :)

### DIFF
--- a/bin/typedoc
+++ b/bin/typedoc
@@ -66,8 +66,7 @@ async function run(app) {
             }
             const json = app.options.getValue("json");
             if (json) {
-                const pretty = app.options.getValue("pretty");
-                await app.generateJson(project, json, pretty);
+                await app.generateJson(project, json);
             }
 
             if (!out && !json) {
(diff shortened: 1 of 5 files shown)


Gerrit0:
I like it! Would just like one change - I'd rather have the `generateJson` method use that option directly instead of requiring all the callers of it to update to use the third argument if they want to respect the option.

cAttte:
sure! yeah, that makes more sense

Gerrit0:
Thanks!

PR #1576, opened 2021-04-23
feat: Add disableAliases option

### LINKED ISSUE #1571: Exports in ambient module don't get included in certain case
## Search terms

ambient, module, export, exports, import, imports, not, included

## Expected Behavior

Shared types imported from other ambient module should be included in both ambient modules.

## Actual Behavior

Shared types imported from other ambient module only get included in 1st ambient module.

(Types that are missing from `alt-server` are marked with red dot on the right)
(Note that also functions are missing, but they're not visible here)

Excerpt

## Steps to reproduce the bug

Clone this repo with specified branch and generate documentation by executing `./docs/build.ps1` script (only Windows is supported for now):
https://github.com/altmp/altv-types/tree/shared-package
You should be able to check documentation on `localhost:8080` or simply check `./docs/api/.manifest`, which contains renamed typedoc JSON output. (In case of problems, simply run `npm i` and `npx typedoc` inside `./docs/` directory)

## Environment

-   Typedoc version: 0.20.35
-   TypeScript version: 4.2.4
-   Node.js version: 14.15.1
-   OS: Windows 10 Pro v1809 x64

### LINKED ISSUE #1574: Including non exported function in generated Documentation
I am facing an issue to include a non exported function to appear in the generated documentation. I am fairly new to TypeScript (and Typedoc as well) it still confuses me what works and what does not, and some online tutorials are no longer relevant.

`import HubRepository from '../../database/repositori...;`
`import HubGroupUserRepository from '../../database/repositori...;`

`/**`
 `* Raw handler to handle approving assets`
` * @param event`
 `* @returns Returns message to inform the user if an asset has been approved`
 `* @notExported`
` */`
 `const rawHandler = async (`
   `event: APIGatewayEvent<DefaultEvent>,`
`): Promise<APIGatewayResponse<AssetResponse | IdResponse>> => {`
    `const { sub: cognitoId } = event.requestContext.authorizer.claims; // Defines cognitoId > APIGatewayProxyCognitoAuthorizer`
`(name) - node_modules\@types\aws-lambda\trigger\api-gateway-proxy.d.ts`
    `const { hubId, assetId } = event.pathParameters;`
`...`

At the bottom of the file:

`export const handler = middy(rawHandler)`
    `.use(jsonBodyParser())`
`...`

In the generated documentation, only the "handler" variable appears, but I would like it to include the "rawHandler" variable as well. As you might notice, I already tried using "@notExported" with the additional plugin, but it still does not work.

Also, is there a possibility to include inline comments in the documentation?

Thank you!

Lhoerion:
Resolves #1571

### DIFF
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -53,9 +53,6 @@ export class Converter extends ChildableComponent<
     @BindOption("excludeProtected")
     excludeProtected!: boolean;
 
-    @BindOption("disableAliases")
-    disableAliases!: boolean;
-
     /**
      * General events
      */
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -53,6 +53,9 @@ export class Converter extends ChildableComponent<
     @BindOption("excludeProtected")
     excludeProtected!: boolean;
 
+    @BindOption("disableAliases")
+    disableAliases!: boolean;
+
     /**
      * General events
      */
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -713,7 +713,6 @@ function convertAlias(
             exportSymbol ?? symbol
         );
     } else if (
-        !reflection ||
         !context.converter.application.options.getValue("disableAliases")
     ) {
         createAlias(reflection, context, symbol, exportSymbol);
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -705,16 +705,17 @@ function convertAlias(
     const reflection = context.project.getReflectionFromSymbol(
         context.resolveAliasedSymbol(symbol)
     );
-    if (!reflection) {
+    if (
+        !reflection ||
+        context.converter.application.options.getValue("disableAliases")
+    ) {
         // We don't have this, convert it.
         convertSymbol(
             context,
             context.resolveAliasedSymbol(symbol),
             exportSymbol ?? symbol
         );
-    } else if (
-        !context.converter.application.options.getValue("disableAliases")
-    ) {
+    } else {
         createAlias(reflection, context, symbol, exportSymbol);
     }
 }
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -713,6 +713,7 @@ function convertAlias(
             exportSymbol ?? symbol
         );
     } else if (
+        !reflection ||
         !context.converter.application.options.getValue("disableAliases")
     ) {
         createAlias(reflection, context, symbol, exportSymbol);
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -73,6 +73,7 @@ export function convertSymbol(
     // but aren't aliased symbols because `export *` was used.
     const previous = context.project.getReflectionFromSymbol(symbol);
     if (
+        !context.converter.disableAliases &&
         previous &&
         previous.parent?.kindOf(ReflectionKind.Module | ReflectionKind.Project)
     ) {
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -712,7 +712,9 @@ function convertAlias(
             context.resolveAliasedSymbol(symbol),
             exportSymbol ?? symbol
         );
-    } else {
+    } else if (
+        !context.converter.application.options.getValue("disableAliases")
+    ) {
         createAlias(reflection, context, symbol, exportSymbol);
     }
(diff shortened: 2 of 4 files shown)


src/lib/converter/converter.ts
```diff
@@ -53,6 +53,9 @@ export class Converter extends ChildableComponent<
     @BindOption("excludeProtected")
     excludeProtected!: boolean;
 
+    @BindOption("disableAliases")
+    disableAliases!: boolean;
```
Gerrit0, line 57: Either mark this as `@internal` or reach into options directly

src/lib/converter/symbols.ts
```diff
@@ -73,6 +73,7 @@ export function convertSymbol(
     // but aren't aliased symbols because `export *` was used.
     const previous = context.project.getReflectionFromSymbol(symbol);
     if (
+        !context.converter.disableAliases &&
```
Gerrit0, line 76: This gets half of it, there's also the convertAlias function on line 700

Gerrit0 (changes requested):

Lhoerion:
First post content was changed due to wrong issue linked

src/lib/converter/symbols.ts
```diff
@@ -711,7 +712,9 @@ function convertAlias(
             context.resolveAliasedSymbol(symbol),
             exportSymbol ?? symbol
         );
-    } else {
+    } else if (
```
Gerrit0, line 715: This isn't the right check here - it should be `if (!reflection || disableAliases)` since otherwise the reflection will be missing.

Gerrit0 (changes requested):

src/lib/converter/symbols.ts
Lhoerion: True, I forgot about that

Gerrit0: Sorry, that's not what I meant.... that should have been added to the if block on line 708, not down here. This should still just be an else.

Lhoerion: Ok, now I've clearer picture of this function's purpose.

Although in which cases reflection might be null while converting alias?

Gerrit0: The reflection will be undefined if the symbol that the alias is pointing to hasn't been converted yet

Gerrit0 (approved):
Looks good!

src/lib/converter/symbols.ts
Lhoerion: Thanks for pointing me in the right direction :smile:

Lhoerion:
@Gerrit0 I came back to this after a while and I've noticed strange extendedTypes behaviour.
While extended types reflections are properly resolved for a class (in my case alt-client.WorldObject), in alt-server.WorldObject they point still on first occurence in whole project (alt-client.WorldObject).
I tried to find where's the problem hiding, but so far I couldn't find logic behind assigning reflection ids to types created from classes heritageClauses.

Gerrit0:
Yeah, that doesn't surprise me. References are created with types having all their links resolved, so you'll end up with multiple reflections with the same symbol. The ImplementsPlugin relies on getting the symbol to perform inheritance analysis, and that hasn't been updated to support Symbol -> Reflection[] (which really needs to happen due to declaration merging anyways, but I haven't figured out a clean way to do it, and it doesn't break too many people)

Lhoerion:
> Yeah, that doesn't surprise me. References are created with types having all their links resolved, so you'll end up with multiple reflections with the same symbol. The ImplementsPlugin relies on getting the symbol to perform inheritance analysis, and that hasn't been updated to support Symbol -> Reflection[] (which really needs to happen due to declaration merging anyways, but I haven't figured out a clean way to do it, and it doesn't break too many people)

Is there an issue created for this? If not, would be nice to have one.

Lhoerion:
Also on closer look, it seems that every reference from "shared" module is pointing to incorrect symbol in 2nd module, so it's even more vast that I originally assumed.

Gerrit0:
I don't think there's a canonical issue for it, there's probably some discussion in the 0.20 tracking thread, but I honestly don't remember... and yep, I kind of expected that for the latter as well.

Come to think of it, `--packages` might be a perfect fit for your repo since it will make TD create two programs and therefore the types will be completely separate... though whether that's a limitation of TD or a feature is kind of undefined right now.

Lhoerion:
In such case I think that we need to either properly document what happens if disableAliases is used or just entirely remove it.

Gerrit0:
Since it hasn't been published in a non-beta version (happy accident), I'm fine with just removing it - it does reveal more obviously problems with the architecture

PR #1596, opened 2021-06-09
Add support for TS entry points in monorepo configuration

Lhoerion:
Firstly, thank you @efokschaner and @Gerrit0 for implementing packages entry points 😃

I've added it as we talked about it in #1567, for now TS files must have extension set explicitly, otherwise they won't be "fetched".

If it's fine as-is, it would be nice to get it released in beta channel ASAP.

### DIFF
--- a/src/lib/utils/package-manifest.ts
+++ b/src/lib/utils/package-manifest.ts
@@ -241,7 +241,7 @@ export function getTsEntryPointForPackage(
         if (e.code !== "MODULE_NOT_FOUND") {
             throw e;
         } else {
-            let tsEntryPointPath = resolve(
+            const tsEntryPointPath = resolve(
                 packageJsonPath,
                 "..",
                 packageTypes ?? packageMain
--- a/src/lib/utils/package-manifest.ts
+++ b/src/lib/utils/package-manifest.ts
@@ -218,40 +218,27 @@ export function getTsEntryPointForPackage(
     ) {
         packageTypes = packageJson.typings;
     }
-    let jsEntryPointPath = resolve(packageJsonPath, "..", packageMain);
-    // In opposition to .js files, .ts(x) extension must be explicitly declared.
-    if (/\.tsx?$/.test(jsEntryPointPath)) {
-        if (existsSync(jsEntryPointPath)) {
-            return jsEntryPointPath;
-        } else {
-            logger.warn(
-                `Could not determine the TS entry point for "${packageJsonPath}". Package will be ignored.`
-            );
-            return ignorePackage;
-        }
-    }
-
-    // The jsEntryPointPath from the package manifest can be like a require path.
+    let entryPointPath = resolve(packageJsonPath, "..", packageMain);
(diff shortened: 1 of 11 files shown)


src/lib/utils/package-manifest.ts
```diff
@@ -205,8 +206,25 @@ export function getTsEntryPointForPackage(
         typeof packageJson.main == "string"
     ) {
         packageMain = packageJson.main;
+    } else if (
+        hasOwnProperty(packageJson, "types") &&
+        typeof packageJson.types == "string"
+    ) {
+        packageMain = packageJson.types;
```
efokschaner, line 213: Should we first allow the default value of index.js to be tested for existence before we resort to the types field?
I could imagine a package which had an index.ts, an index.js, and index.d.ts, and therefore doesn't specify package.main, 

With this change typedoc would now parse index.d.ts for docs rather than index.ts, while it's probably preferable to parse index.ts.

Essentially I'm proposing we only check the types field if we hit the condition in line 220 in the old implementation (although there may be another edge case or intent weirdness I'm missing).

efokschaner:
Thanks for contributing!

src/lib/utils/package-manifest.ts
Lhoerion: Yes, I've changed that behaviour while having something else on my mind:

On the other hand, usually when `main` & `types` are specified, it means that dev has specified a separate file created for storing his types definitions (case when your project is written in JS, but you've added TS types only to not rewrite whole project in TS), so logically we should check out `types`/`typings` first. Previously we had `includeDeclarations` flag for controlling whenever you wanted to include `.d.ts` files in documenting process, but AFAIK it's gone since v0.20.

But for now I'll make it follow behaviour specified in README.

Lhoerion:
__CONTINUATION OF DISCUSSION__

> Should we first allow the default value of index.js to be tested for existence before we resort to the types field?
I could imagine a package which had an index.ts, an index.js, and index.d.ts, and therefore doesn't specify package.main,

>With this change typedoc would now parse index.d.ts for docs rather than index.ts, while it's probably preferable to parse index.ts.

On the other hand, usually when main & types are specified, it means that dev has specified a separate file created for storing his types definitions (case when your project is written in JS, but you've added TS types only to not rewrite whole project in TS), so logically we should check out types/typings first. Previously we had includeDeclarations flag for controlling whenever you wanted to include .d.ts files in documenting process, but AFAIK it's gone since v0.20 (which would be helpful to disable/enable behaviour described above).

efokschaner:
When main is specified (or the default works) we use the source mapping in the js to find the original ts. If we can't find the .ts I agree with checking for types (the .d.ts).
We should prefer finding the .ts though as it's more likely to contain the kind of tsdoc markup + other annotations that power typedoc, not to mention creating links to the .ts rather than the .d.ts in typedoc output.

So, in short, I agree with using the .d.ts when there is no .ts, as long as we try to find that .ts first.

Gerrit0:
This seems reasonable to me, thanks!

PR #1650, opened 2021-08-10
Add support for NO_COLOR environment variable

### LINKED ISSUE #1619: Add support for NO_COLOR environment variable
## Search Terms
Color escapes

## Problem
Today, TypeDoc includes ANSI colors for colorizing output. Most of the time, this is fine, but some tools don't handle it well.

## Suggested Solution
Check for NO_COLOR and don't output color codes if it is set - https://no-color.org/
tsc is adding support for NO_COLOR in an upcoming release, we should support it too.

krisztianb:
Closes #1619

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -222,6 +222,10 @@ export class Application extends ChildableComponent<
         super(DUMMY_APPLICATION_OWNER);
 
         this.logger = new ConsoleLogger();
+        if (this.logger instanceof ConsoleLogger && "NO_COLOR" in process.env) {
+            this.logger.usesColors = false;
+        }
+
         this.options = new Options(this.logger);
         this.options.addDefaultDeclarations();
         this.serializer = new Serializer();
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -222,10 +222,6 @@ export class Application extends ChildableComponent<
         super(DUMMY_APPLICATION_OWNER);
 
         this.logger = new ConsoleLogger();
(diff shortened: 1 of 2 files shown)


Gerrit0:
I'd like to avoid introducing new properties, instead just adding the process.env check to the logger file when creating prefixes. Introducing a new property increases the API surface with no immediate benefit that I see

krisztianb:
Are you ok with making the property private? I think it makes sense not to query the environment variables each time a message is logged.

Gerrit0:
Yeah, I'm good with this.

I was thinking of doing:

```ts
const messagePrefixes = "NO_COLOR" in process.env ? { ... } : { ... }
```

happy5214:
I'm not sure if anyone will see this, but with this feature now supported, can someone add TypeDoc to the list at https://github.com/jcs/no_color? I looked there to see if TypeDoc had `NO_COLOR` support, and I ended up having to dig through the code to figure that out.

phoneticallySAARTHaK:
@happy5214 I believe you can do that yourself? They have direction for the same 

https://github.com/jcs/no_color?tab=readme-ov-file#adding-software

happy5214:
> @happy5214 I believe you can do that yourself? They have direction for the same
> 
> https://github.com/jcs/no_color?tab=readme-ov-file#adding-software

I considered doing it myself, but as someone who just started using TypeDoc today, I didn't want to possibly misrepresent the description or presuppose that TypeDoc's omission was a simple oversight rather than an intentional decision. I just don't know. If anything, I thought I'd ask least seek the permission of a project of which I'm not a contributor or long-time user before adding it to that list.

PR #1667, opened 2021-08-27
Bump Shiki to 0.9.7, fixes #1666; resolve path-parse audit flag, fixes #1665

### LINKED ISSUE #1665: npm audit reports moderate severity dependency threat
## Search terms

Regular Expression Denial of Service in path-parse - https://npmjs.com/advisories/1773

## Expected Behavior

Install without warnings

## Actual Behavior

Installs with warnings

image

## Steps to reproduce the bug

1. `git clone` current repo
2. `npm install`

## Environment

-   Typedoc version: 0.21.7
-   TypeScript version: `4.0.x || ... || 4.4.x` (resolved to 4.4.2)
-   Node.js version: v16.5.0
-   OS: Win10 Pro 64

----

Patch inbound

### LINKED ISSUE #1666: Please bump `shiki.js` from 0.9.3 to 0.9.7
## Search Terms

None applicable

## Problem

We use [shiki.js](https://github.com/shikijs/shiki/) for highlighting.

Our install is at 0.9.3.  The current version is 0.9.7.  Shiki adopted a grammar for a language I care about (FSL) in 0.9.5

## Suggested Solution

I would like to bump Shiki please.  PR inbound

StoneCypher:
I came to bump the highlighter, because a version after this one adopts a grammar that I want.  This resolves #1666 .

In so doing, I realized that one of your deps is warning.  I bump it in the process, through `npm audit --fix`.  This resolves #1665 .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "marked": "^3.0.2",
         "minimatch": "^3.0.0",
         "progress": "^2.0.3",
-        "shiki": "^0.9.3",
+        "shiki": "^0.9.7",
         "typedoc-default-themes": "^0.12.10"
       },
       "bin": {
@@ -2125,10 +2125,9 @@
       "dev": true
     },
     "node_modules/json5": {
-      "version": "2.1.3",
-      "resolved": "https://registry.npmjs.org/json5/-/json5-2.1.3.tgz",
-      "integrity": "sha512-KXPvOm8K9IJKFM0bmdn8QXh7udDh1g/giieX0NLCaMnb4hEiVFqnop2ImTXCc5e0/oHz3LTqmHGtExn5hfMkOA==",
-      "dev": true,
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.0.tgz",
+      "integrity": "sha512-f+8cldu7X/y7RAJurMEJmdoKXGB/X550w2Nr3tTbezL6RwEE/iMcm+tZnXeoZtKuOq6ft8+CqzEkrIgx1fPoQA==",
       "dependencies": {
         "minimist": "^1.2.5"
       },
@@ -2859,9 +2858,9 @@
       }
     },
     "node_modules/path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
-      "integrity": "sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw==",
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.7.tgz",
+      "integrity": "sha512-LDJzPVEEEPR+y48z93A0Ed0yXb8pAByGWo/k5YYdYgpY2/2EsOsksJrq7lOHxryrVOn1ejG6oAp8ahvOIQD8sw==",
       "dev": true
     },
     "node_modules/path-type": {
@@ -3186,12 +3185,13 @@
       }
     },
     "node_modules/shiki": {
-      "version": "0.9.3",
-      "resolved": "https://registry.npmjs.org/shiki/-/shiki-0.9.3.tgz",
-      "integrity": "sha512-NEjg1mVbAUrzRv2eIcUt3TG7X9svX7l3n3F5/3OdFq+/BxUdmBOeKGiH4icZJBLHy354Shnj6sfBTemea2e7XA==",
+      "version": "0.9.7",
+      "resolved": "https://registry.npmjs.org/shiki/-/shiki-0.9.7.tgz",
+      "integrity": "sha512-rOoAmwRWDiGKjQ1GaSKmbp1J5CamCera+I+DMM3wG/phbwNYQPt1mrjBBZbK66v80Vl1/A9TTLgXVHMbgtOCIQ==",
       "dependencies": {
+        "json5": "^2.2.0",
         "onigasm": "^2.2.5",
-        "vscode-textmate": "^5.2.0"
+        "vscode-textmate": "5.2.0"
       }
     },
     "node_modules/signal-exit": {
@@ -5430,10 +5430,9 @@
       "dev": true
     },
     "json5": {
-      "version": "2.1.3",
-      "resolved": "https://registry.npmjs.org/json5/-/json5-2.1.3.tgz",
-      "integrity": "sha512-KXPvOm8K9IJKFM0bmdn8QXh7udDh1g/giieX0NLCaMnb4hEiVFqnop2ImTXCc5e0/oHz3LTqmHGtExn5hfMkOA==",
-      "dev": true,
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.0.tgz",
+      "integrity": "sha512-f+8cldu7X/y7RAJurMEJmdoKXGB/X550w2Nr3tTbezL6RwEE/iMcm+tZnXeoZtKuOq6ft8+CqzEkrIgx1fPoQA==",
       "requires": {
         "minimist": "^1.2.5"
       }
@@ -6008,9 +6007,9 @@
       "dev": true
     },
     "path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
-      "integrity": "sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw==",
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.7.tgz",
+      "integrity": "sha512-LDJzPVEEEPR+y48z93A0Ed0yXb8pAByGWo/k5YYdYgpY2/2EsOsksJrq7lOHxryrVOn1ejG6oAp8ahvOIQD8sw==",
       "dev": true
     },
     "path-type": {
@@ -6225,12 +6224,13 @@
       "dev": true
     },
     "shiki": {
-      "version": "0.9.3",
-      "resolved": "https://registry.npmjs.org/shiki/-/shiki-0.9.3.tgz",
-      "integrity": "sha512-NEjg1mVbAUrzRv2eIcUt3TG7X9svX7l3n3F5/3OdFq+/BxUdmBOeKGiH4icZJBLHy354Shnj6sfBTemea2e7XA==",
+      "version": "0.9.7",
+      "resolved": "https://registry.npmjs.org/shiki/-/shiki-0.9.7.tgz",
+      "integrity": "sha512-rOoAmwRWDiGKjQ1GaSKmbp1J5CamCera+I+DMM3wG/phbwNYQPt1mrjBBZbK66v80Vl1/A9TTLgXVHMbgtOCIQ==",
       "requires": {
+        "json5": "^2.2.0",
         "onigasm": "^2.2.5",
-        "vscode-textmate": "^5.2.0"
+        "vscode-textmate": "5.2.0"
       }
     },
     "signal-exit": {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "marked": "^3.0.2",
     "minimatch": "^3.0.0",
     "progress": "^2.0.3",
-    "shiki": "^0.9.3",
+    "shiki": "^0.9.7",
     "typedoc-default-themes": "^0.12.10"
   },
   "peerDependencies": {


Gerrit0:
The first error in CI is the reason I haven't upgraded Shiki yet ^

> Error: node_modules/shiki/dist/index.d.ts(269,68): error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.

This issue in Shiki has been fixed (https://github.com/shikijs/shiki/pull/203), but a new release hasn't been made yet, so we can't really upgrade without ignoring those errors (ewww...)

If you delete package-lock.json (or yarn.lock) and reinstall, because TypeDoc declares `^0.9.3` as the version, a fresh install with no lockfile should still install 0.9.7.

As for the audit warning - [npm audit is broken by design](https://overreacted.io/npm-audit-broken-by-design/). The reported "vulnerability" is coming from nyc.. which is a development dependency. Sure, there might be vulnerability here.. but in order for it to actually be a problem, you would have to intentionally set up your local environment to exploit yourself.

```
> npm why path-parse
path-parse@1.0.6 dev
node_modules/path-parse
  path-parse@"^1.0.6" from resolve@1.15.1
  node_modules/resolve
    resolve@"^1.3.2" from @babel/core@7.11.1
    node_modules/@babel/core
      @babel/core@"^7.7.5" from istanbul-lib-instrument@4.0.3
      node_modules/istanbul-lib-instrument
        istanbul-lib-instrument@"^4.0.0" from nyc@15.1.0
        node_modules/nyc
          dev nyc@"^15.1.0" from the root project
```

StoneCypher:
> This issue in Shiki has been fixed (shikijs/shiki#203)

It's much easier to just remove isolatedModules in your thing, so that you can have the `as const` expression as desired and still compile it downstream

I'm wiring up a second PR that fixes this without waiting on shiki, and which allows shiki to go back to the decidedly better `as const` statement

That PR also resolves your cast of string to `Theme`

&nbsp;

> As for the audit warning - npm audit is broken by design. The reported "vulnerability" is coming from nyc.. which is a development dependency. 

Uh huh.  Also breaks a lot of tooling.  There's no reason not to just fix it.

StoneCypher:
[New PR](https://github.com/TypeStrong/typedoc/pull/1670) should:

1. Fix the build blockage
2. Update Shiki
3. Fix the dep
4. Fix the novel typecasting requirement

Gerrit0:
> It's much easier to just remove isolatedModules in your thing

Yes, and also wrong. A dependency should not dictate that compiler options must be set to more lax values. isolatedModules in particular is useful because of the additional checks TS performs when it is enabled. I'm not going to turn it off, particularly when the dependency has already accepted a proposed fix. Making a release on shiki is easy. I expect Orta will likely merge https://github.com/shikijs/shiki/pull/213 as soon as he sees it, which will automatically create a release, and then TypeDoc can update to 0.9.8 without turning off compiler checks.

> I'm wiring up a second PR that fixes this without waiting on shiki, and which allows shiki to go back to the decidedly better as const statement

Huh? `const enum` and `as const` are completely orthogonal. The extra assertion is fine, and resolving those types of errors is expected for a pre-1.0 package...

> Also breaks a lot of tooling.

Breaks... what? I've never seen an audit warning do more than get in the way for a package which keeps its dependencies up to date. I'm not opposed to fixing things, just pointing out that the npm community as a whole freaks out about things which aren't real issues frequently.

StoneCypher:
This PR was replaced by something different 😢

Gerrit0:
Published in 0.21.8 - I'm going to be out most of the day and didn't want to make you wait on a release unnecessarily...

StoneCypher:
> > It's much easier to just remove isolatedModules in your thing
>
> Yes, and also wrong. 

Respectfully, no, it's not.

The reason that this was happening is that you're attempting to import an enumeration from a different module.

The specific effect of `--isolatedModules` is to prevent that.  I've been talking to Anders a lot; the entire option is just a performance hack, and really shouldn't exist.  The entire Google team feels that way.

&nbsp;

> A dependency should not dictate that compiler options must be set to more lax values.

Respectfully, there is no "strictness" to `--isolatedModules`.  All the option does is prevent the importation of types, but what you are doing is explicitly to import a type.

I don't mean to be rude, but you're engaging in a bug.

&nbsp;

> I'm not going to turn it off, particularly when the dependency has already accepted a proposed fix.

The fix that you sent them isn't correct.

&nbsp;

> Huh? const enum and as const are completely orthogonal.

That's true.  They bear no relationship to one another.

However, the value `as const` was correct, previously.  I can no longer use `shiki` the correct way in my own projects because of this change that you've made for them.  This also requires them to restate values that should be write-once; the `as const` is a formal enabler to allowing them to extract the array from the values as a type directly.  This is actually explicitly forbidden by the Google typescript rules

Unfortunately I'm not very good with people, and this is a subtle bug, so I don't think I'm going to be able to convince you that you've made a mistake here. 

This was similar to my experience with the `shiki` folks, who have core mistakes in their `package` that they refuse to repair because they use third party tools which are immune to their bugs.

Unfortunately, you're a major downstream user, so I don't think I'm going to be able to convince them to repair their bug, which affects me, because it faults you on your bug, which you believe isn't a bug.

This would be very easily fixed if you'd discard your isolation, which would be appropriate, ***because your module is not actually isolated***.

However, you seem to believe that telling a module to be isolated is a function of strictness (it is not,) and therefore I seem to be unable to get you to reconfigure the module to be correct, causing downstream problems elsewhere

I end up having to add about 20 lines of type juggling to import shiki as a result of this

I hope that you will consider the possibility that this is, in fact, a bug. 

Unfortunately, you already threw away and overrided my work, so I don't think I can get us there ☹️ 

&nbsp;

> Breaks... what? I've never seen an audit warning do more than get in the way for a package

Basically every CI/CD setup with publish halts on npm audit security warnings, for the obvious reason.

I've been manually publishing 40 packages since this started.

StoneCypher:
> Published in 0.21.8 - I'm going to be out most of the day and didn't want to make you wait on a release unnecessarily...

Consider that when someone has tried to contribute several times in a row, and you decide to redo what they did without them, you are kind of closing the door in their face, and throwing away what they did

A lot of people take pride in helping out

StoneCypher:
In the meantime, this "fix" actually makes my situation worse, because it makes it so that the `shiki` people really can't fix their bug until you change your mind about this fringe compiler option which really doesn't affect you in any way

Please be considerate and understand that what you're doing here causes downstream problems for other people

Gerrit0:
> Consider that when someone has tried to contribute several times in a row, and you decide to redo what they did without them, you are kind of closing the door in their face, and throwing away what they did
>A lot of people take pride in helping out

You're right, I apologize, I should have waited on you.

> However, the value `as const` was correct

I'm still confused. I didn't remove an `as const`. I agree that `BUNDLED_THEMES` being marked with `as const` also being a good change. The exported types from Shiki are left unchanged by my changes. <https://github.com/shikijs/shiki/pull/203/files>

```diff
- export const enum FontStyle {
+ export enum FontStyle {
```

> the entire option is just a performance hack, and really shouldn't exist

This seems odd. In a Q&A in the TS discord last year, Ryan posted the following. I did some searching from public posts from Anders on isolatedModules, didn't find anything to indicate he disagreed. The part I care about isn't the performance, but that it prevents non-modules from being included in the project.

>  > If there is one thing you would want to get rid off in today's TypeScript, what would it be?
>
> Anything banned under --isolatedModules is something we now regret being in the language

StoneCypher:
> The part I care about isn't the performance, but that it prevents non-modules from being included in the project.

But.  You're trying to use something that explicitly has to come from a module?

The original purpose of `--isolatedModules` was to accomodate other compilers, mostly babel, because they do a one-pass on source.  Typescript does a full tree on source instead.

This is germane because both support an implicit numbering for numbered enumerations, and because the syntax allows two collections of enumerations sharing the same values to be the same.

***Which sucks if you get them out of order in one of the two places.***

Now, anyone might be like "well don't be an idiot, let's just pull them in from this other file"

the problem comes in when the "different source" is the same source

consider the case of a test rig like jest.  as a dagger to my heart, they're actually generally built on babel, not typescript, and there's a good reason for that: typescript doesn't expose its ast, so doing things like annotating source with coverage tagging is nigh impossible.  their coping mechanism is to set up a babel dialect which imitates typescript in a discarding fashion, and just throws away all the type annotations.

the problem is, babel is a single pass compiler, so there are times at which the file it's looking at produces different results ***in the context of being included from above***

This comes up a lot, if you just look at the babel mailing list around tooling stuff, especially late ava, early jest, and heckling by the rome people.  Also the old typescript api service from typescript 2 had this single-pass problem

&nbsp;

> > If there is one thing you would want to get rid off in today's TypeScript, what would it be?
>
> Anything banned under --isolatedModules is something we now regret being in the language

that's super weird, because as i understand it this includes enumerations

could you link me to that?  i tried googling the text and several ostenisble variations on the name but it didn't come up.  i'm not discord sophisticated, and if you were able to find that on the spot, it seems likely that you are

if ryan said that then the most likely case is that i'm slinging either an obsolete or a fundamentally originally misunderstood viewpoint, but also that's ... that's super surprising to me at this time, and i want to see if there's a misunderstanding afoot.  ryan is like three different trumps over anything i say minimum, and that's being generous to me

that all said, that text is fairly unambiguous, and i'm may be doing some cartman style hoop jumping

but the general issue as I understand it is that when you use `--isolatedModules` the ts compiler steps down to a one pass itself, [and emits the numeric literals in place of the enumeration, meaning the enumeration is actually gone](https://www.typescriptlang.org/tsconfig#isolatedModules)

and i guess in this case it kind of feels like that defeats the purpose of the enumeration, because all it's really doing is stripping the type back out and enforcing it in the weak way that javascript would

i mean, look, i got my language in typedoc.  i'm giddy, i got what i need, i should shut up and go home.

but also, this is what i believe here.  ymmv

Gerrit0:
> But. You're trying to use something that explicitly has to come from a module?

Yes - and this is fine. `isolatedModules` does not mean that importing things is disallowed. It means that files must be able to be compiled without knowledge of contents of the imported files. This is exactly the babel use case that you mentioned. The reason importing `const enum`s is disallowed is that without `isolatedModules`, TypeScript will compile:

```ts
// a.ts
const enum Foo { a }
// b.ts
import { Foo } from "./a"
const a = Foo.a
```
into
```js
// b.js
const a = 0 /* a */;
```
... but with `isolatedModules` I can't replicate this because the transpilation for `b.ts` has no knowledge of `a.ts`.

> typescript doesn't expose its ast

Huh? This is definitely off. TypeDoc uses TypeScript's AST. https://ts-morph.com/ is a fairly popular wrapper around TypeScript that makes the AST easier to use.

> that's super weird, because as i understand it this includes enumerations

I a post from sandersn best describes the team's stance on this, again from discord: https://discord.com/channels/508357248330760243/508357638677856287/760917427159695452
```
There's a spectrum:

## Why Would You Do This (We Do This) ##

- namespaces containing values

## We Can't We Stop Doing This Ourselves ##

- const enums
- enums from consts
- enums at all

## Don't Do This Unless It's Really Convenient ##

- overloads

## Legitimate Stylistic Choice Except Where It's Wrong ##

- interface vs type alias
```

> could you link me to that? 

You'll have to join the TypeScript discord: https://discord.com/invite/typescript
Then once joined, go to: https://discord.com/channels/508357248330760243/746090347863474261/746117428097974272
This is in the ts-team-qna-20-08-20 channel under Archived, it's pretty short, might be informative to read through the whole thing.

> and emits the numeric literals in place of the enumeration, meaning the enumeration is actually gone

You have this backwards. With `isolatedModules`, TS refuses to emit the literals. Without it, `const` enums will result in literals being emitted. [Playground](https://www.typescriptlang.org/play?isolatedModules=false#code/PTAEAEEsGcHsBsCGAXApgEwLK3QV3qtAFygBmi80qAUKgB4AOsATsqAN4C+11AxrADtobVANwBbUADFYsDqEQAaUACNu-IW0SgAvNNkA6REA)

> and i guess in this case it kind of feels like that defeats the purpose of the enumeration, because all it's really doing is stripping the type back out and enforcing it in the weak way that javascript would

Numeric enumerations in the first place are completely unsafe.

```ts
enum Foo { a, b }
const a: Foo = 123324 // no error!
```

StoneCypher:
> > But. You're trying to use something that explicitly has to come from a module?
>
> Yes - and this is fine. isolatedModules does not mean that importing things is disallowed. 

Nobody said it did.  

In order for an enumeration to be imported, it must be imported from a module, according to the documentation page I gave earlier.

Removing the specific noun changes the nature of the statement critically.

I didn't say importing things was disallowed.  I said "enumerations have to come from a module, you're importing an enumeration, and you're forbidding importing things that have to come from other modules."

I think the key difference in how you and I see this is that you believe you're still importing the enumeration.  I do not.  I believe you're importing a stripped number which is not the type.

&nbsp;

> This is exactly the babel use case that you mentioned. 

Thanks, I know.  This is why I brought it up.

&nbsp;

> ... but with isolatedModules I can't replicate this because the transpilation for b.ts has no knowledge of a.ts.

Thanks, this was my point.  I'm not sure why you're trying to explain this back to me.

&nbsp;

> > typescript doesn't expose its ast
>
> Huh? This is definitely off. TypeDoc uses TypeScript's AST. 

It is off because it is significantly altered again.

What I was trying to describe is jest's need for a read/write AST, so that it can write to the AST.

A read only inspection of the AST is not what I describe.

The jest team explains this as the reason they used Babel.

&nbsp;

> > and emits the numeric literals in place of the enumeration, meaning the enumeration is actually gone
>
> You have this backwards.

I'm sorry, but no, I don't.  The part I was complaining about is that the enumeration - the type - is gone.

Did I misunderstand what's actually left behind?  Yes, as a nearly ten year user of typescript, I did.

Because I'm being forced to try to understand Javascript in terms of Typescript, and to guess what the result will be.  And the result appears to be radically lower quality code, and the loss of a type.

But I made an irrelevant mistake.  And unfortunately, when you're appealing to another programmer to approach something a different way, and they don't want to, if you say "here is the value being lostt," instead of looking at what you're saying, they'll point out that you misspelled lost.

Okay

The point is, the type is gone, and experienced developers no longer correctly recognize what the code is actually doing.  The emitted code does significantly more work with no benefit, and the types that Typescript is involved to create are being destroyed.

And it really doesn't matter what, specifically, is being left in their place

This is not desirable

&nbsp;

> a post from sandersn best describes the team's stance on this, again from discord:

I seem to be reading this post as saying "we can't stop doing this either," which is a direct contravention of the previous quote by a different person.

I suspect he can't stop doing this either because he needs the same thing that I need here: real type checking, instead of burning the types to the ground in the hope of pursuing a configuration element that delivers no clearly stated value.

So far the only justification given is that it is "stricter," but then it's clearly stated that the author openly believes that this changes what's being compiled, which to me is the exact opposite of strict.

This does not appear to prevent any defects, and it does appear to introduce them.

I confess, I feel like there's no actual supporting first principle here, and given that it's causing downstream problems in multiple libraries, including forcing a non-desirable patch on others, I just don't understand the motivation here.

&nbsp;

> Anything banned under --isolatedModules is something we now regret being in the language

So, I did ask you to source this quote by Ryan.  You decided to source some other unrelated thing.

However, this was enough.  This got me to the channel, and I was able to cut and paste the quote into the search box, which turned up a stray random offhanded comment out of context by Ryan.

And I mean.

Look, being direct?  This could only have been found by search.  It's not actually part of the discussion you're referencing, and it's out of context.  This is a wide reaching generic comment, and I've seen him use const enum since this in production code.

I don't think he was thinking about this when he said this, and I think this is fairly obviously a wide reaching offhanded comment that shouldn't be taken as universal design advice.

I don't think there is any meaningful specific reason to do this, and I think the power of search is being used to find any long justification for a thing that's causing real world problems without any specific cause.

The manual page explicitly gives this as a consequence.  The community strongly suggests against using this flag.  Your upstream dependency had to make an undesirable change to support this.  They are now required to repeat data because they cannot use the language feature designed exactly for what they're doing right now.

My opinion is that this is quite clearly a bug.

I'm also of the opinion that it's been well demonstrated, sourced in the manual, and that I'm basically dealing with someone who's googling for reasons to say no.

Unfortunately, I will now add 25 lines of unnecessary type juggling to my code, and lose much of the expressive power of Typescript.

Part of open source is genuinely listening when other people have needs.

Thank you for a product that delivers me value.

&nbsp;

> Numeric enumerations in the first place are completely unsafe.

The end result of this code is dramatically less safe than doing things the normal typescript way and not engaging obscure compiler configurations which are not appropriate in context

The end result of this code is to force other libraries to adapt to you, and lose their type safety in the process

You gain nothing from this

I do not understand why you decline to be flexible

PR #1677, opened 2021-09-07
Add support for a `typedocMain` field in package.json.

### LINKED ISSUE #1674: Generating documentation while using webpack 4 and ts-loader.
## Search terms

webpack, ts-loader, yarn, workspace

## Expected Behavior

I expected that I could document a package in a Yarn workspace that had been bundled using webpack. The command would look something like: typedoc --packages ./packages/<name>

## Actual Behavior

The issue is that the detection logic doesn't pick up the original source file of the package.

```Entry point "/<snip>/build/webpack:/webpack/universalModuleDefinition" does not appear to be built by the tsconfig found at "/<snip>/tsconfig.json"```

## Steps to reproduce the bug

Run `yarn webpack --mode production`.
Output a `umd` bundle.
With the `devtool` setting: `source-map`
Using: `typedoc --packages ./packages/<name>`.

Types are generated using `tsc` directly, and output to a `lib-types` and a `d.ts` rollup is done using `api-extractor`.

Webpack:
- Entrypoint: "src/index.ts"
- Output: "build/index.js"

package.json:
- main: "build/index.js"
- typings: "build/index.d.ts"

Versions:
- webpack: ^4
- ts-loader: ^8

## Environment

-   Typedoc version: 0.21.9
-   TypeScript version: 4.2.4
-   Node.js version: v12.20.1
-   OS: OSX

itsjamie:
This can be used to generate a tree of your source using `tsc` and point
at it to generate documentation outside your normal build process.

This takes a different tack and instead of falling back to the typings, this allows you to set a specific packageMain value that is built directly with TSC. This can act as an escape hatch for projects with weird builds that involve webpack / rollup / etc, and more complex typings setups with api-extractor to go along with the bundling.

For example, in our particular case, we use webpack 4 and api-extractor to bundle packages that are used at development time, but publicly are exposed as a single package. Therefore we bundle the TS via webpack and tsloader, and bundle the declaration files using an api-extractor rollup.

Resolves #1674.

### DIFF
--- a/src/lib/utils/package-manifest.ts
+++ b/src/lib/utils/package-manifest.ts
@@ -207,6 +207,11 @@ export function getTsEntryPointForPackage(
     let packageMain = "index.js"; // The default, per the npm docs.
     let packageTypes = null;
     if (
+        hasOwnProperty(packageJson, "typedocMain") &&
+        typeof packageJson.typedocMain == "string"
+    ) {
+        packageMain = packageJson.typedocMain
+    } else if (
         hasOwnProperty(packageJson, "main") &&
         typeof packageJson.main == "string"
     ) {


itsjamie:
Just a note, you don't even need to generate the tree with `tsc` outside the binary. You can literally point `typedocMain` at your entry point. Because you support running directly on the TypeScript source, this means no extra steps.

cheelahim:
@Gerrit0 What's your take on this one? I have a few projects which would benefit from this option.

Gerrit0:
I need to pull it down and play with it, fix the CI failure, add a test, and write some docs for it, it seems like an elegant solution though.

Gerrit0:
TIL, GitHub's `gh` tool merges remotely :/ that's annoying, wanted to play with it first!

itsjamie:
For future reference, If you want to use the `gh` tool locally, you could use: `gh pr checkout 1677`

Gerrit0:
Good to know, thanks! This has been released in 0.22.0

PR #1691, opened 2021-09-12
Remove old options from basic example run & run.bat

srmagura:
Looks like the `--module` and `--target` options were removed in a previous release.


Gerrit0:
Thanks! Yeah, that's been broken for a while now. Support for specifying TypeScript options on the command line was dropped in 0.20 if I remember correctly.

PR #1704, opened 2021-09-20
Infer the name of a destructured parameter

### LINKED ISSUE #1703: Derive parameter name of desctructured parameter from JSDoc comment
## Search Terms

__namedParameters, destructured parameter,  `@param`

## Problem

TypeDoc sets the name of any destructured parameters to "__namedParameters", even when a name is given in the JSDoc comment. This also results in the description of this parameter (and any child properties) in the JSDoc comment being ignored.

This relates to the "`@param` dot syntax" that is currently [under consideration in TSDoc](https://github.com/microsoft/tsdoc/issues/19). TypeDoc already supports the dot syntax for parameters with names (i.e. when they are not being destructured), but it doesn't work for the more common scenario where the parameter is destructured inline.

For example, with this code snippet, all descriptions are rendered correctly:

```typescript
/**
 * Concatenate strings 'a' and 'b'
 *
 * @param options - The options.
 * @param options.a - The 'a' string.
 * @param options.b - The 'b' string.
 */
function concatAB(options: { a: string, b: string }) {
  const { a, b } = options;
  return a + b;
}
```

But in this example, the `@param` tags end up being ignored because TypeDoc doesn't understand that the `options` param is the descructured parameter:

```typescript
/**
 * Concatenate strings 'a' and 'b'
 *
 * @param options - The options.
 * @param options.a - The 'a' string.
 * @param options.b - The 'b' string.
 */
function concatAB({ a: string, b: string }) {
  return a + b;
}
```

## Suggested Solution

We could set the name of the parameter to the name given in the JSDoc comment. This would also ensure the descriptions are correctly matched up. This relies upon the assumption that the JSDoc `@param` tags are present for all parameters, and that they are correctly ordered, so we can add a check to ensure this replacement only occurs if this appears to be the case (or maybe we can make this configurable?).

I am unaware of what other similar tools do in this circumstance.

I'd be happy to do this in a plugin instead, but I'm not sure that it would be possible. This relies upon the parsed comment, so it can't be triggered by the `EVENT_CREATE_SIGNATURE`, `EVENT_CREATE_DECLARATION`, or `EVENT_CREATE_TYPE_PARAMETER` event, because how would we be certain the comment was resolved by then? I think it would need to be done during the "resolve" phase, which is when the information in the parsed comment gets incorporated into the signature at the moment. So we could do this in a plugin that gets triggered by the beginning of the resolve phase, but that would be challenging without duplicating the functionality of [`addCommentToSignatures`](https://github.com/TypeStrong/typedoc/blob/949441255885d42ccdf4a9330402f6b01b18b759/src/lib/converter/plugins/CommentPlugin.ts#L320), or somehow guaranteeing that the plugin gets run before that function.

Gudahtt:
The name of a destructured parameter can now be inferred from a JSDoc comment. The name is only inferred if the number of top-level `@param` tags in the JSDoc comment exactly matches the number of parameters.

Closes #1703

### DIFF
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -349,8 +349,21 @@ export class CommentPlugin extends ConverterComponent {
                 childComment.tags ||= [...comment.tags];
             }
 
-            signature.parameters?.forEach((parameter) => {
+            signature.parameters?.forEach((parameter, index) => {
                 let tag: CommentTag | undefined;
+                if (childComment && parameter.name === "__namedParameters") {
+                    const commentParams = childComment?.tags.filter(
+                        (tag) =>
+                            tag.tagName === "param" &&
+                            !tag.paramName.includes(".")
+                    );
+                    if (
+                        signature.parameters?.length === commentParams.length &&
+                        commentParams[index].paramName
+                    ) {
+                        parameter.name = commentParams[index].paramName;
+                    }
+                }
                 if (childComment) {
                     moveNestedParamTags(childComment, parameter);
                     tag = childComment.getTag("param", parameter.name);
--- a/src/test/converter/function/function.ts
+++ b/src/test/converter/function/function.ts
@@ -80,6 +80,102 @@ export function functionWithRest(...rest: string[]): string {
 }
 
 /**
+ * This is a function with a destructured parameter.
+ *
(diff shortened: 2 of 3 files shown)


Gerrit0:
Looks good! I would have put the tests exactly where you did, or potentially in converter2/behavior, depending on my mood at the time.

PR #1706, opened 2021-09-21
Use data-theme attribute for themes on root element

Shane4368:
**Use CSS color-scheme property**
This sets light/dark theme on scrollbars and form elements.

**Remove switch statement and directly apply theme in setTheme function**
This change allows additional themes added by a plugin to work without attaching a new listener to #theme element; also prevents rechecking of localStorage to add the custom theme to body.classList.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "typedoc",
       "version": "0.22.4",
       "license": "Apache-2.0",
       "dependencies": {
--- a/src/lib/output/themes/default/assets/typedoc/Theme.ts
+++ b/src/lib/output/themes/default/assets/typedoc/Theme.ts
@@ -13,17 +13,5 @@ export function initTheme(choices: HTMLOptionElement) {
 }
 
 function setTheme(theme: ThemeChoice) {
-    switch (theme) {
-        case "os":
-            document.body.classList.remove("light", "dark");
-            break;
-        case "light":
-            document.body.classList.remove("dark");
-            document.body.classList.add("light");
-            break;
-        case "dark":
-            document.body.classList.remove("light");
-            document.body.classList.add("dark");
-            break;
-    }
+    document.documentElement.dataset.theme = theme;
 }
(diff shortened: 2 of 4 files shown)


Gerrit0:
This seems reasonable, but I'm not sure it's sufficient. If you add a new dark theme option, but the user's OS theme is set to light, code blocks in the rendered page will be rendered with the light theme when selecting your dark theme...

Is there any major benefit to using a data-theme attribute rather than keeping with classes? Doing this switch is a breaking change that doesn't seem necessary.

Shane4368:
Based on my testing, the code blocks will match the selected theme. https://typedoc-theme-test-shane4368.surge.sh/classes/Bird.html#beginFlight

Using classes for themes would require additional logic to replace the current theme with the selected theme.
```html
<html class="default no-js not-mobile toggle-visibilityprivate toggle-inherited dark">
```
Using the above as reference, if you wanted to change theme, you'd have to do `html.classList.remove("dark")` then `html.classList.add("selected-theme")`. But "dark" could be any value; you'd have to track the current theme or `html.classList.remove(localStorage.getItem("tsd-theme"))`.

`body.light { color-scheme: light; }` also doesn't override if it's set on the root element, resulting in dark scrollbar on light theme if OS theme is dark.

If it's a breaking change, you may add this in 0.23.

Gerrit0:
Tracking the current theme doesn't seem too bad to me - at least until 0.23, where the data attribute is probably better going forward. I didn't know about the colored scrollbars possibility - that's neat, and a really compelling reason to move to `:root`

Just a note: The hardcoded hl-1 through hl-7 might work today, but there's no guarantee that it works in the future. Those classes are completely dependent on the themes the user chooses. If someone writes a theme that has 50 different colors, it'll behave badly. I'd really rather give plugins like yours a way to handle this nicely. Maybe themes in the dropdown could specify two values?

Shane4368:
The hl colours being out of my control was an oversight.
> Maybe themes in the dropdown could specify two values?

Like setting a custom hl theme before the html gets written to file, if it's supported? I'm okay with that.

Gerrit0:
I was thinking of making values contain two items, e.g. `hl-light light` and `hl-dark dark`, then your themes would specify `hl-light custom` for example.

Unfortunately supporting custom highlighter themes for each scheme is *expensive*, adding ~400ms per 1000 lines of code highlighted for each theme (https://github.com/shikijs/shiki/issues/234)

Shane4368:
Oh. That should be fine.

Issue #1556, opened 2021-03-29
Update the examples folder with a real example


Gerrit0:
The `examples/basic` folder might have originally contained an example, but has been used for UI tests for a really long time. Conflating these two things was a bad idea, since it means that the examples folder that ~80 people look at every week contains almost useless info.

srmagura:
I could work on this if you want @Gerrit0.

What should the existing `examples` folder be renamed to? `testProjects`?

What would you like to see in the new example? I have a variety of already-commented code that I could pull from. Or I could write some new code just for this.

Gerrit0:
That would be great! I'm not too particular about what to call the visual tests - it might actually be best to move it into src/test/renderer.

Ideally what I'd like to see in an example is something that shows how TypeDoc handles language constructs, so something with exported variables, classes, functions, overloaded functions, etc...

I'd also like to include features which people might not immediately think about, like a class `X extends Array<string>` which resolves the type parameters in inherited methods, and other special casing, like changing `as const` variables into enums with `@enum` (#1675), detecting variable-functions and displaying them as functions, exporting the same variable under an alternative name... probably other stuff I'm missing.

Once this is done, I'd really like to prominently link to both the example folder and generated documentation on the website.

srmagura:
Thanks for the guidance 🙂. I will probably work on this this weekend.

- I think the `example` folder should have its own `package.json` and `tsconfig.json`. I was planning on adding React as a dependency of the example project so that I can show how to document a React component.
- Do you want the generated documentation to be committed or gitignored?

Gerrit0:
Separate package/config - I agree. React sounds fine.

The generated output should be .gitignored, it will be published on typedoc-site, but I think rebuilding it like /api is done today is the right way to go about it.

PR #1717, opened 2021-09-29
Move visual tests to src/test/renderer

srmagura:
Working towards completing https://github.com/TypeStrong/typedoc/issues/1556.

Btw, the visual regression tests workflow seems to be broken. I'm not sure if I broke it or if it was already like this. The visual tests work when I run them locally.

```
Run npx reg-suit run --test
[reg-suit] info version: 0.10.16
[reg-suit] info Skipped to detect the previous snapshot key because key generator plugin is not set up.
[reg-suit] info Skipped to fetch the expected data because expected key is null.
[reg-suit] info Comparison Complete
[reg-suit] info    Changed items: 0
[reg-suit] info    New items: 0
[reg-suit] info    Deleted items: 0
[reg-suit] info    Passed items: 0
[reg-suit] info Use 'snapshot_1632892749' as the current snapshot key because key generator plugin is not set up.
[reg-suit] info Skipped to publish the snapshot data because publisher plugin is not set up.
[reg-suit] info Skipped to notify result because notifier plugins are not set up.
```


src/test/renderer/self/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cd ${0%/*}
+node ../../../../bin/typedoc --name "TypeDoc Documentation" --tsconfig ../../../../tsconfig.json --readme ../../../../README.md --out doc/ --entryPointStrategy expand ../../../../src/lib/
```
Gerrit0, line 3: I think we can just delete this folder now, there's a typedoc.json under .config which is used to create the docs on the website

Gerrit0 (changes requested):
One thing, then this looks good

Gerrit0:
Thanks!

PR #1743, opened 2021-10-19
Support githubprivate.com in GitHubPlugin

nlfurniss:
Some Github Enterprise clients have a `githubprivate.com` domain, and this supports that

### DIFF
--- a/src/lib/converter/plugins/GitHubPlugin.ts
+++ b/src/lib/converter/plugins/GitHubPlugin.ts
@@ -72,6 +72,13 @@ export class Repository {
                     repoLinks[i]
                 );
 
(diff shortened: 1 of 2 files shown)


Gerrit0:
Is there some GitHub documentation somewhere where I can look up what domains GH enterprise can use? I've had no issue finding any, wondering if there's other ones that are missing.

nlfurniss:
> Is there some GitHub documentation somewhere where I can look up what domains GH enterprise can use? I've had no issue finding any, wondering if there's other ones that are missing.

Sadly I couldn't find any either 😕. I also fixed the prettier issues

PR #1744, opened 2021-10-19
Increase `git ls-files` maxBuffer to 10MB

nlfurniss:
[This line](https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/plugins/GitHubPlugin.ts#L97) in GitHubPlugin throws an error (`Error: spawnSync git ENOBUFS`) for my repo, which has 19K tracked files. If I add `"*.?s"` to ls-files, I still have 14.5K files and the same error.

By increasing the buffer, this feature will work for very large repos


Gerrit0:
Thanks!

PR #1748, opened 2021-10-20
Updated readme to remove minimal theme

RunDevelopment:
I noticed that the minimal theme was still mentioned in the readme. [The website](https://typedoc.org/guides/options/#theme) has already been updated, so here is an update for the readme.


Gerrit0:
Thanks!

PR #1750, opened 2021-10-20
Fixed gap after footer

### LINKED ISSUE #1749: Gap after footer if `hideGenerator: true`
## Search terms

hideGenerator, footer, dark mode

## Expected Behavior

The footer should visually be the last element on the page.

image

## Actual Behavior

There is a gap after the footer. You can only see this gap in dark mode.

image

This is reproducible with any TypeDoc config that contains `"hideGenerator": true`.

## Environment

-   Typedoc version: 0.22.6
-   TypeScript version: 4.4.2
-   Node.js version: 16.x
-   OS: Wind10
-   Browser: Chrome 94, Firefox 93

RunDevelopment:
This fixes #1749. The screenshot in the expected section actually uses this fix. 

The problem was that the margin of the last `<p>` went outside the footer.

image

image

This PR adds an invisible pseudo-element to fix that.

image

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -641,6 +641,10 @@ footer {
     border-top: 1px solid var(--color-panel-divider);
     background-color: var(--color-panel);
 }
+footer:after {
+    content: "";
+    display: table;
+}
 footer.with-border-bottom {
     border-bottom: 1px solid var(--color-panel-divider);
 }


Gerrit0:
The invisible pseudo element really feels like a hack to me... CSS is awful.

PR #1769, opened 2021-10-28
Remove `no-js` class when JS is enabled

### LINKED ISSUE #1768: `no-js` class is not removed when JS is enabled
## Search terms

html, no-js, js

## Expected Behavior

The `no-js` is removed from the `html` element when JS is enabled.

## Actual Behavior

The `html` element always has `no-js` class.

image

## Steps to reproduce the bug

Open any page generated by TypeDoc.

## Environment

-   Typedoc version: 0.22.7
-   TypeScript version: 4.4.2
-   Node.js version: 16.x
-   OS: Win10

RunDevelopment:
This fixes #1768.

### DIFF
--- a/src/lib/output/themes/default/layouts/default.tsx
+++ b/src/lib/output/themes/default/layouts/default.tsx
@@ -29,7 +29,12 @@ export const defaultLayout = (context: DefaultThemeRenderContext, props: PageEve
         </head>
         <body>
             <script>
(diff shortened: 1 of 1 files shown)


RunDevelopment:
See [here](https://github.com/TypeStrong/typedoc/issues/1768#issuecomment-955710322).

PR #1774, opened 2021-10-30
Fix months being off by one in CHANGELOG

srmagura:
I noticed that the Changelog indicated I had traveled back in time to implement GitLab source linking... 😂


Gerrit0:
Oops! Time is hard.

srmagura:
More like, why in the world does `getUTCMonth()` start counting at 0??? It makes no sense lol.

PR #1783, opened 2021-11-07
Ensure meta charSet is within first 1024 bytes

RunDevelopment:
This fixes the problem I mentioned [here](https://github.com/TypeStrong/typedoc/commit/8c707b3d95cffc4fccff6e9b347b71306bdb0889#r59589758).

I choose not to document this. While the docs say "immediately after the opening `<head>` tag", I don't think plugin devs should be concerned with this minor implementation detail.


Gerrit0:
I think that's fair - theoretically the charset tag could be added with a hook that has an order of `-Infinity`, and nobody would be able to tell the difference ;)

PR #1805, opened 2021-11-26
Upgrade dependencies of example

srmagura:
This fixes a compilation error in the example. The example was updated to use the `Awaited` type introduced in TypeScript 4.5, but the example's `package-lock.json` was still referencing TypeScript 4.4.


Gerrit0:
Oops, thanks!

PR #1808, opened 2021-11-26
Distinguish between block and other tags when parsing comments

### LINKED ISSUE #1807: Some JSDoc tags prevent class docstrings from showing
## Search terms

tags, docstring, class, extends, jsdoc

## Expected Behavior

TypeDoc should be able to render docstrings no matter what JSDoc tags are present in them.

## Actual Behavior

Some tags (`@class`, `@extends`) prevent class documentation from showing.

For instance the documentation for these three classes will not render:

```ts
/**
 * @class NoDocs1
 *
 * This is another class.
 */
class NoDocs1 {}

/**
 * @extends BaseClass
 *
 * This is an other class.
 */
class NoDocs2 extends BaseClass {}

/**
 * @class NoDocs3
 * @extends BaseClass
 *
 * Some docs.
 */
class NoDocs3 extends BaseClass {}
```

NoDocs1

However, adding a made-up tag afterwards restores the class documentation:

```ts
/**
 * @class WithDocs1
 * @madeupTag
 *
 * This is a different class.
 */
class WithDocs1 {}

/**
 * @class WithDocs2
 * @extends BaseClass
 * @anotherMadeupTag
 *
 * This is a different class.
 */
class WithDocs2 extends BaseClass {}
```

WithDocs1

I have only tested `@class` and `@extends` tags for this behavior, but it's possible that this also affects other tags (and nodes other than classes).

## Steps to reproduce the bug

1. Clone [the reproduction repository](https://github.com/ejuda/typedoc-tags-in-docstrings-repro).
2. `npm ci`
3. `npm run docs`
4. This will generate documentation in `docs` directory.

## Environment

-   Typedoc version: 0.22.10
-   TypeScript version: 4.5.2
-   Node.js version: 14.15.4
-   OS: Windows 10

ejuda:
When parsing the contents of comments, TypeDoc treats all tags as block tags. That is, if it encounters a tag, it parses whatever follows it as tag contents, until it reaches the end of the comment or finds another tag.

According to [TSDoc standard](https://tsdoc.org/pages/spec/tag_kinds) this is correct behaviour for block tags but not for modifier or inline tags. This PR modifies the behaviour by ensuring that only block tags are parsed this way and that for modifier or inline tags parsing of tag content stops when we encounter an empty line.

As a side effect, when some JSDoc tags were used at the beginning of the comment (e.g. `@class`), the entire docstring was often removed. This is because TypeDoc also attempts to remove redundant JSDoc tags from the comments. Erroneous assignment of the entire comment as the tag content meant that everything was deleted. Following this fix, only the actual tag content (e.g. the class name for `@class` tag) will be removed.

Closes #1807.

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -4,6 +4,28 @@ import { Comment, CommentTag } from "../../models/comments/index";
 import type { Logger } from "../../utils";
 
 /**
+ * List of block tags as defined by TSDoc standard.
+ *
+ * The TSDoc standard states that "All text following a block tag, up until the
+ * start of the next block tag or modifier tag, is considered to be the block
+ * tag’s tag content." Therefore we need to differentiate between block tags
+ * and other kinds of tags to know where to stop processing the tag content.
+ */
+const BLOCK_TAGS = [
+    "decorator",
+    "deprecated",
+    "defaultValue",
+    "example",
+    "param",
+    "privateRemarks",
+    "remarks",
+    "returns",
+    "see",
+    "throws",
+    "typeParam",
+];
+
+/**
  * Check whether the given module declaration is the topmost.
  *
  * This function returns TRUE if there is no trailing module defined, in
@@ -189,7 +211,7 @@ export function parseComment(
     text: string,
     comment: Comment = new Comment()
 ): Comment {
-    let currentTag: CommentTag;
+    let currentTag: CommentTag | undefined;
     let shortText = 0;
 
     function consumeTypeData(line: string): string {
@@ -200,7 +222,12 @@ export function parseComment(
 
     function readBareLine(line: string) {
         if (currentTag) {
-            currentTag.text += "\n" + line;
+            currentTag.text += "\n";
+            if (!BLOCK_TAGS.includes(currentTag.tagName) && line === "") {
+                currentTag = undefined;
+            } else {
+                currentTag.text += line;
+            }
         } else if (line === "" && shortText === 0) {
             // Ignore
         } else if (line === "" && shortText === 1) {
--- a/src/test/converter2/issues/gh1807.ts
+++ b/src/test/converter2/issues/gh1807.ts
@@ -1,0 +1,50 @@
+/**
+ * This is a class.
+ */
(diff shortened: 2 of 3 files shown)


Gerrit0:
I see one problem with this - `@class` is a block tag. JSDoc has only two types of tags - block and inline, and the only inline tags are `@link` (+ synonyms) and `@tutorial`... This is why the [JSDoc documentation](https://jsdoc.app/tags-class.html) puts the `@class` annotation at the end of the comment. I think this is probably working as intended.

ejuda:
I understand this concern and the JSDoc spec does not make it easier - `@class` is indeed a block tag although [the advertised syntax](https://jsdoc.app/tags-class.html) does not suggest that anything after a new line should be parsed as tag content.

That being said, I think the PR makes the user experience better: the change will not hurt those who are compliant with the JSDoc standard but will help those who did something slightly odd and added extra documentation after the class block. I'd be happy to go through the JSDoc spec and look for tags which can be expected to behave more like TSDoc block tags (e.g. `@licence`).

I also think the change helps improve correctness of parsing TSDoc modifier tags. From what I understand, currently the following comment:

```ts
/**
 * @beta
 *
 * A class.
 */
class SomeClass {}
```

would be parsed as an `@beta` tag with `A class.` as its tag content. Since modifier tags are expected to have empty tag content, this is not correct, and [TSDoc playground](https://tsdoc.org/play#) interprets `A class.` as the class description. While this issue is not currently visible for the user, the change would increase code correctness.

ejuda:
@Gerrit0 Would it be possible for you or another maintainer to have a look at my previous comment please? I just want to know whether this PR has a chance of being accepted or whether I need to work around the issue. Thanks! 😃

Gerrit0:
I do want to look more closely at this - probably won't have time until next weekend though.

ejuda:
@Gerrit0 any further thoughts on this? :)

Gerrit0:
I think this has been superseded by work done for 0.23, which completely reworked comment parsing to allow TypeDoc to be compatible with TSDoc.... I don't really want to continue limping along the current parser since 0.23 should be ready *soon*... should be less than a month for release.

Issue #1804, opened 2021-11-25
Optional arguments are not marked as optional (JSDoc)


3bl3gamer:
## Search terms

optional, isOptional, argument, parameter, flags, JSDoc

## Expected Behavior

Expected `isOptional` flag to be set for arguments, marked as optional with JSDoc (`@param {type} [arg]`).

## Actual Behavior

Argument is required, but it's type gets union'ed with `undefined`.

## Steps to reproduce the bug

```js
// index.js
/**
 * @param {number} [arg]
 */
export function foo(arg) {}
```

```jsonc
// tsconfig.json
{
	"compilerOptions": {
		"allowJs": true,
		"checkJs": true,
		"strict": true,
		"declaration": true,
		"declarationDir": "./types",
		"emitDeclarationOnly": true
	},
	"exclude": ["node_modules"]
}
```

```bash
./node_modules/.bin/typedoc index.js --json docs.json
cat docs.json
```

Output:

```json
Info: JSON written to ./docs.json
{
	"id": 0,
	"name": "undefined",
	"kind": 1,
	"kindString": "Project",
	"flags": {},
	"originalName": "",
	"children": [
		{
			"id": 1,
			"name": "foo",
			"kind": 64,
			"kindString": "Function",
			"flags": {},
			"sources": [
				{
					"fileName": "index.js",
					"line": 4,
					"character": 16
				}
			],
			"signatures": [
				{
					"id": 2,
					"name": "foo",
					"kind": 4096,
					"kindString": "Call signature",
					"flags": {},
					"comment": {},
					"parameters": [
						{
							"id": 3,
							"name": "arg",
							"kind": 32768,
							"kindString": "Parameter",
							"flags": {},
							"type": {
								"type": "union",
								"types": [
									{
										"type": "intrinsic",
										"name": "undefined"
									},
									{
										"type": "intrinsic",
										"name": "number"
									}
								]
							}
						}
					],
					"type": {
						"type": "intrinsic",
						"name": "void"
					}
				}
			]
		}
	],
	"groups": [
		{
			"title": "Functions",
			"kind": 64,
			"children": [
				1
			]
		}
	],
	"sources": [
		{
			"fileName": "index.js",
			"line": 4,
			"character": 0
		}
	]
}
```

Works correctly when used with declarations generated by `tsc`:

```bash
./node_modules/.bin/tsc --project tsconfig.json
./node_modules/.bin/typedoc types/index.d.ts --json docs.json
cat docs.json
[...]

Issue #567, opened 2017-07-26
Comment are not rendered when brackets are around an optional parameter


krenor:
Using [JSDoc Syntax](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

```@param {string} [key] An optional but fancy key```

![With brackets](https://user-images.githubusercontent.com/13726968/28626597-be17ca44-721f-11e7-90d8-c5869608f134.png)

```@param {string} key An optional but fancy key```

![Without brackets](https://user-images.githubusercontent.com/13726968/28626627-cee4c5c0-721f-11e7-83c6-de36b0bb4221.png)

aciccarello:
Hi @krenor, TypeDoc does not completely follow the JSDoc syntax but I think this is something that should be supported. PRs are welcome.

microshine:
I've got the same problem.
I use [Document This plugin](https://marketplace.visualstudio.com/items?itemName=joelday.docthis) for VisualStudio Code which uses correct JSDOC annotations ([JSDOC Optional parameters and default values](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)).

ST-DDT:
Linking to the PR with a proposed fix for better visibility:
https://github.com/TypeStrong/typedoc/pull/1810

PR #1810, opened 2021-11-27
Fixed JSDoc comments and optional params parsing

3bl3gamer:
Closes https://github.com/TypeStrong/typedoc/issues/1804 and https://github.com/TypeStrong/typedoc/issues/567

### DIFF
--- a/src/lib/converter/factories/comment.ts
+++ b/src/lib/converter/factories/comment.ts
@@ -238,6 +238,10 @@ export function parseComment(
             if (param) {
                 paramName = param[0];
                 line = line.substr(paramName.length + 1).trim();
+                // "[optionalArg]" -> "optionalArg"
+                if (/^\[.*\]$/.test(paramName)) {
+                    paramName = paramName.slice(1, -1);
+                }
             }
             line = consumeTypeData(line);
             line = line.replace(/^-\s+/, "");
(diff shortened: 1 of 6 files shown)


pkuczynski:
Any chance to get this merged anytime soon?

ST-DDT (commented):
This works very well for:

````js
 * @param {string} [name] Some optional param.
````

but not for this syntax:

````ts
 * @param [name] Some optional param.
````

as used by tsdocs/typescript.

This is caused by [consumeTypeData](https://github.com/TypeStrong/typedoc/pull/1810/files#diff-228e49910f19c44b8ed2de5981e9f30f3dd398702d47196ee5c241051f0baf33R236) that erases all leading `{}` and `[]` blocks. But if it encounters both, then the cleanup of the `{}` does not remove the separating space and thus the `[]` won't be removed. If the `[]` is the first section, then it will get removed.

Can someone enlighten me why the `[]` is removed/what is it used for originally?

pkuczynski:
> Can someone enlighten me why the `[]` is removed/what is it used for originally in jsdocs?

Its to mark optional parameter: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

ST-DDT:
> Its to mark optional parameter: [jsdoc.app/tags-param.html#optional-parameters-and-default-values](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values)

Currently these get removed completely, so any parameter name that might be in there will be lost, before it is analyzed.
Rendering it unusable. I thought there might be a different usecase where you don't need the thing inside the `[]`.

Gerrit0:
I believe with 0.23, this is handled by the completely reworked comment parser, thanks for the PR though!

PR #1819, opened 2021-12-07
Add option to warn on undocumented items

### LINKED ISSUE #1817: Warning on undocumented items
## Search Terms

- require
- docs
- warning

## Problem

It would be nice to require the inclusion of documentation when a new type is introduced.

## Suggested Solution

Output a warning if an item do not have any documentation associated with it. This could be configured for `functions`, `interfaces`, `types`, `methods`, `classes`, and `variables`.

Nokel81:
Resolves #1817


src/lib/validation/documentation.ts
```diff
@@ -0,0 +1,38 @@
...
+        (prev, cur) => (prev |= ReflectionKind[cur]),
+        0
+    );
+
+    for (const ref of project.getReflectionsByKind(kinds)) {
+        const symbol = project.getSymbolFromReflection(ref);
+        if (!ref.comment && symbol?.declarations) {
+            const decl = symbol.declarations[0];
+            const sourceFile = decl.getSourceFile();
+            const { line } = ts.getLineAndCharacterOfPosition(
```
Gerrit0, line 21: There's an off by one error here, getLineAndCharacterOfPosition returns a 0 based line, but we should display a 1 based line

src/lib/validation/documentation.ts
```diff
@@ -0,0 +1,38 @@
...
+                sourceFile,
+                decl.getStart()
+            );
+            const file = normalizePath(
+                path.relative(process.cwd(), sourceFile.fileName)
+            );
+
+            if (file.startsWith(`node_modules${path.sep}`)) {
```
Gerrit0, line 29: I think this will break badly in monorepos where node_modules is in a directory above where typedoc might be run.

Gerrit0 (changes requested):
Can you add a test?

src/lib/validation/documentation.ts
Nokel81: Ah thanks, will fix.

src/lib/validation/documentation.ts
Nokel81: How about an option to specify folders to ignore within?

Nokel81: How would you recommend going about this? For the usecase that I have I don't want to block building docs if a dependency doesn't have docs.

Dayday10 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+### Features
+
+- Added `--validation.ensureDocumented` option to warn on items that are not documented
```
Gerrit0, line 5: This doesn't match the option name

src/lib/validation/documentation.ts
Gerrit0: I think just `file.includes("node_modules")` would be fine. In a parent directory we'd end up with `"../node_modules/foo"`, which we shouldn't warn about.

Gerrit0 (changes requested):

src/lib/validation/documentation.ts
Nokel81: Sounds good

Dayday10 (approved):

Gerrit0:
Thanks! Will be included in the next release, probably next weekend.

PR #1826, opened 2021-12-22
Changed how index is loaded to improve performance

### LINKED ISSUE #1825: Use `JSON.parse` to load `searchData`
## Search terms

JSON.parse, load, search, performance, searchData

## Expected Behavior

TypeDoc loads search index before the user has a chance to start typing in the search bar

## Actual Behavior

For large indexes (we have a 17MB  one) loading the index takes up to 6-7s. Most of this time is spent in parsing and loading the object literal.

## Suggestion

Use `JSON.parse("<index data>")` instead to load the index. JSON parsing is much faster than JS parsing as the grammar is simpler, and when the index is large the difference is quite big. 

## Steps to reproduce the bug

It's hard to reproduce this without actually sharing our existing index. If needed I can create a dummy repository

## Environment

-   Typedoc version: 0.21.6
-   TypeScript version:  4.3.5
-   Node.js version: 14
-   OS: Windows 10

dragomirtitian:
This PR fixes #1825. Using `JSON.parse` ensures faster loading as the grammar for JSON is much simpler, so parsing JSON instead of JavaScript will be much faster (1s vs 7s in out use case)

We could consider JSON imports at a later date when that proposal reaches stage 4.

### DIFF
--- a/src/lib/output/plugins/JavascriptIndexPlugin.ts
+++ b/src/lib/output/plugins/JavascriptIndexPlugin.ts
@@ -106,6 +106,9 @@ export class JavascriptIndexPlugin extends RendererComponent {
             index,
         });
 
-        writeFileSync(jsonFileName, `window.searchData = ${jsonData}`);
+        writeFileSync(
+            jsonFileName,
+            `window.searchData = JSON.parse(${JSON.stringify(jsonData)});`
+        );
     }
 }


Gerrit0:
JSON imports I believe will run into the same issue as using regular imports when browsing files locally (i.e. not in a web browser), so I think we'll probably want to avoid those unfortunately. LGTM once lint passes!

Gerrit0 (approved):

PR #1838, opened 2022-01-09
Improve contributing experience

albyrock87:
While trying to contribute to the project I found it very difficult to debug unit tests.
Therefore I created this PR to improve dev-experience
- speedup tsc using incremental flag (unfortunately `ts-node` doesn't support `incremental` flag yet)
- add Mocha Explorer VSCode extension support

I had to move `ts-internal` types to a folder to make `ts-node` compile correctly.

image


Gerrit0:
I'm not sure I want to add something that recommends people install extensions whenever they open the project in VSCode... particularly when the description of that extension includes:

> In version 1.59, VS Code added an official API and UI for running tests, which provides all the functionality of this extension and more. Therefore this extension is now deprecated. I will keep maintaining it so it will remain usable but I don't plan to add any major new features to it.

Is there a better way of doing this?

Gerrit0 (changes requested):
https://github.com/TypeStrong/typedoc/pull/1838#issuecomment-1019397963

albyrock87:
@Gerrit0 I looked at available VSCode Mocha engines and unfortunately no one supports native test API.
The reason for this is that Test Explorer extension which is being used by Mocha Test Explorer has a new flag:
> You can keep using this extension as before, but you now also have the option to use VS Code's native testing UI instead by setting testExplorer.useNativeTesting to true in your VS Code settings.

So the author of Mocha Test Explorer will not spend more time trying to convert everything to native API.

What we could do, is to force that flag on the workspace settings.

Gerrit0:
Thanks for looking into this! Sounds like an opportunity for a new project... in the meantime, this works.

Issue #1842, opened 2022-01-14
Add clickable anchor link icon on hover of member titles


adeniszczyc:
## Search Terms

Anchor links, tags

## Problem

On hover of methods across generated typedoc documentation, it is currently not possible to copy a link with an anchor to a specific member. This is common behavior across documentation pages, such as in the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/intro.html).

## Suggested Solution

An anchor link is displayed on hover of a method name, which on click navigates to the anchor link and can be copied from the URL bar. This can be seen in the following GIF:

----
![anchor-link-demo](https://user-images.githubusercontent.com/2099219/149553461-41b93476-e234-4cee-8167-8c81ded29529.gif)
----

PR #1843, opened 2022-01-14
Add clickable anchor link icon on hover of member titles

adeniszczyc:
Resolves https://github.com/TypeStrong/typedoc/issues/1842

This adds an anchor link icon on hover to member titles, to allow copying the URL with the anchor link directly to the member. The icon used for the anchor link is from [tabler-icons](https://github.com/tabler/tabler-icons) (MIT License).


Gerrit0:
I like it! Thanks!

PR #1849, opened 2022-01-17
Fix offset from top of screen on member anchor links

adeniszczyc:
Follow up fix to https://github.com/TypeStrong/typedoc/pull/1843.

In the above PR, the anchor ID was moved to the member `<section>` tag, rather than the `<a>` tag within the section. This had the side-effect of removing the offset from the top of the screen upon clicking on the anchor link. This PR fixes this change by moving the anchor ID back inside of the `<section>` such that upon navigating to the anchor it will be `100px` from the top (as defined by the `.tsd-anchor` class) such that the member shows below the header.


Gerrit0:
Thanks! Not sure how I didn't notice that when playing with it locally

Issue #1881, opened 2022-03-11
Random search results are showing when the query is empty


ychan167:
## Search terms
search, search results 

## Expected Behavior
When the search text is empty/ cleared, it is expected that there are no search results showing.

## Actual Behavior
There are random search results when the search text is empty.
![search_actual](https://user-images.githubusercontent.com/101419436/157914173-c3310ade-f32c-4d1d-8054-1dcf94fef48b.gif)

## Steps to reproduce the bug
1. Go to [example doc](https://typedoc.org/example/index.html)
2. Click search
3. Input some search text
4. Remove the search text

## Environment
- Typedoc version: 0.22.13
- TypeScript version: 4.6.2
- Node.js version: 14.18.0
- OS: Windows 10

Gerrit0:
Resolved by #1882

PR #1882, opened 2022-03-11
fix: do not show random search results when query is empty

ychan167:
This change should fix the problem where random search results are showing when the search text is empty/ cleared and would address https://github.com/TypeStrong/typedoc/issues/1881.

Before:
search_actual

After:
search_expected


Gerrit0:
Thanks!

Issue #1878, opened 2022-03-06
Last signature is missing when multiple are defined


ST-DDT:
## Search terms

- signature
- missing

## Expected Behavior

If I scan a method with multiple signatures, then I expect all of them to be present in the typedoc.json/reflection data.

## Actual Behavior

The last one is always missing.
I need that at least for the default value.

## Steps to reproduce the bug

````ts
export class SignatureTest {
  /**
   * Test with multiple signatures.
   *
   * @param a The value to return.
   */
  multiSignature(a: boolean): boolean;
  multiSignature(a: number): number;
  multiSignature(a: unknown = 1): unknown {
    return a;
  }
}
````

<details>
<summary><strong>typedoc.json</strong></summary>

````diff
{
	"id": 0,
	"name": "@faker-js/faker",
	"kind": 1,
	"kindString": "Project",
	"flags": {},
	"originalName": "",
	"children": [
		{
			"id": 1,
			"name": "SignatureTest",
			"kind": 128,
			"kindString": "Class",
			"flags": {},
			"children": [
[...]
				{
					"id": 4,
					"name": "multiSignature",
					"kind": 2048,
					"kindString": "Method",
					"flags": {},
					"sources": [...]
					"signatures": [
						{
							"id": 5,
							"name": "multiSignature",
							"kind": 4096,
							"kindString": "Call signature",
							"flags": {},
							"comment": {
								"shortText": "Test with multiple signatures."
							},
							"parameters": [
								{
									"id": 6,
									"name": "a",
									"kind": 32768,
									"kindString": "Parameter",
									"flags": {},
									"comment": {
										"shortText": "The value to return.\n"
									},
									"type": {
										"type": "intrinsic",
										"name": "boolean"
									}
								}
							],
							"type": {
								"type": "intrinsic",
								"name": "boolean"
							}
						},
						{
							"id": 7,
							"name": "multiSignature",
							"kind": 4096,
							"kindString": "Call signature",
							"flags": {},
							"parameters": [
								{
									"id": 8,
									"name": "a",
									"kind": 32768,
									"kindString": "Parameter",
									"flags": {},
									"type": {
										"type": "intrinsic",
										"name": "number"
									}
								}
							],
							"type": {
								"type": "intrinsic",
								"name": "number"
							}
						},
+						{
+							"id": 9,
+							"name": "multiSignature",
+							"kind": 4096,
+							"kindString": "Call signature",
+							"flags": {},
+							"parameters": [
+								{
[...]

PR #1890, opened 2022-03-22
Move type serializers to models

fb55:
I started looking into converting serialized data back into reflections. The entire serialisation logic seemed more difficult to understand than necessary, so this is my attempt at simplifying things. Every reflection can now serialise itself, using `super` calls when inheriting props. 

With this, we will be able to colocate serialisers and deserialisers as part of the models.


fb55:
One thing to note is that we don't serialise enough information to actually be able to render documentation from it. So to properly do this, we will have to add the remaining information to the JSON output.

src/lib/models/sources/file.ts
```diff
@@ -4,13 +4,14 @@ import type { ReflectionGroup } from "../ReflectionGroup";
...
 import type { DeclarationReflection } from "..";
+import type { SourceReference as JSONSourceReference } from "../../serialization/schema";
 
 /**
  * Represents references of reflections to their defining source files.
  *
  * @see {@link DeclarationReflection.sources}
  */
-export interface SourceReference {
+export class SourceReference {
```
fb55, line 14: `SourceReference` had to be made a class for this to work.

Gerrit0:
Immediate concern, haven't had time to look at it closely: 

It looks like this might remove the ability for plugins to add additional information to the serialized JSON, which is an important feature. #1878 for a recent example of its use.

fb55:
I've added back support for plugins. Let me know what you think!

fb55:
To explain this change with a bit more depth: When writing logic to move the existing JSON structure back to reflections, it was cumbersome to have to switch back and forth. With these changes, GitHub Copilot is able to generate most of the parsing logic. I have a follow-up ready that will add that part. This PR isn't strictly necessary for that, but given that the `fromObject` logic will have the same structure as this PR, it makes sense to me to get this through as well.

src/lib/serialization/components.ts
```diff
@@ -1,20 +1,14 @@
...
+ * object can be serialized by it. This is done dynamically at runtime via a `supports` method.
  *
  * Additionally, each {@link Serializer} plugin must define a predicate that instructs the group
  * it belongs to.
- *
- * Serializers are grouped to improve performance when finding serializers that apply to a node,
- * this makes it possible to skip the `supports` calls for `Type`s when searching for a
- * `Reflection` and vise versa.
  */
 export abstract class SerializerComponent<T> {
```
Gerrit0, line 13: I think this can just become a non-generic interface now - never liked the horrible static `PRIORITY` thing and it only lived this long because of backwards compatibility.

Gerrit0 (changes requested):
This is awesome, I love it! I'd like a couple minor changes.

src/lib/serialization/components.ts
```diff
@@ -63,29 +28,20 @@ export abstract class SerializerComponent<T> {
...
-}
-
-export abstract class ReflectionSerializerComponent<
-    T extends Reflection
-> extends SerializerComponent<T> {
     /**
-     * Filter for instances of {@link Reflection}
+     * Legacy additional check for the `supports` method.
+     *
+     * @deprecated Use `supports` below instead.
```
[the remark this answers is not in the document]
fb55: Done. I've added an error if users still supply this method, as it might lead to false-positives otherwise.

src/lib/serialization/components.ts
fb55: `toObject` still takes the generic's type, so I left it for now. I did however convert `SerializerComponent` to an interface.

src/lib/serialization/schema.ts
```diff
@@ -314,17 +303,19 @@ export interface MappedType
...
-export interface UnknownType extends Type, S<M.UnknownType, "type" | "name"> {}
+export interface UnknownType extends Type, S<M.UnknownType, "name"> {}
 
 /**
  * Technically not correct, the `type` property will be set by the abstract serializer.
  * But to allow tagged literals, the `type` property is instead defined by each child type.
  */
-export interface Type {}
+export interface Type {
+    type: keyof TypeKindMap;
```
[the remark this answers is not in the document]
fb55: Good point, `SomeType` is a tagged union again.

src/lib/models/types.ts
```diff
@@ -202,23 +198,22 @@ function wrap(type: Type, bp: number) {
...
-    /**
-     * The type of the array elements.
-     */
-    elementType: Type;
-
-    constructor(elementType: Type) {
+    constructor(
+        /**
+         * The type of the array elements.
+         */
```
Gerrit0, line 204: I'd rather not do this as a matter of style - either get rid of the comment, or keep the declaration and parameter separate. Mixing parameters and comments makes the code harder to scan quickly

src/lib/models/types.ts
```diff
@@ -504,32 +499,28 @@ export class OptionalType extends Type {
...
+         * the type is of the form `val is string`
+         */
+        public asserts: boolean,
+
+        /**
+         * The type that the identifier is tested to be.
+         * May be undefined if the type is of the form `asserts val`.
+         * Will be defined if the type is of the form `asserts val is string` or `val is string`.
+         */
+        public targetType?: SomeType
```
Gerrit0, line 521: Same as above

src/lib/models/types.ts
```diff
@@ -858,14 +849,13 @@ export class TemplateLiteralType extends Type {
...
-     * The ordered type elements of the tuple type.
-     */
-    elements: Type[];
-
-    constructor(elements: Type[]) {
+    constructor(
+        /**
+         * The ordered type elements of the tuple type.
+         */
+        public elements: SomeType[]
```
Gerrit0, line 856: Same as above

src/lib/serialization/schema.ts
```diff
@@ -303,19 +293,17 @@ export interface MappedType
...
+export interface UnknownType extends Type, S<M.UnknownType, "type" | "name"> {}
 
 /**
  * Technically not correct, the `type` property will be set by the abstract serializer.
  * But to allow tagged literals, the `type` property is instead defined by each child type.
  */
-export interface Type {
-    type: keyof TypeKindMap;
-}
+export interface Type {}
```
Gerrit0, line 306: The comment can go away now - I'd like to keep the interface though so that it's available as a target for declaration merging (don't think anyone will use it, but....)

src/lib/serialization/schema.ts
```diff
@@ -331,7 +319,7 @@ export interface Comment extends Partial<S<M.Comment, "blockTags">> {
     modifierTags?: string[];
 }
 
-export interface CommentTag extends S<M.CommentTag, "tag" | "name"> {
+export interface CommentTag extends Type, S<M.CommentTag, "tag" | "name"> {
```
Gerrit0, line 322: Not a type

src/lib/serialization/schema.ts
```diff
@@ -350,4 +338,5 @@ export type CommentDisplayPart =
       };
 
 export interface SourceReference
-    extends S<M.SourceReference, "fileName" | "line" | "character"> {}
+    extends Type,
```
Gerrit0, line 341: That's not a type

Gerrit0 (changes requested):

src/lib/serialization/schema.ts
fb55: Good catch

src/lib/models/types.ts
fb55: Makes sense. I've moved this to the constructor documentation.

src/lib/serialization/schema.ts
fb55: Removed the comment. I am a bit split about keeping `Type` — given that TS has a structural type system, keeping an empty interface feels redundant. But let's leave it for now.

Gerrit0:
Thanks! This is a huge improvement for maintainability of serializers.

PR #1892, opened 2022-03-22
Improve some more code

fb55:
Similar to #1889 — updates some code to make it more readable.

### DIFF
--- a/src/lib/converter/plugins/SourceLinkPlugin.ts
+++ b/src/lib/converter/plugins/SourceLinkPlugin.ts
@@ -252,8 +252,11 @@ export class SourceLinkPlugin extends ConverterComponent {
     private getRepository(fileName: string): Repository | undefined {
         // Check for known non-repositories
         const dirName = Path.dirname(fileName);
-        if (this.ignoredPaths.has(dirName)) {
-            return;
+        const segments = dirName.split("/");
+        for (let i = segments.length; i > 0; i--) {
+            if (this.ignoredPaths.has(segments.slice(0, i).join("/"))) {
+                return;
+            }
         }
 
         // Check for known repositories
@@ -275,10 +278,7 @@ export class SourceLinkPlugin extends ConverterComponent {
         }
 
         // No repository found, add path to ignored paths
-        const segments = dirName.split("/");
-        for (let i = segments.length; i > 0; i--) {
-            this.ignoredPaths.add(segments.slice(0, i).join("/"));
-        }
+        this.ignoredPaths.add(dirName);
     }
 
     /**
--- a/src/lib/converter/plugins/SourceLinkPlugin.ts
+++ b/src/lib/converter/plugins/SourceLinkPlugin.ts
@@ -220,7 +220,7 @@ export class SourceLinkPlugin extends ConverterComponent {
     /**
      * List of paths known to be not under git control.
      */
-    private ignoredPaths: string[] = [];
+    private ignoredPaths = new Set<string>();
 
     @BindOption("gitRevision")
     readonly gitRevision!: string;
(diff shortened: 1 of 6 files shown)


src/lib/converter/plugins/SourceLinkPlugin.ts
```diff
@@ -279,7 +277,7 @@ export class SourceLinkPlugin extends ConverterComponent {
         // No repository found, add path to ignored paths
         const segments = dirName.split("/");
         for (let i = segments.length; i > 0; i--) {
-            this.ignoredPaths.push(segments.slice(0, i).join("/"));
+            this.ignoredPaths.add(segments.slice(0, i).join("/"));
```
fb55, line 280: Without a set, this could add sub-paths multiple times.

@Gerrit0 My understanding of this is that this would always ignore the entire path if a single file is ignored. Eg. if `src/components/generated` is in the `.gitignore`, this would ignore everything in `src`. I suspect that we'd rather just add the directory by itself, and not the path leading up to it.

fb55: In other words, we might want to move this loop to the top of the function — the `dirName` should probably be checked if any part of it is ignored.

Gerrit0: >  this would always ignore the entire path if a single file is ignored. Eg. if `src/components/generated` is in the .gitignore, this would ignore everything in `src`.

I read this the same way, and I agree, this definitely looks wrong. I'm surprised there haven't been any bug reports about it.

src/lib/output/themes/default/DefaultTheme.tsx
```diff
@@ -167,8 +167,7 @@ export class DefaultTheme extends Theme {
             event.project.groups.forEach(DefaultTheme.applyGroupClasses);
         }
 
-        for (const id in event.project.reflections) {
-            const reflection = event.project.reflections[id];
+        for (const reflection of Object.values(event.project)) {
```
Gerrit0, line 170: Behavior change here! I think you meant `of Object.values(event.project.reflections)`

src/lib/output/themes/lib.tsx
```diff
@@ -83,10 +81,11 @@ export function classNames(names: Record<string, boolean | null | undefined>, ex
 export function hasTypeParameters(
     reflection: Reflection
 ): reflection is Reflection & { typeParameters: TypeParameterReflection[] } {
-    if (reflection instanceof DeclarationReflection || reflection instanceof SignatureReflection) {
-        return reflection.typeParameters != null && reflection.typeParameters.length > 0;
-    }
-    return false;
+    return (
```
Gerrit0, line 84: I think I prefer the original if I ever had to step into this function, given that I don't think that's ever happened... fine as is.

src/lib/output/themes/default/DefaultTheme.tsx
fb55: Ohh, great catch!

```suggestion
        for (const reflection of Object.values(event.project.reflections)) {
```

src/lib/converter/plugins/SourceLinkPlugin.ts
fb55: I've added a fix for this issue.

src/lib/output/themes/lib.tsx
fb55: Yeah, this one is very much up to personal preference.

Gerrit0:
Thanks! It'll be interesting to see if anyone notices differences from the source link plugin. Given that the previous broken behavior never got reported... I kind of suspect nobody will

PR #1910, opened 2022-04-06
Fix recursive visitor in order to call intersection callback

nlepage:
Hi 

While using `makeRecursiveVisitor`, I noticed the `intersection` callback wasn't called as expected.
This PR fixes the problem.

Cheers.


Gerrit0:
Oops! Good catch, thanks!

PR #1914, opened 2022-04-07
Display flags next to class name

### LINKED ISSUE #1874: Clearly indicate abstract classes
## Search Terms

abstract, class, constructor

## Problems

Consider the following class:

```ts
export abstract class SomeClass {
    abstract abstractProp: boolean;
    abstract abstractMethod(): void;
}
```
<ol>
<li>While TypeDoc does indicate which methods and properties are abstract:

image

 it does not provide any indication on whether the class itself is abstract:

image

 This means that the users cannot easily establish whether the class they are looking at is abstract - they have to rely on abstract methods and properties being present.</li>

<li>Secondly, even if a constructor of an abstract class may be not explictly defined (as in the example above), TypeDoc will still add the constructor to the documentation page. While this is useful for concrete classes (indicating the class has a default constructor), it is potentially misleading for abstract classes, as they cannot be instantiated directly.</li>
</ol>

## Suggested Solutions

<ol>
<li>It would be easier to spot abstract classes if the _Abstract_ badge was present prominently near the class name, for example in the header:

image

<li>If the constructor of an abstract class has not been explicitly defined, TypeDoc should not add an empty constructor, to further emphasise that the class cannot be instantiated directly. If the constructor has been defined, it should be preserved, as it might e.g. be accompanied by a docstring.</li>
</ol>

ejuda:
Reflections in TypeDoc can be assigned flags, describing certain
properties of the reflection, e.g. abstract, private, readonly etc.
These flags were rendered using badges for most type of reflections, but
not for reflections which are displayed on their own page (like classes,
interfaces etc.). This made it difficult to establish e.g. whether a
class is abstract (see #1874).

This PR fixes the issue by rendering flags in the page titles next to
the name of the documented entity.

Styling of the badges has been amended, to account for them now showing
next to much bigger headings. The styling was inspired by [Bootstrap
badges](https://getbootstrap.com/docs/4.0/components/badge/).

Partially resolves #1874.

### DIFF
--- a/src/lib/output/themes/default/partials/header.tsx
+++ b/src/lib/output/themes/default/partials/header.tsx
@@ -1,7 +1,7 @@
 import type { Reflection } from "../../../../models";
 import { JSX } from "../../../../utils";
 import type { PageEvent } from "../../../events";
-import { hasTypeParameters, join } from "../../lib";
+import { hasTypeParameters, join, renderFlags } from "../../lib";
 import type { DefaultThemeRenderContext } from "../DefaultThemeRenderContext";
 
 export const header = (context: DefaultThemeRenderContext, props: PageEvent<Reflection>) => (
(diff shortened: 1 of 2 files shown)


ejuda:
Preview of the changes (for abstract classes):

image

Gerrit0:
I like it, thanks!

PR #1917, opened 2022-04-10
Added more render hooks

matteobruni:
Added new render hooks: `content.begin`, `content.end`, `navigation.begin`, `navigation.end`


Gerrit0:
Seems reasonable to me, thanks!

Issue #1740, opened 2021-10-15
Support for enum-like object declarations enums


michkot:
## Search Terms

as const enum
enum-like objects
union of literals

## Problem

The support for https://github.com/TypeStrong/typedoc/issues/1675 works only for .ts files, but not for .d.ts files (not even those generated from the supported syntax). Could the suport for "as const" object enums be extended for enum-object declarations?
See the https://github.com/TypeStrong/typedoc/issues/1675 for full context.

eg. in addition to

> /** @enum */
> export const EntryPointStrategy = { 
>      /** 
>       * abcd
>       */ 
>      Resolve: "resolve", 
>  } as const; 

there would be also support to detect the following as an enum too:

> /** @enum */
> export ( declare ) const EntryPointStrategy: { 
>      /** 
>       * abcd
>       */ 
>      readonly Resolve: "resolve", 
>  }; 

(This is especially needed when typedoc is executed after pre-processing the .ts APIs by e.g. tsc / api-extractor to .d.ts)

There is even a "convenience variant" for cases when this is manually written:
> /** @enum */
> export const EntryPointStrategy: Readonly<{ 
>      /** 
>       * abcd
>       */ 
>      Resolve: "resolve", 
>  }>; 

but that might be more annoying to detect.

michkot:
I have played around with this a bit recently again, and it seems we are missing per-element jsdoc in all but the standard enum case (at latest version). I am gonna open a follow-up issue once I have time to summarize it.

Issue #1675, opened 2021-09-05
Support for `as const` enums


Gerrit0:
## Search Terms

`as const`, enumerations

## Problem

Since TypeScript's `enum` is rather inconvenient to work with, some packages define their own enum-like objects, TypeDoc does this too.

https://github.com/TypeStrong/typedoc/blob/664e27cfe7b0a91ca3ea8b299f89ca265238fb8c/src/lib/utils/entry-point.ts#L15-L35

However, the generated documentation doesn't make it obvious that this value should be treated as an enumeration.

![image](https://user-images.githubusercontent.com/19329837/132142432-1546c005-b374-4703-ac05-1ec736629599.png)
![image](https://user-images.githubusercontent.com/19329837/132142452-e816e14e-1b6f-4614-9837-2fb868f5406a.png)

## Suggested Solution

Detect variables annotated with `@enum` with all readonly properties that are string literals, and convert them as an enumeration.

If there is also a type alias by the same name whose content is `typeof T[keyof typeof T]`, remove it from the documentation since enums are also types.

Gerrit0:
I'm going to look at this again before releasing 0.23, should TypeDoc do this even if you don't tag the variable with `@enum`?

Vote: :+1: :-1:

michkot:
@Gerrit0 Could this be extended for enum-object declarations?

eg. instead of 

> /** @enum */
> export const EntryPointStrategy = { 
>      /** 
>       * abcd
>       */ 
>      Resolve: "resolve", 
>  } as const; 

there is (in .d.ts ) just

> /** @enum */
> export const EntryPointStrategy: { 
>      /** 
>       * abcd
>       */ 
>      readonly Resolve: "resolve", 
>  }; 

there is even a "convenience variant":
> /** @enum */
> export const EntryPointStrategy: Readonly<{ 
>      /** 
>       * abcd
>       */ 
>      Resolve: "resolve", 
>  }>; 

I am happy to open new issue for this if this is doable!

PR #1920, opened 2022-04-12
Support enum-like objects with numeric literal values

### LINKED ISSUE #1918: Support enum-like objects with numeric literal values
## Search terms

enum, enum-like object

## Problem

TypeDoc already supports enum-like objects with string literal values via #1675 and #1740. However, enum-like objects with numeric literal values are not supported. For the following code example:

```ts
// index.d.ts

/**
 * This enumeration defines some values.
 *
 * @enum
 */
export declare const SomeStringLiteralEnum: {
    /**
     * This is the auto property.
     */
    readonly AUTO: "auto";
    /**
     * This is the manual property.
     */
    readonly MANUAL: "manual";
};

/**
 * This enumeration also defines some values.
 *
 * @enum
 */
 export declare const SomeNumericLiteralEnum: {
    /**
     * This is the auto property.
     */
    readonly AUTO: 0;
    /**
     * This is the manual property.
     */
    readonly MANUAL: 1;
};
```

TypeDoc will document `SomeStringLiteralEnum` as an enum and `SomeNumericLiteralEnum` as a variable:

image

## Suggested Solution

Support enum-like objects with numeric literal values in the same way as those with string literal values. I think it would only require some minor changes in `src/lib/converter/symbols.ts`.

ejuda:
TypeDoc already supports enum-like objects with string literal values
via #1675 and #1740. Enum-like objects with numeric literal values
should also be supported, as they are a popular choice for developers
wishing to avoid use of TypeScript enums.

This PR adds support for such objects. Tests and a new example have
been added to cover this type of object.

Resolves #1918.

### DIFF
--- a/src/test/converter2/behavior/asConstEnum.ts
+++ b/src/test/converter2/behavior/asConstEnum.ts
@@ -1,3 +1,5 @@
+/* Enum-like objects with string literal values */
+
 export const SomeEnumLike = {
     a: "a",
     b: "b",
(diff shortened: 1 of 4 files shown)


src/test/converter2/behavior/asConstEnum.ts
```diff
@@ -23,3 +25,31 @@ export const ManualEnumHelper: Readonly<{ a: "a" }> = {
 export const WithoutReadonly = {
     a: "a",
 } as { a: "a" };
+
+/* Enum-like objects with numeric literal values */
+
+export const SomeEnumLikeNumeric = {
```
Gerrit0, line 31: Thanks for adding tests! One note - the converter2 tests, unlike the converter tests, do not take snapshots of everything in the test file. Instead, the converted project is passed to a method to do assertions on the converted result. In this case, in `src/test/behaviorTests.ts`, there is an `asConstEnum` function which is responsible for testing that this file gets converted correctly.

Gerrit0 (changes requested):

src/test/converter2/behavior/asConstEnum.ts
ejuda: Ah, I see - I think I fixed it now. Sorry! :)

Gerrit0 (approved):
Thanks!

PR #1938, opened 2022-05-26
fix: grammar

Zamiell:


Gerrit0:
Next time: `npm run rebuild_specs` should save some time

PR #1944, opened 2022-05-30
chore: adding ruler

Zamiell:
It's very annoying to work in this codebase and not have a ruler. Thus, this PR.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,8 @@
 {
+    // This matches the ".editorconfig" file.
+    "editor.rulers": [100],
+    "editor.tabSize": 4,
+
     "typescript.tsdk": "./node_modules/typescript/lib",
     "[typescript]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"


Gerrit0:
I'm not a fan of a bunch of noise on the screen... I think this is personal preference which should be added to your user settings, not to the repo.

Gerrit0:
If there was a rule regarding how to format comments - I'd agree! But... there isn't. There are pieces of text in this repo which have lines manually broken at roughly 80 characters (example: README.md) and others which don't bother to break paragraphs at all (example: .github/CONTRIBUTING.md)... and both are fine. So long as it isn't painful to read, you're free to wrap one comment at 120 characters, the next at 40...

Zamiell:
> If there was a rule regarding how to format comments - I'd agree!

OK, fair enough!

But consider this. Presumably, you agree that having a `max-line-len` rule set on the repository is a good idea for normal code. (I mean effectively, not literally, because Prettier takes care of this and disables the ESLint rule.)

But Prettier doesn't format comments. I think that you can take all the same arguments that you would use to justify a `max-line-len` rule for normal code, and apply them to commented code.

If developers can't be trusted to manually break lines, why should they be trusted to manually break comments?
If long lines makes things difficult to read for people with disabilities, why should comments be exempt? They will be inhibiting in the exact same way.
And so on.

So my response would be that if your codebase has a maximum length rule for code, then it follows that you should probably also have a maximum length rule for comments that are inside of the code.

Gerrit0:
Prettier exists on this repo because it prevents merge conflicts for me to figure out when people that have different formatting extensions installed work on the same file ;) In my experience, the majority of devs *can* be trusted to format code in a stylistically consistent manner to surrounding code... though I guess the exceptions to that result in the rule... haven't run into any in this project yet :)

PR #1956, opened 2022-06-23
feat: enum expressions

Zamiell:
This PR fixes the bug where TypeDoc will interpret the following code as a variable instead of an enum:

```ts
/** @enum */
export const Foo = {
  SOME_VALUE: 1 << 0,
};
```

The tests fail because one of the spec files changed.

### DIFF
--- a/src/lib/converter/symbols.ts
+++ b/src/lib/converter/symbols.ts
@@ -868,8 +868,15 @@ function isEnumLike(checker: ts.TypeChecker, type: ts.Type, location: ts.Node) {
 
     return type.getProperties().every((prop) => {
         const propType = checker.getTypeOfSymbolAtLocation(prop, location);
-        return propType.isStringLiteral() || propType.isNumberLiteral();
+        return isValidEnumProperty(propType);
     });
+}
+
+function isValidEnumProperty(type: ts.Type) {
+    return hasAnyFlag(
+        type.flags,
+        ts.TypeFlags.NumberLike | ts.TypeFlags.StringLike
+    );
 }
 
 function convertVariableAsEnum(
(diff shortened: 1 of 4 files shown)


Gerrit0:
Thanks for getting rid of the eval! I haven't had time to pull it down and play with it yet, but a few thoughts:

- I think the unknown type branch is unreachable since we can only ever enter that function if we have a string or number.
- How does this play with unions? `1 | string` should probably prevent `@enum` from triggering.
- It would be nice to add test cases in src/test/converter2/behavior/asConstEnum and the corresponding check code in src/test/behaviorTests - asConstEnum.
- Should non-literal properties be required to be read only for an object to be considered an enum? This doesn't matter with literals since there's only one value, but for non-literals... if an variable isn't immutable it makes for a bad enum.

Zamiell:
> I think the unknown type branch is unreachable since we can only ever enter that function if we have a string or number.

Yeah but I like it for exhaustiveness purposes, to ensure that no undefined behavior happens if the code is ever refactored.

> How does this play with unions? 1 | string should probably prevent @enum from triggering.

It does. I added a test to demonstrate this.
I don't like this behavior though and I think it should be changed. :)

> It would be nice to add test cases in src/test/converter2/behavior/asConstEnum and the corresponding check code in src/test/behaviorTests - asConstEnum.

Ok, I added that.

> Should non-literal properties be required to be read only for an object to be considered an enum? This doesn't matter with literals since there's only one value, but for non-literals... if an variable isn't immutable it makes for a bad enum.

Here I would rather that TypeDoc let the end-user be in control.
We shouldn't try to outsmart the end user and account for every one of their use cases.
If the end-user puts @enum, then TypeDoc should make it an enum.
(I can understand the value in preventing the user from shooting themselves in the foot in the case of e.g. a enum member with a weird type union, but in general TypeDoc should air on the side of letting the user be in complete control.)

Gerrit0:
> Yeah but I like it for exhaustiveness purposes, to ensure that no undefined behavior happens if the code is ever refactored.
I could also throw an error if you like.

Dead code is bad code, an error would be better, but not having it is probably even better. The worst case scenario here is that an enum member doesn't have a type.

> Here I would rather that TypeDoc let the end-user be in control.
We shouldn't try to outsmart the end user and account for every one of their use cases.
If the end-user puts `@enum`, then TypeDoc should make it an enum.

There's a balance here, I agree, TypeDoc isn't a linter, but documentation writers aren't the only people affected by this. It's also important to consider TypeDoc plugin developers + maintainers. An "enum" is a well defined concept in TypeScript. If a TypeDoc model claims to be an enum member but has call signatures, well... I wouldn't fault any plugin developer for not considering handling that nonsensical case.

src/lib/converter/symbols.ts
```diff
@@ -899,10 +914,24 @@ function convertVariableAsEnum(
...
+            // "1 << 1". We cannot safely evaluate the code to find out the value, so revert to
+            // printing "number" or "string", similar to what TypeScript itself does.
+            if (isNumberLike(propType)) {
+                reflection.type = new IntrinsicType("number");
+            } else if (isStringLike(propType)) {
+                reflection.type = new IntrinsicType("string");
+            } else {
+                reflection.type = new IntrinsicType("unknown");
+            }
+        }
```
Gerrit0, line 934: With the extra cases, I think it's worth calling out to `convertType` now

```suggestion
        reflection.type = context.converter.convertType(context, propType);

        if (propType.isStringLiteral() || propType.isNumberLiteral()) {
            reflection.defaultValue = JSON.stringify(propType.value);
        }
```

Gerrit0 (changes requested):
I think we can simplify the logic a bit

Gerrit0 (approved):

Gerrit0:
Thanks! Released in 0.22.18

PR #1958, opened 2022-06-24
Fix group boosts

shmax:
Not sure how we got this mixed-up, but at some point we moved the code that computes boost inside the same logical block that computes human-readable group names. Long story short, only the first of a group type was having its boost computed.


Gerrit0:
Thanks! Yeah, that's definitely an issue

shmax:
Thanks for the quick review 👍

PR #1960, opened 2022-06-25
Fix category boosts

shmax:
When walking over the child categories for a reflection, we were looking up category information in the `categories` collection, but this doesn't get written to until the end of the block, so there was no guarantee it would be found. Most straightforward fix seems to be to just use the raw category name when checking for boosts.


Gerrit0:
Yikes, that's a problem!

shmax:
Thanks for the quick merge. Hopefully that's the last of the issues... I'll do a little more testing, then return to the defaults issue when I can.

PR #1975, opened 2022-07-01
Add skipErrorChecking option

Gerrit0:
I think I'll regret this in a week, so pushing to a branch for now...


Gerrit0:
Still think I'm going to regret it, but it's coming in... will see how this goes.

PR #1983, opened 2022-07-04
fixes for schema.json

Zamiell:
there are 3 fixes in this PR:

1) properly define json files in settings.json so that they don't throw errors in vscode when you open them
2) add `"allowTrailingCommas": true` as documented here: https://github.com/microsoft/vscode/issues/102061#issuecomment-656541600
3) lowercase first letter of enum members so that they match the stylization in the docs

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,14 +1,22 @@
 {
+    "files.associations": {
+        "tsdoc.json": "jsonc", // TSDoc allows comments.
+        "typedoc.json": "jsonc" // TypeDoc uses the JSONC parser for its configuration files.
+    },
+
     "typescript.tsdk": "./node_modules/typescript/lib",
+
     "[typescript]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
     },
     "[typescriptreact]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
     },
+
     "prettier.enable": true,
     "prettier.configPath": ".config/.prettierrc.json",
     "prettier.ignorePath": ".config/.prettierignore",
+
     "eslint.workingDirectories": [".", "./example"],
     "mochaExplorer.configFile": ".config/mocha.test-explorer.json",
     "cSpell.words": ["cname", "tsbuildinfo", "tsdoc"]
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,11 +6,31 @@
 
     "typescript.tsdk": "./node_modules/typescript/lib",
 
+    // Automatically run the formatter when certain files are saved.
+    "[javascript]": {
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,14 +1,22 @@
 {
+    "files.associations": {
```
Gerrit0, line 2: Question, do you know if this is something that we could get included in VSCode's defaults?

scripts/generate_options_schema.js
```diff
@@ -72,8 +77,10 @@ addTypeDocOptions({
                     ).map;
                 data.enum =
                     map instanceof Map
-                        ? [...map.keys()]
```
Gerrit0, line 75: This won't quite do the right thing for all enums, specifically [CommentStyle](http://typedoc.org/api/enums/CommentStyle.html#JSDoc). I think grabbing the value would be a better solution.

Gerrit0 (changes requested):

.vscode/settings.json
Zamiell: i've no idea, but i would guess not.
they hard-code an exception for their own files (e.g. settings.json, tsconfig.json), but if i were microsoft i wouldn't want to maintain a hard-coded mapping of community things

scripts/generate_options_schema.js
```diff
@@ -72,8 +77,10 @@ addTypeDocOptions({
...
-                        : Object.keys(map).filter((key) => isNaN(+key));
+                        ? [...map.keys()].map(lowerCaseFirstLetter)
+                        : Object.keys(map)
+                              .filter((key) => isNaN(+key))
+                              .map(lowerCaseFirstLetter);
```
Gerrit0, line 83: I was proposing this:

```suggestion
                        ? [...map.values()]
                        : Object.keys(map)
                              .filter((key) => isNaN(+key))
                              .map((key) => map[key]);
```

... however, I did this locally, and it doesn't quite work because [LogLevel](https://typedoc.org/api/enums/LogLevel.html) is a numeric enum. `--logLevel verbose` is invalid, while `--logLevel Verbose` is valid, so we also can't just lowercase the first letter.

We could do:

```suggestion
                        ? [...map.values()]
                        : Object.keys(map)
                              .filter((key) => isNaN(+key))
                              .map((key) =>
                                  typeof map[key] === "number" ? key : map[key]
                              );
```

To use the values for non-numeric mapped types, which would end up using camel-case (sorta, is `jsdoc` considered more camel case than `JSDoc`?)

`LogLevel` could be made a non-numeric enum, but that's a breaking change, so would have to wait for 0.24.

Zamiell:
re: camelcase
doesn't matter much to me, i'll defer to your judgement

Gerrit0 (approved):

PR #1997, opened 2022-07-12
fix #1994

### LINKED ISSUE #1994: excludeNotDocumented Won't Emit the Function Declaration
## Search terms

excludeNotDocumented d.ts function Declaration

## Expected Behavior

The function declarations should be generated when `excludeNotDocumented` is `true` in the following cases: 
- Methods inside classes
- Exported function in Module. 

## Actual Behavior

- Methods inside classes
- Exported function in Module. 

The functions (call signatures) from above documents is not generated. 

## Steps to reproduce the bug

Step 1: 
`git clone https://github.com/capraynor/typedoc-method-declaration-not-extrated.git` 

Step 2:
run `npm run build-api-doc`

Regarding to this file: 
https://github.com/capraynor/typedoc-method-declaration-not-extrated/blob/master/packages2/module1/index.d.ts

case1 is not generated (not good) 
case2 is not generated (OK, because it is private one AND `excludePrivate=true`) 
case3 is not generated (not good)

Please note that: `excludeNotDocumented = true`, the bug occurs. `excludeNotDocumented=false`, everything works fine. 

## Environment

-   Typedoc version: `0.23.7`
-   TypeScript version: `4.7.4`
-   Node.js version: `v16.15.1`
-   OS: `Windows`

Will share more information if we've found a clue. 

Feel free to comment if you know which point should we check.

cary-hu:
All the existing test passed, And the PR is ready to review.


capraynor:
#1994

Gerrit0:
Thanks!

PR #2027, opened 2022-08-10
Document correct command to fix linting errors.

citkane:
As documented in `.github/CONTRIBUTING.md#linting`, a user should be able to automatically fix linting errors with `npm run lint -- --fix`

The --fix flag however causes:
- [warn] Ignored unknown option --fix.

use the --write for the correct behavior.

*note:
The linter has has re-formatted the start of 'CONTRIBUTING.md' as per this pull request.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -6,15 +6,21 @@ This document includes a set of guidelines for contributing to TypeDoc. These ar
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


Gerrit0:
Ah shoot, that's a leftover from before we switched to prettier, so there wasn't a `&& npm run prettier` in the lint command, `--fix` is still correct for eslint... don't think many people bother with eslint's fixer though, thanks!

PR #2038, opened 2022-08-22
Remove default template headers of Hierarchy and Type Parameters from DefaultThemeRenderContext when overridden.

kaphula:
If you create your own theme by extending `DefaultThemeRenderContext` and attempt to override `this.hierarchy` or `this.typeParameters` with empty content (disable them), their HTML headers get generated to the output document regardless. This PR removes the HTML headers if the default template is overridden.

```typescript
export class FooterOverrideThemeContext extends DefaultThemeRenderContext {
  constructor(theme: DefaultTheme, options: Options) {
    super(theme, options);
    // disable:
    this.hierarchy = () => { return (<></>)}
    this.typeParameters = () => { return(<></>)}
  }
}

```

### DIFF
--- a/src/lib/output/themes/default/templates/reflection.tsx
+++ b/src/lib/output/themes/default/templates/reflection.tsx
@@ -18,7 +18,6 @@ export function reflectionTemplate(context: DefaultThemeRenderContext, props: Pa
 
             {hasTypeParameters(props.model) && (
                 <section class="tsd-panel tsd-type-parameters">
-                    <h4>Type Parameters</h4>
                     {context.typeParameters(props.model.typeParameters)}
                 </section>
             )}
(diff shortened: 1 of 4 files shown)


src/lib/output/themes/default/templates/reflection.tsx
```diff
@@ -18,15 +18,13 @@ export function reflectionTemplate(context: DefaultThemeRenderContext, props: Pa
 
             {hasTypeParameters(props.model) && (
                 <section class="tsd-panel tsd-type-parameters">
```
Gerrit0, line 20: Any reason the `<section>` (and maybe even the condition) didn't move into the helpers? Seems like you'd still have elements you don't want if modifying these helpers.

Gerrit0 (approved):
Thanks!

Issue #2024, opened 2022-08-01
Default options for a custom mixed option are overridden then deleted when some, but not all, options are set in typedoc.json


citkane:
## Search terms

options addDeclaration custom

## Expected Behavior

When adding custom options with `app.options.addDeclaration`, the `defaultValue` fields will be overridden by values in `typedoc.json`, otherwise retained as per default.

## Actual Behavior
If some values are defined in `typedoc.json`, these override the defaults and the remainder are deleted.
If an empty object for the custom key is defined in `typedoc.json`, all values are deleted.
If no custom key is defined in `typedoc.json`, all values are retained as per defualt values.

## Steps to reproduce the bug
[minimal repro pr](https://github.com/TypeStrong/typedoc-repros/pull/17)
[minimal repro](https://github.com/citkane/typedoc-repros/tree/bug-options)

## Environment

-   Typedoc version: 23.10
-   TypeScript version: 4.7.4
-   Node.js version: 16.16.0
-   OS: Ubuntu 22.04

Gerrit0:
This is working as expected for mixed options. Mixed options could be objects, numbers, strings, or even a function... for this reason, TypeDoc performs no processing on them. Your plugin will need to do something like:

```ts
const opts = { ...app.options.getValue("opt"), ...defaults }
```

If your object only contains boolean flags, a Flags option type might be a better fit, TypeDoc does do some additional processing for that case.

citkane:
Thank you - your suggested workaround is appreciated.

In practice, default options will need to be read in the declaration scope before they are lost, and then merged after the reader is run;
```ts
app.options.addDeclaration({...});
const defaultOptions = app.options.getValue("opt");

// this below, or in a new scope after app.bootstrap eg. after a renderhook
app.options.addReader(new TypeDocReader());
app.options.read(new Logger());
// end

const myScopedOptions = {...defaultOptions, ...app.options.getValue("opt")}
```
From the theme / plugin developer's perspective, the API method of `app.options.getValue` is now redundant for custom keys. For me, this feels verbose and messy. 

However, given that;
- when a custom option key IS NOT passed in `typedoc.json`, the default values ARE retained, 
- when a custom option key IS passed in `typedoc.json`, the default values are ARE NOT retained

I assert that the present behavior for mixed options is inconsistent.

I expect that there are many reasons why you are implementing the behavior this way, but I argue:
- for a theme / plugin developer this level of the API should behave consistently with no workarounds required.
[...]

PR #2040, opened 2022-08-26
New BOOTSTRAP_END event and ParameterType.Object

citkane:
This addresses issue #2024 
- Creates an event hook at the end of Application.bootstrap
- Creates a new Declaration.Object type which overrides sub-keys without overwriting others.


src/lib/utils/options/declaration.ts
```diff
@@ -580,16 +581,18 @@ const converters: {
 export function convert(
     value: unknown,
     option: DeclarationOption,
-    configPath: string
+    configPath: string,
+    oldValue?: unknown
```
citkane, line 587: This is optional otherwise a whole bunch of tests need to be re-factored.

src/lib/utils/options/declaration.ts
```diff
@@ -503,6 +526,12 @@ const converters: {
         option.validate?.(value);
         return value;
     },
+    [ParameterType.Object](value, option, _configPath, oldValue) {
+        option.validate?.(value);
```
citkane, line 530: It may be worth writing a default validation to check that the value is indeed as expected or error?

There appears to be no type checking on any of the declaration type values through the abstraction layers, so the above question applies to all of them.

src/test/issueTests.ts
```diff
@@ -725,6 +728,40 @@ export const issueTests: {
         );
     },
 
+    gh2024() {
```
Gerrit0, line 731: Probably going to pull this out into options.test.ts - these tests are really for checking things which depend on conversion, otherwise we're doing more work than necessary.

src/lib/application-events.ts
```diff
@@ -0,0 +1,3 @@
+export const ApplicationEvents = {
```
Gerrit0, line 1: I'm going to end up putting a `Application.EVENT_BOOTSTRAP_END` on the `Application` class as well for discoverability.

Gerrit0 (approved):

src/lib/utils/options/declaration.ts
Gerrit0: That's a fair point, most options today silently perform a conversion... I'm not sure it matters too much since it only really affects people who write a js config file without any type checking...

citkane: I was thinking more specifically of a userland scenario where a plugin is created:
```ts
app.options.addDeclaration({
    name: 'whatever',
    type: ParameterType.Object,
    defaultValue:{...}
})
```
I can now call;  
`options.setValue('whatever', [..])`,  
`options.setValue('whatever', {..})`,  
`options.setValue('whatever', 'string')`,  
`options.setValue('whatever', true)`  
and as far as type checking and the compiler is concerned, this is all OK.

This may even be OK at runtime until some esoteric error emerges due to data corruption which will be difficult to debug. 

Also, more specific to `Object` only; according to type checking, `defaultValue:[...]`, an array, is OK in the declaration and will safely compile. This will however cause strange behavior for the [object merging](https://github.com/TypeStrong/typedoc/blob/c265735d77c8ce5432f78b1a3a4d77a0145c1d4a/src/lib/utils/options/declaration.ts#L532), which will not fail but will break data structure and cause unexpected errors elsewhere. This is why I suggested default validators, for here something like:
```ts
//for data and maybe oldData do
if(Array.isArray(data) || typeof data !=='object`) throw new Error('data is not an Object')
```

The root of the problem for `oldData` is that the [type checking on overloads](https://github.com/TypeStrong/typedoc/blob/c265735d77c8ce5432f78b1a3a4d77a0145c1d4a/src/lib/utils/options/declaration.ts#L181-L200) allows `Mixed` (and it's clone `Object`) to be an Array Object or an Object Object (curse you javascript!!!!!). I have stared at this type checking code  for hours before conceding that I just cant figure how it works...

This is not really a userland problem with `Mixed`, but is for `Object` 

This is also not a critical issue. More of a pedantic twitch...
Users who declare one datatype and then set another over it probably deserve what they get. Just hope it is not me on a hangover day!

Issue #1793, opened 2021-11-13
Pretty-print Type Aliases


kael-shipman:
## Search terms

pretty pretty-print type alias type aliases

## Expected Behavior

This might just be a setting or something that I'm totally missing. If so, fine to just point me there. I could also see this being 100% intentional behavior. That seems reasonable, given the other information on the page, but if that's the case, I'd like to convert this to a feature request. For almost all of my types, the best way to understand them is to just show the pretty-printed type definition with comments stripped out, then offer the comments further below as you've done already. It would be great for this to be an option.

Given the typescript source:

```ts
 export type MyType = {
  one: string;
  two: string;
  three: string;
  four: string;
  names: {
    me: string;
    you: string;
  }
}
```

The generated docs should show:
```ts
T MyType: {
  one: string;
  two: string;
  three: string;
  four: string;
  names: {
    me: string;
    you: string;
  }
}
```

## Actual Behavior

Given the above Typescript source, the generated docs show the following:

```ts
T MyType: { one: string; two: string; three: string; four: string; names: { me: string; you: string } }
```

## Steps to reproduce the bug

1. Clone https://github.com/kael-shipman/typedoc-bug
2. run `npm install`
3. run `npm run docgen`

## Environment

-   Typedoc version: 0.22.8
-   TypeScript version: 4.4.4
-   Node.js version: 14.18.1
-   OS: Ubuntu Studio Linux 20.04.3

Gerrit0:
This is expected behavior right now. We don't have custom rendering to try to pretty print any types. It isn't just object types either, https://typedoc.org/api/modules.html#KeyToDeclaration is less than useful... I believe this is something that might be fixed with a theme update that's a work in progress.

kael-shipman:
Thanks for the explanation! Seems totally reasonable. Feel free to close this or merge it into the theme work.

Gerrit0:
I'll leave this open until I've confirmed it's resolved :)

jgresham:
captainTorch's output in the PR looks great! 

Anyone else know of any work arounds until the PR gets released?

jgresham:
Awesome. Thanks!

PR #2067, opened 2022-10-03
Pretty print types in Default theme

captainTorch:
I faced same issue as mentioned [here](https://github.com/TypeStrong/typedoc/issues/1793) - large types looks ugly because of no line wraps or indentation.

I offer an improvement for DefaultTheme that allows to render types as formatted tree using `DefaultThemeRenderContext` as storage when recursively rendering nesting types.

Example of working:

before

after

I think this feature should be controlled with some additional configuration options (at least, use this feature or no and indent spaces quantity), but I'm not sure about your policy of adding theme-related options.


Gerrit0:
Haven't looked too closely, but; can this flag be local to the type renderer, rather than global for potential misuse by other templates?

PR #2070, opened 2022-10-07
Feature: allow plugins to return async values from getUrls

zhenyasav:
We found this necessary to support alternative templating systems which require async time to read templates from disk or otherwise do thinking to determine the URL mappings based on something async.

Because `await` on scalars is a no-op this change seems to be inert to old code.

Many thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### Features
+-   Plugins can now optionally return a promise from `getUrls` to allow async determination of url mappings
+
 ### Bug Fixes
 
 -   Private parameter properties will no longer be ignored, #2064.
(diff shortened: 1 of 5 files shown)


Gerrit0:
This seems like an inappropriate place for async code to me... reading templates is something that really shouldn't be done within a function meant to define the structure of your documentation output. The pre-render actions added by https://github.com/TypeStrong/typedoc/commit/5ddbbf69e376936bd9ef2d9ce6fa2cd0ecdca2e3 seem like a better place for that (I really need to just get the next set of breaking changes merged/released, without waiting for a large feature... not spending much time here). I'm struggling to come up with a good reason to have any async code in this function. Would the `preRenderAsyncJobs` meet your need?

PR #2078, opened 2022-10-16
Allow user-configurable links in header & sidebar

### LINKED ISSUE #1830: Add GitHub link and/or link to the project site (instead of the typedoc site), to the default theme?
## Search Terms

Would it be possible to add a GitHub link and/or a link to the actual site (instead of the typedoc site), to the default theme?

## Problem

I find for typedoc sites, I get to the typedoc site, and would like to return back to the original site, or just go to the GitHub page for the current project, and am unable to, e.g. [barba.js.org/api](https://barba.js.org/api/) (there is no button or link to the original site for barbajs, it's not a major problem, but it throws me off a bit, and I feel I'm not the only one who finds it off putting)

## Suggested Solution

Add a GitHub button and/or a original site button to the navbar of the default typedoc theme.

futurGH:
Closes #1830 

This PR introduces three options;
- `titleLink`: a string URL that the title links to; defaults to current behaviour of linking to documentation homepage
- `navigationLinks`: a `Record<string, string>` mapping labels to URLs to include in the header
- `sidebarLinks`: a `Record<string, string>` mapping labels to URLs to include in the sidebar

Naming is a bit iffy since the sidebar is internally referred to as navigation, but the label "header" is taken by the title portion of the page so this is the best I could come up with.

### DIFF
--- a/example/typedoc.json
+++ b/example/typedoc.json
@@ -10,5 +10,12 @@
     },
     "searchGroupBoosts": {
         "Classes": 1.5
+    },
+    "navigationLinks": {
+        "Docs": "https://typedoc.org/guides/overview",
+        "GitHub": "https://github.com/TypeStrong/typedoc"
+    },
+    "sidebarLinks": {
+        "API": "https://typedoc.org/api"
     }
 }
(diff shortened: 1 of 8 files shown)


Gerrit0:
Thanks!

PR #2086, opened 2022-10-27
Fix wrong handling of missing/empty titleLink option value

### LINKED ISSUE #2085: Adding a link to the home page (index.html in the destination directory)
## Search terms
index.html "home page" link

## Question
How does one provide a link to the initial index page that the readme document is rendered on? After navigating away from this, the only way back is via the back button. I have tried adding in a "navigationLinks" field with the value "index.html" but this just gets made relative to module root, so only actually works on the home page.

krisztianb:
Fixes #2085


Gerrit0:
Thanks!

PR #2102, opened 2022-11-12
Fix #2101 support externalSymbolLinkMappings wildcard

### LINKED ISSUE #2101: externalSymbolLinkMappings wildcard
## Search Terms

externalSymbolLinkMappings wildcard

## Problem

I have a library that heavily relies on three.js package. I import and expose ~20 different objects from `three`. It is not really important for me that the links for these types are accurate but at least I would like to point the [three.js documentation index](https://threejs.org/docs/).

## Suggested Solution

Allow `externalSymbolLinkMappings` to declare wildcard mapping, that would be used if no other mapping match.

```json
    "externalSymbolLinkMappings": {
      "@types/three": {
        "*": "https://threejs.org/docs/"
      },
      "three": {
        "*": "https://threejs.org/docs/"
      }
    }
```

It would also cover cases when an external library is a single long page.

mistic100:
This solves #2101 with the syntax suggested in said issue.

```json
{
    "externalSymbolLinkMappings": {
        "external-lib": {
            "SomeObject": "https://external-lib.site/docs/SomeObject",
            "*": "https://external-lib.site/docs"
        }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+### Features
+
+-   Allow to define wildcard mapping in `externalSymbolLinkMappings`.
+
 ## v0.23.20 (2022-11-03)
 
 ### Bug Fixes
--- a/internal-docs/third-party-symbols.md
+++ b/internal-docs/third-party-symbols.md
@@ -32,6 +32,20 @@ detected as belonging to the `typescript` package rather than the `global` packa
 }
 ```
 
+A wildcard can be used to provide a fallback link to any unmapped type.
+
+```jsonc
+// typedoc.json
+{
+    "externalSymbolLinkMappings": {
+        "external-lib": {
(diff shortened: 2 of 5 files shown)


mistic100:
A question I have is should we bring this further ? Like regex match ? or at URL templating ?

```json
{
  "*": "https://external-lib.site/docs/${name}"
}
```

I feel it will never cover all possible cases (In my case three.js documentation URL is categorized so only an full explicit configuration can work if I want complete links). So perhaps just keep it simple.

Gerrit0:
I think simple is good, at least initially. Maybe glob matching in the future, but I don't really see a use case for it that wouldn't be better served by writing a plugin. This option is really meant for one offs, for bigger things (linking to MDN, Node docs) a plugin is much more scalable.

Gerrit0:
This is perfect, thank you!

Issue #185, opened 2016-01-29
Async processing in plugins


meirgottlieb:
Hello,

[A plugin I wrote](https://www.npmjs.com/package/typedoc-plantuml) process PlantUML diagrams in comments and generates links to images in the comments for those diagrams. The supporting libraries that do the image generation are async. Is there a way to tell TypeDoc to wait until the async processing in the plugin has completed before continuing? Or at least before reporting that document generation has completed? As it is now it works but TypeDoc reports that it is finished before the images have actually finished generating.

Thanks!
Meir

PR #2114, opened 2022-12-09
fix: await async plugin load

achingbrain:
[5ddbbf6](https://github.com/TypeStrong/typedoc/commit/5ddbbf69e376936bd9ef2d9ce6fa2cd0ecdca2e3) added a documentation note that says plugins can return a promise from their `load` function, but that promise needs to be awaited otherwise it'll resolve outside of the normal execution flow (e.g. it could resolve after doc generation has finished) and can also cause `unhandledPromiseRejection`s.

The change here is to await the result of `initFunction`, which means making the `loadPlugins` function async and cascading that asynchronicity upwards which is why it's so noisy.

Also adds tests for loading sync/async and very slow plugins as I couldn't see where these were tested.

Refs: #185

### DIFF
--- a/bin/typedoc
+++ b/bin/typedoc
@@ -16,57 +16,97 @@ const ExitCodes = {
 const td = require("..");
 const { getOptionsHelp } = require("../dist/lib/utils/options/help");
 
-const app = new td.Application();
+async function main () {
+    const app = new td.Application();
 
-app.options.addReader(new td.ArgumentsReader(0));
-app.options.addReader(new td.TypeDocReader());
-app.options.addReader(new td.TSConfigReader());
-app.options.addReader(new td.ArgumentsReader(300));
+    app.options.addReader(new td.ArgumentsReader(0));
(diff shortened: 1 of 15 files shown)


Gerrit0:
That commit is still on the `beta` branch - hasn't been merged to master yet :)

achingbrain:
Aha, right - I thought because https://github.com/TypeStrong/typedoc/projects/10#card-68146387 was moved to "done" in July that meant it was now possible for the load function of a plugin to be async.  I guess it's still waiting a release?

If this functionality is already present you can probably close this then, though maybe the tests are worth keeping?

Gerrit0:
I agree, testing plugin load is useful, it looks like it's already tested on the beta branch, looks like I did it when adding support for ESM plugins, <https://github.com/TypeStrong/typedoc/blob/beta/src/test/utils/plugins.test.ts>

Thanks for this anyways!

PR #2129, opened 2022-12-25
Introduce an options reader for `package.json` files

### LINKED ISSUE #2112: typedoc doesn't recognize `package.json`, only `typedoc.json`
## Search terms

Running ``npm run docs`` for an npm script ``docs: typedoc src/index.ts`` does not read properties from ``package.json``:

```json
{
  // ...
  "typedocOptions": {
    "entryPoints": ["src/index.ts"],
    "out": "docs/typedocs",
    "includeVersion": true
  }
}
```

Only from ``typedoc.json``:

```json
{
   "entryPoints": ["src/index.ts"],
   "out": "docs/typedocs",
   "includeVersion": true
}
```

## Expected Behavior

read from ``package.json``

## Actual Behavior

reads only from ``typedoc.json``

## Steps to reproduce the bug

Run ``typedoc src/index.ts`` with the typedoc specs only in ``package.json``.

## Environment

-   Typedoc version: ``0.23.21``
-   TypeScript version: ``4.9.3``
-   Node.js version: ``v19.2.0``
-   OS: ``Ubuntu 20.04.5 LTS (Focal Fossa)``

jm4rtinez:
This PR introduces the capability to read TypeDoc options from `package.json` files. By default, the `PackageJsonReader` runs after the `TypeDocReader` but before the `TSConfigReader` so that users may specify a path to a `tsconfig.json` file.

Resolves #2112


src/lib/utils/options/readers/package-json.ts
```diff
@@ -0,0 +1,84 @@
...
+        const absPath = resolve(cfgPath);
+
+        // Search for `package.json` first at the provided config path, or at
+        // `cfgPath/package.json` if the provided path is a directory.
+        const path = [absPath, join(absPath, "package.json")].find(
+            (p) => fs.existsSync(p) && fs.statSync(p).isFile()
+        );
+
+        if (!path) {
+            if (container.isSet("options")) {
```
Gerrit0, line 28: This shouldn't be an error - options specified the location of typedoc.json, for test projects I frequently have several typedoc.json files. I think we should just silently do nothing here.

src/lib/utils/options/readers/package-json.ts
```diff
@@ -0,0 +1,84 @@
...
+                return;
+            }
+        } catch (err) {
+            ok(err instanceof Error);
+            this.logParsePkgFailure(logger, path);
+            return;
+        }
+
+        const opts: Record<string, any> | undefined = pkg["typedocOptions"];
+        if (opts === null || typeof opts !== "object") {
```
Gerrit0, line 54: It should not be an error if this key is not set, in that case, the user is likely configuring typedoc through another method.

Gerrit0 (changes requested):
Thanks! Couple comments

src/lib/utils/options/readers/package-json.ts
jm4rtinez: Gotcha, we'll carry on silently if a file path can't be resolved.

src/lib/utils/options/readers/package-json.ts
jm4rtinez: Okay sounds good. We'll first check if the key exists or not, and only raise an error if the key is set but the data is invalid.

jm4rtinez:
Thanks for the feedback! I've made some changes. Let me know if there's anything else.

Gerrit0 (approved):
I like it! This needs to wait for 0.24 to be merged, since it will otherwise conflict with the non-standard parsed by typedoc in packages mode from the package.json file, but it'll come in then!

PR #2133, opened 2022-12-29
tweak vscode and mocha tooling somewhat

boneskull:
The Mocha changes are just refactors/housekeeping, but the VS Code changes improve the workflow for a vscode development environment.

#### Mocha

- Sorted the JSON
- Added a `$schema` ref
- Use `require: ts-node/register` here
- Remove `require: ts-node/register` from scripts in `package.json`

#### VS Code
- Creates some build and watch tasks
- Removed invalid task
- Update `build_themes.js` to run async and support watch mode (watch mode cannot run using `esbuild.buildSync()`)

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,16 +4,81 @@
     "version": "2.0.0",
     "tasks": [
         {
-            "label": "build",
-            "type": "shell",
-            "command": "npm run build",
-            "problemMatcher": ["$tsc"]
+            "label": "build-tsc",
+            "type": "npm",
+            "script": "build:tsc",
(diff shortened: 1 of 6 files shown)


.vscode/tasks.json
```diff
@@ -4,16 +4,81 @@
     "version": "2.0.0",
     "tasks": [
         {
-            "label": "build",
+            "label": "build-tsc",
```
Gerrit0, line 7: Heh, guess you can tell how often I actually use these... I always just have `yarn build:tsc` running in a terminal.

Gerrit0 (approved):
Thanks!

PR #2134, opened 2022-12-29
fix: allow traversal of symlinks in glob

### LINKED ISSUE #2130: cannot use symlinked entry points with 'packages' strategy
## Search terms

symlink

## Expected Behavior

It should find the entry point in e.g., `node_modules/foo` when `foo` is a symlink to a package that contains a `tsconfig.json` and/or a `typedoc.entryPoint` in its `package.json`

## Actual Behavior

TypeDoc does not find the entry point.

## Steps to reproduce the bug

1. Create package `foo` w/ `typedoc.json` containing an entry point that is a symlinked package (e.g., via `npm link`).
2. Create a `tsconfig.json` and appropriate entry point in `foo`
3. Set `entryPointStrategy` to `packages`.
4. Run `typedoc --logLevel Verbose`; note that it only finds the entry point from `foo`.

## Environment

-   Typedoc version: 0.23.23
-   TypeScript version: 4.7.4
-   Node.js version: 18.12.1
-   OS: macOS

* * *

I would be willing to open a PR to address this, but I'm of the opinion that a battle-tested external library should probably be pulled in rather than add further complexity to the current implementation. What do you think?

boneskull:
Closes #2130

This adds a `followSymlinks` option to the `glob()` function, and enables it when searching for entry points.

If `true`, and a symbolic link is encountered, a stat will be taken of the symlink's target. If a dir or file, the symlink is handled like any other dir or file.  If a symbolic link itself (symlink to a symlink), we take a stat of the next until we find a dir or file, then handle it.

There is a little bit of caching to avoid extra I/O, and protection from recursive symlinks.  However, it's possible (though unlikely) that the FS could cause a "max call stack exceeded" exception.  If someone runs into this, we can change the implementation to a loop instead of a recursive function.

Apologies for blasting the `do..while`.  I love a good `do` myself, but splitting out the lambda functions make it untenable.

### DIFF
--- a/src/lib/utils/entry-point.ts
+++ b/src/lib/utils/entry-point.ts
@@ -203,7 +203,7 @@ export function getExpandedEntryPointsForPaths(
 function expandGlobs(inputFiles: string[]) {
     const base = getCommonDirectory(inputFiles);
     const result = inputFiles.flatMap((entry) =>
-        glob(entry, base, { includeDirectories: true })
+        glob(entry, base, { includeDirectories: true, followSymlinks: true })
     );
     return result;
 }
--- a/src/lib/utils/fs.ts
+++ b/src/lib/utils/fs.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs";
 import { promises as fsp } from "fs";
 import { Minimatch } from "minimatch";
-import { dirname, join } from "path";
+import { dirname, join, relative } from "path";
 
 /**
  * Get the longest directory path common to all files.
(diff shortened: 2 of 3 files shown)


boneskull:
I should also note that I'm unsure how well `glob()` works on Windows.  I didn't feel confident that swapping `"/"` for `path.sep` was a good idea though.

Gerrit0:
> I didn't feel confident that swapping "/" for path.sep was a good idea though.

Your intuition was correct - minimatch expects paths to be joined with `/`. I'll test this on a Windows box before merging.

boneskull:
I confirmed it myself: this doesn't work on Windows. 😄

Will get it working

boneskull:
(FWIW, the "glob handles root match" test fails on Windows in `master`; I'll fix that too)

boneskull:
OK, I have the tests passing in Windows (Win 10, PowerShell).  The code itself was fine, but the tests made poor assumptions.

The test fix I made for an unrelated failure on Windows (6c1b4ac) is dubious and I think it'd necessitate a PR into [typestrong/fs-fixture-builder](/typestrong/fs-fixture-builder) to correct the underlying problem.  What I was seeing was that `fix.cwd` contained _both_ POSIX and Win32 path separators.  Likely there's a `path.resolve()` or `path.join()` in there somewhere that needs to be `path.posix.x()` to jive with minimatch

Gerrit0:
That's fun, the normalize is a good idea anyways - I'd rather not care what separators `fix.cwd` uses, I bet tests have been broken on Windows for a while, haven't booted it up to test for a few months. Will get to this today :)

Gerrit0:
Thanks!

PR #2163, opened 2023-02-06
Pretty print types improvements

captain-torch:
Addition for PR [Pretty print types in Default theme](https://github.com/TypeStrong/typedoc/pull/2067)

I found that in some cases (for example, in rendering union of types) depth counter used to work incorrect.

This PR closes this issue.


Gerrit0:
The depth counter worked incorrectly because it wasn't always incremented and decremented in pairs - I found and fixed this before releasing TypeDoc with your original PR. The `Math.max` method hides the issue rather than fully fixing it, you shouldn't see this problem if you update your TypeDoc (fork?)

PR #2164, opened 2023-02-08
fix(fs): glob will traverse node_modules if asked to

### LINKED ISSUE #2151: glob implementation should not ignore node_modules
## Search terms

node_modules glob

## Expected Behavior

I expected to be able to use an entry point located within a `node_modules` folder using the `packages` strategy, as suggested in [this comment by @Gerrit0](https://github.com/TypeStrong/typedoc/issues/2075#issuecomment-1279987587).

## Actual Behavior

If TypeDoc's glob implementation encounters a `node_modules` dir anywhere within a path, it is [explicitly ignored](https://github.com/typestrong/typedoc/blob/0340ef455303fc3287deb1c1776f7b224cd89ef0/src/lib/utils/fs.ts#L219).

## Steps to reproduce the bug

1. Create a `typedoc.json` with an entry point in `node_modules`
2. Run `typedoc`
3. Failure: no entry points found

Alternatively:

1. Create a `typedoc.json` with an entry point in `node_modules` and another (known good entry point) _not_ in a `node_modules`
2. Run `typedoc`
3. Notice that the entry point in `node_modules` is excluded from the build

## Environment

-   Typedoc version: v0.23.24
-   TypeScript version: v4.7.4 / v4.9.4
-   Node.js version: v18.x
-   OS: macOS

* * *

A potential fix is to remove the explicit check for `node_modules` from the `glob` implementation.  As it's unclear to me _why it's present_, this may have unintended consequences. Removing it does not fail the test suite.  Another route would be to support an `ignoreNodeModules` option for `glob` -- defaulting to `true` -- and be set to `false` when computing entry points. 

My workaround right now is a monkeypatch which bypasses Node's subpath exports restrictions:

```js
import {sync as glob} from 'glob';

require(path.join(
    somePackageRoot,
    'node_modules',
    'typedoc',
    'dist',
    'lib',
    'utils',
    'fs.js'
)).glob = glob;
```

boneskull:
If the `pattern` passed to `glob()` contains `node_modules`, then `node_modules` will be traversed.

This is _naive_ in that it does not care about the _context_ of `node_modules` in the pattern; it could be a negation, or whatever (e.g., some passes `!**/node_modules/**`).  However, `minimatch` should still do its thing, so hopefully this won't be too bad.

Closes #2151

### DIFF
--- a/src/lib/utils/fs.ts
+++ b/src/lib/utils/fs.ts
@@ -149,7 +149,11 @@ export function glob(
     // cache of fs.realpathSync results to avoid extra I/O
     const realpathCache: Map<string, string> = new Map();
     const { includeDirectories = false, followSymlinks = false } = options;
-
+    // if we _specifically asked_ for something in node_modules, fine, otherwise ignore it
+    // to avoid globs like '**/*.ts' finding all the .d.ts files in node_modules.
+    // however, if the pattern is something like `!**/node_modules/**`, this will also
+    // cause node_modules to be considered, though it will be discarded by minimatch.
+    const shouldIncludeNodeModules = pattern.includes("node_modules");
     let dir = dirs.shift();
 
     const handleFile = (path: string) => {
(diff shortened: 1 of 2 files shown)


boneskull:
@Dayday10 I don't see any suggested changes..? weird.

Gerrit0 (approved):
I like it, thanks for adding tests! I'll look at getting this merged & released this weekend.

Gerrit0:
Thanks! Should get released today.

PR #2184, opened 2023-02-24
feat: add validator run event to allow plugins to perform custom validation

### LINKED ISSUE #2183: Custom validation
## Search Terms

validate
validation

## Problem

I want to create a custom validation check but that doesn't currently seem possible.

## Suggested Solution

Allow plugins access to an "on validate" event. They can then run their own logic and return either `ture` or `false`.

RebeccaStevens:
fix #2183

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -111,7 +111,7 @@ export class Application extends ChildableComponent<
      * Emitted when validation is being run.
      * The listener will be given an instance of {@link ProjectReflection}.
      */
-    static readonly EVENT_VALIDATION_RUN = ApplicationEvents.VALIDATION_RUN;
+    static readonly EVENT_VALIDATE_PROJECT = ApplicationEvents.VALIDATE_PROJECT;
 
     /**
      * Create a new TypeDoc application instance.
@@ -437,7 +437,7 @@ export class Application extends ChildableComponent<
             validateLinks(project, this.logger);
         }
 
-        this.trigger(Application.EVENT_VALIDATION_RUN, project);
+        this.trigger(Application.EVENT_VALIDATE_PROJECT, project);
 
         this.logger.verbose(`Validation took ${Date.now() - start}ms`);
     }
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -108,6 +108,12 @@ export class Application extends ChildableComponent<
     static readonly EVENT_BOOTSTRAP_END = ApplicationEvents.BOOTSTRAP_END;
 
     /**
+     * Emitted when validation is being run.
+     * The listener will be given an instance of {@link ProjectReflection}.
+     */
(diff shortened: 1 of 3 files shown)


Gerrit0:
I don't think I want the validator to be a separate thing - it's not nearly complicated enough to deserve that... I'd rather just add the validate event to `ApplicationEvents` and emit it on the application, then if any validator logs or listener to that method logs an error, TypeDoc will exit with the ValidationError exit code.

If it does deserve pulling out into its own class someday - I'd still want to avoid the component and emit that event from the application I think... I'm not convinced that there's value for plugins in seeing a new class.

RebeccaStevens:
@Gerrit0 I've reworked this PR to address those issue :)

RebeccaStevens:
When working on my plugin that relies on this PR, `reflection.comment.getTags("@param")` always return an empty array. I'm not having this issue with any other tags. Am I missing something for this tag or is this a bug?

src/lib/utils/events.ts
```diff
@@ -433,6 +436,12 @@ export class EventDispatcher {
         context?: any,
         priority?: number
     ): this;
+    on(
```
Gerrit0, line 439: I'd rather not do this for now - better to have none of the events typed than some of them. Eventually it'd be nice to have autocomplete, but that'll probably not come until removing `@Component`...

src/lib/application.ts
```diff
@@ -107,6 +107,12 @@ export class Application extends ChildableComponent<
      */
     static readonly EVENT_BOOTSTRAP_END = ApplicationEvents.BOOTSTRAP_END;
 
+    /**
+     * Emitted when validation is being run.
+     * The listener will be given an instance of {@link ProjectReflection}.
+     */
+    static readonly EVENT_VALIDATION_RUN = ApplicationEvents.VALIDATION_RUN;
```
Gerrit0, line 114: I think `validateProject` might be better - it immediately implies that the callback will be passed a project, thoughts?

RebeccaStevens: So `ApplicationEvents.VALIDATE_PROJECT`?

Gerrit0: Yep, and `validateProject` for the string, otherwise lgtm

RebeccaStevens: Thanks all set up :)

Gerrit0:
Thanks! Should release today.

Gerrit0:
> When working on my plugin that relies on this PR, reflection.comment.getTags("@param") always return an empty array. I'm not having this issue with any other tags. Am I missing something for this tag or is this a bug?

Just realized I forgot to answer this -- `@param` tags are moved to the `ParameterReflection` instances, so you generally shouldn't have any `@param` tags, your plugin should probably check that `signatureReflection.parameters` all have comments.

PR #2194, opened 2023-03-08
Fix sidebar margins on mobile devices

### LINKED ISSUE #2191: Rendering of the navigation pane on iPad is shifted down
I noticed a small problem with rendering the generated site on an iPad: the navigation pane seems shifted down and the last link is not visible:

IMG_0253

When rendered with Safari on macOS, it is ok:

Screenshot 2023-03-06 at 17 27 20

The [site](https://xpack.github.io/logger-ts/) was generated with 0.23.25.

futurGH:
Fixes #2191.

The margin at the top of the sidebar is caused when the display is precisely 1024px wide, because of overlapping min-width and max-width media queries. Also switched to flex for the mobile sidebar to prevent unnecessary gaps.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -483,10 +483,9 @@ blockquote {
     .has-menu .col-menu {
         visibility: visible;
         transform: translate(0, 0);
-        display: grid;
-        align-items: center;
-        grid-template-rows: auto 1fr;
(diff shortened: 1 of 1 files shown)


Gerrit0:
Thanks!

PR #2212, opened 2023-03-27
the --includeVersion flag now checks if the prefix "v" already exists…

muratgozel:
… in the version field of the package to prevent duplicate v's in the project name.


Gerrit0:
Thanks!

PR #2254, opened 2023-04-22
Fix Indentation of `const script = ...`

FlippieCoetser:
Improve the readability of the documentation by indenting the `const script = ...`

### DIFF
--- a/internal-docs/custom-themes.md
+++ b/internal-docs/custom-themes.md
@@ -21,7 +21,7 @@ Google Analytics. This could be done with the following theme:
 ```tsx
 import { Application, DefaultTheme, PageEvent, JSX } from "typedoc";
 
-const openWebAnalyticsScript = `
+const script = `
 (function() {
     var _owa = document.createElement('script'); _owa.type = 'text/javascript';
     _owa.async = true; _owa.src = '${site}' + '/modules/base/js/owa.tracker-combined-min.js';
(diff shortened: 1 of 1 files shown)


internal-docs/custom-themes.md
```diff
@@ -31,13 +31,13 @@ class MyThemeContext extends DefaultThemeRenderContext {
...
 
         const script = `
-(function() {
-    var _owa = document.createElement('script'); _owa.type = 'text/javascript';
-    _owa.async = true; _owa.src = '${site}' + '/modules/base/js/owa.tracker-combined-min.js';
-    var _owa_s = document.getElementsByTagName('script')[0]; _owa_s.parentNode.insertBefore(_owa,
-    _owa_s);
-}());
-`.trim();
+            (function() {
```
Gerrit0, line 34: The indentation here was intentional, the extra whitespace added here will be included in the output, which likely isn't what is desired. Perhaps moving the `const script` outside of the class so that it is naturally non-indented would be better?

Gerrit0 (changes requested):

internal-docs/custom-themes.md
FlippieCoetser: As suggested, I moved the `const script` outside the class. Not sure if the name is suitable.

FlippieCoetser: Moved the script outside of class

Gerrit0:
Thanks!

PR #2309, opened 2023-06-10
feat: add extra entry-point related paths on reflections

GerkinDev:
This PR adds extra informations about the entrypoint on declarations & projects.

As a plugin author, I would use those informations to read some files relative to each entrypoint. I found a hack to do this with 0.23.x, but it does not work for 0.24.x. So, the simplest solution I found was to add the required informations directly in TypeDoc.

Note that this inability to bring this hack from 0.23.x to 0.24.x prevents me from releasing versions of [@knodes/typedoc-plugin-pages](https://www.npmjs.com/package/@knodes/typedoc-plugin-pages) & [@knodes/typedoc-plugin-code-blocks](https://www.npmjs.com/package/@knodes/typedoc-plugin-code-blocks) compatible with 0.24.x onwards.

Feedbacks are welcome. This is a quick approximate draft, that I will continue by adding more precise tests if you'll be susceptible to accept such addition.

Thanks !

### DIFF
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -4,6 +4,7 @@ import type { Application } from "../application";
 import {
     Comment,
     CommentDisplayPart,
+    EntrypointInfos,
     ProjectReflection,
     Reflection,
     ReflectionKind,
@@ -368,6 +369,7 @@ export class Converter extends ChildableComponent<
             // Special case for when we're giving a single entry point, we don't need to
             // create modules for each entry. Register the project as this module.
             context.project.registerReflection(context.project, symbol);
+            context.project.entrypointInfos = EntrypointInfos.fromDocumentationEntrypoint(entryPoint);
             context.project.comment = symbol
                 ? context.getComment(symbol, context.project.kind)
                 : context.getFileComment(node);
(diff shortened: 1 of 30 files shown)


Gerrit0:
Is there a reason to do this for _every_ reflection, rather than just the top level module created for an entry point? It feels like a lot of redundant information to include it in all declaration reflections. Including `rootDir` is also really suspicious to me, particularly when the paths included in the serialized output appear to also repeat the rootDir info, and the root directory is really a project level attribute, not a reflection level one.

juozasmiskinis:
Any chance we could revive this discussion? Typedoc is currently the only maintained solution to generate documentation from Typescript comments and having the option to add custom pages would be very useful.

treardon17:
I'm also interested in this feature 👍

blaiprat:
👍 Also interested into having this functionality added

Gerrit0:
Based on the use cases mentioned in the description, I believe TypeDoc 0.26's [`Comment.sourcePath`](https://github.com/TypeStrong/typedoc/blob/84e4b858f5f7a66c6a11be20109d3bceb525f263/src/lib/models/comments/comment.ts#L398-L405) attribute meets this need.

Note: This attribute is not serialized, as it includes the absolute path on a machine, and JSON files are intended to be suitable for use in areas without that being available. It's also **not** the path of the reflection's declaration location, but the path to the file where the comment exists. This can be different if a comment was inherited from a base class defined in another file.

Gerrit0:
0.26 is now out, as nobody has told me that I'm wrong about the comment paths meeting this need... I'm going to assume they do.

PR #2370, opened 2023-08-23
feat(default-value): treat as code block for better styling

roggervalf:
as default tag, I would like to see defaultValue tag with the same styling


Gerrit0:
This needs to only be done if the jsdoc compat option is turned on, but I can fix that up, thanks!

Issue #2373, opened 2023-08-25
Support typescript 5.2.x


Th3S4mur41:
https://www.npmjs.com/package/typescript/v/5.2.2

Probably need to double check https://github.com/TypeStrong/typedoc/issues/2369

TechQuery:
Yes, crashed:

```text
TypeDoc exiting with unexpected error:
TypeError: Cannot read properties of undefined (reading 'name')
    at D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\types.js:564:100
    at Array.map (<anonymous>)
    at Object.convertType (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\types.js:564:34)
    at convertType (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\types.js:106:34)
    at Converter.convertType (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\converter.js:91:40)
    at Object.convertProperty (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\symbols.js:395:41)
    at convertSymbol (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\symbols.js:135:41)
    at convertSymbols (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\symbols.js:141:9)
    at Object.convertClassOrInterface (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\symbols.js:331:5)
    at convertSymbol (D:\Work\idea2app\Idea-React\node_modules\.pnpm\typedoc@0.24.8_typescript@5.2.2\node_modules\typedoc\dist\lib\converter\symbols.js:135:41)
[...]

PR #2374, opened 2023-08-25
Added support for TypeScript 5.2

Th3S4mur41:
This PR adds support for TypeScript 5.2

Closes https://github.com/TypeStrong/typedoc/issues/2373


li-jia-nan (approved):

Gerrit0:
This bumps the version, but doesn't actually support it. If you try to use TS 5.2 features, it'll crash.

PR #2379, opened 2023-08-28
Do not exit when in watch mode

### LINKED ISSUE #2378: Watch mode not working after upgrade to 0.25.0
## Search terms

Watch mode

## Expected Behavior

Use TypeScript's incremental compiler to watch source files for changes and build the docs on change.

## Actual Behavior

Does not create docs in the watch mode and exits prematurely.

## Steps to reproduce the bug

1. Upgraded to `typedoc@0.25.0` to allow support for `typescript@5.2.2`.
2. Checkout https://github.com/HemalPatil/td-bug
3. Do `npm install` and then `npm run watch:docs`.
4. Expected to see docs generated in watch mode like it did with `typedoc@0.24.8`.
5. If you change the typedoc version back to 0.24.8 in the package.json in sample repo and install the dependencies (may have to do force install because of missing typescript peer dependency), this feature works fine.

Have the configs for watch mode changed? Seems unlikely after reading the official TypeDoc docs.

## Environment

-   Typedoc version: 0.25.0
-   TypeScript version: 5.2.2
-   Node.js version: 18.17.1
-   OS: Ubuntu 22.04 and MacOS 13

HemalPatil:
Resolves #2378

### DIFF
--- a/src/lib/cli.ts
+++ b/src/lib/cli.ts
@@ -7,6 +7,7 @@ const ExitCodes = {
     ValidationError: 4,
     OutputError: 5,
     ExceptionThrown: 6,
+    Watching: 7,
 };
 
 import * as td from "typedoc";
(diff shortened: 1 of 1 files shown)


HemalPatil:
@Dayday10 the details are added in Issue #2378 

Before the commit [Support ESM config files](https://github.com/TypeStrong/typedoc/commit/5c977aea2e781bf8a97f02b59862df707f966d7b), the CLI simply used to set the process exitCode, even in watch mode, and let the process terminate eventually. But now `process.exit()` is called instead which terminates typedoc before it can finish writing the docs.

I added the `Watching` exit code so the CLI knows that it shouldn't exit and let tsc continue in watch mode.

Gerrit0 (approved):

PR #2387, opened 2023-09-03
feat: `stripYamlFrontmatter` option

### LINKED ISSUE #2381: Bug: Frontmatter not removed in Markdown Readme
## Search terms

markdown, frontmatter, readme

## Expected Behavior

The markdown parser in typedoc (for readmes) removes any yaml frontmatter as it is just metadata and should not be displayed to the end-user.

## Actual Behavior

YAML frontmatter is converted to html and shown. Example:
```md
---
title: Vision
---

# SHIFTY - Vision

```
becomes
grafik

However, we should just see the headline and not `title: Vision`.

## Steps to reproduce the bug

1. Create a readme with YAML frontmatter (like above)
2. Build the docs
3. See the frontmatter contents in the generated docs

## Environment

-   Typedoc version: 0.24.8
-   TypeScript version: 5.0.0
-   Node.js version: 18.17.1
-   OS: Windows & macOS

## Additional Info
Happy to fix that in a PR. Should be as simple as removing frontmatter via regex before passing it to `marked`. Maybe this should be a new config option?

hrueger:
fixes #2381

I really hope this is how it was intended to be implemented and I didn't miss anything. If not or if there are any other things that need to be fixed, just let me know 👍

### DIFF
--- a/src/lib/converter/plugins/PackagePlugin.ts
+++ b/src/lib/converter/plugins/PackagePlugin.ts
@@ -133,7 +133,7 @@ export class PackagePlugin extends ConverterComponent {
     private processReadmeContents(contents: string) {
         if (this.stripYamlFrontmatter) {
             return contents.replace(
-                /^[\r\n\s]*---\r?\n[\s\S]*?\r?\n---[\r\n\s]+/,
+                /^\s*---\r?\n[\s\S]*?\r?\n---\s*?\r?\n\s*/,
                 "",
             );
         }
--- a/src/lib/converter/plugins/PackagePlugin.ts
+++ b/src/lib/converter/plugins/PackagePlugin.ts
@@ -24,6 +24,9 @@ export class PackagePlugin extends ConverterComponent {
     @BindOption("readme")
     readme!: string;
 
+    @BindOption("stripYamlFrontmatter")
+    stripYamlFrontmatter!: boolean;
+
     @BindOption("entryPointStrategy")
     entryPointStrategy!: EntryPointStrategy;
 
@@ -99,7 +102,9 @@ export class PackagePlugin extends ConverterComponent {
         if (this.readme) {
             // Readme path provided, read only that file.
             try {
-                this.readmeContents = readFile(this.readme);
+                this.readmeContents = this.processReadmeContents(
+                    readFile(this.readme),
+                );
                 this.readmeFile = this.readme;
             } catch {
                 this.application.logger.error(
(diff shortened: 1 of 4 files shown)


src/lib/converter/plugins/PackagePlugin.ts
```diff
@@ -118,11 +123,23 @@ export class PackagePlugin extends ConverterComponent {
...
+                    result.content,
+                );
             }
         }
     }
 
+    private processReadmeContents(contents: string) {
+        if (this.stripYamlFrontmatter) {
+            return contents.replace(
+                /^[\r\n\s]*---\r?\n[\s\S]*?\r?\n---[\r\n\s]+/,
```
Gerrit0, line 136: Unless I'm missing something, isn't this equivalent to `/^\s*---\r?\n[\s\S]*?\r?\n---\s+/`?

This looks like it'll also catch the following, which I don't think it should?

```md
---
x
--- text here
```

hrueger: > Unless I'm missing something, isn't this equivalent to /^\s*---\r?\n[\s\S]*?\r?\n---\s+/?

Oh yeah, you're right. I'm not a specialist when it comes to regexes, I missed that whitespaces include newlines ;-)

> This looks like it'll also catch the following, which I don't think it should?

Good catch. GitHub does not render that as frontmatter. What do you think about this regex?

```
^\s*---\r?\n[\s\S]*?\r?\n---\s+\r?\n
```

Gerrit0: How about this?

```
^\s*---\r?\n[\s\S]*?\r?\n---\r?\n\s*
```

Having the newlines after required whitespace seems weird, I'm pretty sure we just care that `---` appears on a line by itself for closing frontmatter, then want to strip any leading whitespace.

hrueger: Hm, this does not match the frontmatter at all if there's a space after the last `---` (before the linebreak). Or do you think that this is out of spec?

Gerrit0: Ah, good catch, an extra `\s*?` there perhaps?

hrueger: Sorry for the delay. Why `*?`? I might be wrong, but I thought that `*` means `0 or more of the preceding token` making the `?` obsolete.

`^\s*---\r?\n[\s\S]*?\r?\n---\s+\r?\n`?
That is almost the one I posted in the first comment, just with a `*` instead of a `+` after the `---\s`.

Gerrit0: Yes, `*` means that, `*?` means do that same thing, but make the match non-greedy so that `\r?\n` matches earlier in the string. It shouldn't actually matter in this case

hrueger: Ah, makes sense. Didn't think about that. So `^\s*---\r?\n[\s\S]*?\r?\n---\s*?\r?\n\s*`?

Gerrit0: LGTM!

hrueger: Done

Gerrit0 (approved):

hrueger:
Although github status says everything is fine, I've seen about 15 workflow runs already fail today. I believe this has nothing to do with this PR, let's just wait a bit and run again.

PR #2400, opened 2023-09-15
Handling for multi word searches

### LINKED ISSUE #2399: Add wildcards to each word in a multi-word search
## Search Terms

search index multi-word

## Problem

The current search implementation wraps the entire search query in a wildcard, e.g. a search for `foo` will become `*foo*`.  This works well with single word queries, but does not work particularly well with multi-word queries.  For example, given the following:

```
class SerialConnection {
    ...
}
```

Searching for `serial` or `connection` will correctly find that class, but searching for `serial connection` will give no results.  This is because the actual query passed into Lunr is `*serial connection*` .  There are times when I'm searching documentation and I can only remember multiple snippets of a particular function or class, and it would be nice to be able to narrow the results by providing as many separate terms as I remember.

schiem:
Split searches on space and wraps each part in a wildcard.  This replaces the current current behavior of wrapping the entire search query in a wildcard.

Current behavior:

-     "create" becomes "*create*"
-     "create item" becomes "*create item*"

New behavior:

-     "create" becomes "*create*"
-     "create item" becomes "*create* *item*"

Resolves #2399

### DIFF
--- a/src/lib/output/themes/default/assets/typedoc/components/Search.ts
+++ b/src/lib/output/themes/default/assets/typedoc/components/Search.ts
@@ -157,9 +157,22 @@ function updateResults(
     const searchText = query.value.trim();
 
     // Perform a wildcard search
-    // Set empty `res` to prevent getting random results with wildcard search
-    // when the `searchText` is empty.
-    let res = searchText ? state.index.search(`*${searchText}*`) : [];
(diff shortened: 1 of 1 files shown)


Gerrit0:
Seems reasonable, thanks!

Issue #2111, opened 2022-12-08
Support for custom output directory structure


NullVoxPopuli:
## Search terms

Cloudflare pages

## Expected Behavior

Visiting a link for a `function` works
## Actual Behavior

Cloudflare pages  reverts to the `/index.html`
because `/functions` is used for Cloudflare-specific stuff.

## Steps to reproduce the bug

1. Deploy any typedoc output to Cloudflare pages.
2. Click any `function` on any of the docs pages.

Example: 

I have this output:
```
Permissions Size User          Date Modified Name
.rw-rw-r--   143 nullvoxpopuli  8 Dec 13:23  .nojekyll
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  assets
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  classes
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  functions
.rw-rw-r--   23k nullvoxpopuli  8 Dec 13:23  index.html
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  interfaces
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  modules
.rw-rw-r--  9.1k nullvoxpopuli  8 Dec 13:23  modules.html
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  types
drwxrwxr-x     - nullvoxpopuli  8 Dec 13:23  variables
```

Which is deployed to: ember-resources.pages.dev/
On _any_ page, such as: https://ember-resources.pages.dev/modules/util_cell
I click the available `functions`, such as: https://ember-resources.pages.dev/functions/util_cell.cell.html, but I am redirected to a poorly rendered version of https://ember-resources.pages.dev

My config:
```
{
  "tsconfig": "../ember-resources/tsconfig.json",
  "entryPoints": [
    "../ember-resources/src/index.ts",
    "../ember-resources/src/util/"
  ],
  "entryPointStrategy": "expand",
  "readme": "../README.md",
  "exclude": [],
  "json": "./docs.json",
  "out": "dist",
  "pretty": true,
  "excludePrivate": true,
  "excludeProtected": false,
  "excludeExternals": true,
  "disableSources": true,
  "searchInComments": true,
  "categorizeByGroup": false,
  "plugin": [
  ]
}
```

## Environment
```
typedoc 0.23.14
typescript 4.8.3

❯ node -v
v16.15.1

❯ lsb_release -a
[...]

PR #2407, opened 2023-09-30
feat: Make `mappings` and `getMapping` public

NikolaRHristov:
Allows for the overriding of `mappings`, so that they can be overriden by child themes that extend `DefaultTheme`.

This will resolve current issues such as https://github.com/TypeStrong/typedoc/issues/2111 with the creation of a simple theme like in https://github.com/NikolaRHristov/TypeScriptESBuild/blob/main/Source/Class/Theme.ts

### DIFF
--- a/src/lib/output/themes/default/DefaultTheme.tsx
+++ b/src/lib/output/themes/default/DefaultTheme.tsx
@@ -77,7 +77,7 @@ export class DefaultTheme extends Theme {
     /**
      * Mappings of reflections kinds to templates used by this theme.
      */
-    private mappings: TemplateMapping[] = [
+    mappings: TemplateMapping[] = [
         {
             kind: [ReflectionKind.Class],
             directory: "classes",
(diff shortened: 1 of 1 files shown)


NikolaRHristov:
I just noticed I can override [buildUrls](https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/themes/default/DefaultTheme.tsx#L199), nevermind. 😁 Still useful though. Will leave it open if you want to merge.

Gerrit0:
I'm not sure I want to merge this -- I've never liked how the URL building was implemented, and making this part of the public API seems like a step in the wrong direction...

I've been experimenting with alternative generation schemes on the beta branch, but am not fully happy with it yet

NikolaRHristov:
Should I close the PR?

Gerrit0:
Feel free to leave it open for now, if I vent get the rework into a state I'm happy with, I'll reconsider merging :)

Issue #1135, opened 2019-11-14
header id conflicts


egoist:
- [x] I have checked [issues with enhancement label](https://github.com/TypeStrong/typedoc/labels/enhancement) and found no duplicates

## Problem

Sometimes the IDs in `README.md` are conflicted with the IDs generated from types.

## Suggested Solution

Maybe add prefix to the ID generated from types?

Gerrit0:
I may be missing something, but when does this actually happen? At least in the default theme, it looks like we don't render anything with anchors but the readme on the page that displays it.

https://typedoc.org/api/

egoist:
> it looks like we don't render anything with anchors but the readme on the page that displays it.

That's exactly what I mean (https://majo.egoist.sh):

![image](https://user-images.githubusercontent.com/8784712/68932788-46297180-07ce-11ea-9959-3f0e202a5f02.png)

These two things use the same ID, you may add a prefix to either one of them.

Gerrit0:
Thanks, I didn't consider the minimal theme. PR welcome to add a prefix, I think it would be better to prefix the readme.

The whole anchor / url scheme needs reworking (two variables `Foo` and `foo` will have duplicate IDs), but that's a much larger task

PR #2413, opened 2023-10-19
Prepend markdown anchor links with '#md:'

mogelbrod:
Typedoc prepends `md:` to heading anchor ids generated by marked (added to fix https://github.com/TypeStrong/typedoc/issues/1135), but currently doesn't update any existing anchor links that points to them. This PR should address that.

### DIFF
--- a/src/lib/output/themes/MarkedPlugin.ts
+++ b/src/lib/output/themes/MarkedPlugin.ts
@@ -195,6 +195,19 @@ output file :
         if (!markedOptions.renderer) {
             markedOptions.renderer = new Marked.Renderer();
 
+            markedOptions.renderer.link = (href, title, text) => {
+                // Prefix the #anchor links `#md:`.
+                href =
+                    href
+                        ?.replace(/^#(?:md:)?(.+)/, "#md:$1")
+                        .replace(/"/g, "&quot;") || "";
+                let html = `<a href="${href}"`;
+                if (title != null)
+                    html += ` title="${title.replace(/"/g, "&quot;")}"`;
+                html += `>${text}</a>`;
+                return html;
+            };
+
             markedOptions.renderer.heading = (text, level, _, slugger) => {
                 const slug = slugger.slug(text);
                 // Prefix the slug with an extra `md:` to prevent conflicts with TypeDoc's anchors.


src/lib/output/themes/MarkedPlugin.ts
```diff
@@ -195,6 +195,19 @@ output file :
...
+            markedOptions.renderer.link = (href, title, text) => {
+                // Prefix the #anchor links `#md:`.
+                href =
+                    href
+                        ?.replace(/^#(?:md:)?(.+)/, "#md:$1")
+                        .replace(/"/g, "&quot;") || "";
+                let html = `<a href="${href}"`;
+                if (title != null)
+                    html += ` title="${title.replace(/"/g, "&quot;")}"`;
+                html += `>${text}</a>`;
```
mogelbrod, line 207: Awkward to have to construct the HTML tag manually - not sure if marked or typedoc provides an API for this?

Gerrit0: marked doesn't, typedoc does via the jsx templating engine, I'll make that change after merging

Gerrit0: Actually... I think we can revert this annoying prefixing now! The minimal theme was removed a while ago, so I'm not sure there's a reason for it anymore...

Gerrit0: ... sadly not. I'd like to rework the renderer at some point to make this possible, but headings in comments still get linked. It feels like there ought to be some other non-terrible way to do this, that doesn't make us write different headings than npm/github does when rendering...

Gerrit0:
Thanks!

src/lib/output/themes/MarkedPlugin.ts
mogelbrod: Perhaps going the other direction would work; adding prefixes to the headings which typedoc "owns" (generates) and leaving the regular markdown ids unprefixed?

Gerrit0: I considered that when first adding the prefix, but typedoc generates a *lot* of anchors, there are generally far, far fewer markdown header links.

PR #2418, opened 2023-10-23
Add internal-first sort strategy

rsanchez:
I have a use case for displaying internal properties before external/inherited properties. This PR adds `internal-first` as a sort strategy.

Companion documentation PR: https://github.com/TypeStrong/typedoc-site/pull/68

### DIFF
--- a/src/lib/utils/sort.ts
+++ b/src/lib/utils/sort.ts
@@ -19,6 +19,7 @@ export const SORT_STRATEGIES = [
     "visibility",
     "required-first",
     "kind",
+    "internal-first",
 ] as const;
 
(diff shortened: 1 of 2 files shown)


Gerrit0:
Thanks for the PR! I have some concerns with the naming -- "internal" is not the same as "non-external", the function you've added should really be called `external-last`, an `internal-first` name would sort reflections marked with `@internal` before those without that annotation.

rsanchez:
@Gerrit0 Thanks for the feedback. I agree that `external-last` is a more apt name. I've updated this PR with the new name and also updated companion PR for the docs site.

Gerrit0:
Perfect, thanks!

PR #2433, opened 2023-10-31
Fix declared flags option

ocavue:
When setting an option with the type of `ParameterType.Flags`, if `this._values[declaration.name]` doesn't exist yet, TypeDoc will throw `[error] Cannot convert undefined or null to object`. This PR fixes it by assigning an empty object when it is undefined.

This PR also improves the logging by adding some extra information about the current key when `setValue` method throws an error.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -227,7 +227,7 @@ export class Application extends ChildableComponent<
             } catch (error) {
                 ok(error instanceof Error);
                 if (reportErrors) {
-                    this.logger.error(error.message);
+                    this.logger.error(`Failed to set option ${key}: ${error.message}`);
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


ocavue:
https://stackblitz.com/github/issueset/typedoc-markdown-flags-issue?file=config%2Ftypedoc-base.json

You can open this link to see the `[error] Cannot convert undefined or null to object` error. 

I'm using the `namedAnchors` option from `typedoc-plugin-markdown`. Maybe this is an issue on the `typedoc-plugin-markdown` side.

Gerrit0:
Ah, looks like I messed up in `Options.copyForPackage`, that function copies the declarations from the root options object, but neglects setting their default values. typedoc-plugin-markdown is doing the right thing.

Gerrit0:
I don't think I want to add the logging to all of the readers, the errors thrown from the Options class (especially those that can be caused by bad configuration) should already have the option name in them, will review those to make sure they look right.

ocavue:
thanks for your fix！

Gerrit0:
Thank you! Your repro made it quickly obvious what the problem was once I looked at it :)

PR #2434, opened 2023-10-31
fix: gitRevision fix in sourceLinkTemplate

swarnpallav:
`gitRevision` is not being replaced with gitRevision value provided in config.


Gerrit0:
How on earth did I miss this?! Thanks!

PR #2445, opened 2023-11-20
chore(deps): bump typescript to 5.3.2

li-jia-nan:
Resolves #2445

### DIFF
--- a/.config/.prettierignore
+++ b/.config/.prettierignore
@@ -12,3 +12,6 @@
 ../example/docs/
 **/tmp
 **/.vs
+
+# Remove once Prettier has support
+../src/test/converter2/behavior/resolutionMode.ts
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ yarn-error.log
 /src/test/renderer/*/doc
 /src/test/renderer/testProject/json.json
 **/node_modules/
+!src/test/converter2/behavior/node_modules/
 /coverage/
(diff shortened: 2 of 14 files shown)


li-jia-nan:
@Gerrit0 Hi, can you review this PR ? thanks

Gerrit0:
I definitely want a test that checks that typedoc interacts with import attributes correctly - particularly that it doesn't get confused by resolution-mode. Looks like that's the only change in this TS version that matters to typedoc

li-jia-nan:
> I definitely want a test that checks that typedoc interacts with import attributes correctly - particularly that it doesn't get confused by resolution-mode. Looks like that's the only change in this TS version that matters to typedoc

I'm sorry, I don't know how to add a test, can you give me some tips?

Gerrit0:
Take a look in the src/test folder - particularly the behavior.c2.test.ts file. Tests in that file use the src/test/converter2/behavior files as input, so you can set up a file which references another file that has both esm and cjs variants that export a different type, and make sure that the converted reflections also respect that. I should have time Thursday or this weekend to build tests if you don't have the time.

Gerrit0:
Thanks!

li-jia-nan:
> Thanks!

You're welcome, I'd like to know when is the next release ?

Gerrit0:
Today or tomorrow, have a couple issues I'd like to get in

PR #2500, opened 2024-02-10
Add module readmes to json output

### LINKED ISSUE #2499: Module DeclarationReflections are missing readme property in --json output with --entryPointStrategy=packages
## Search terms

readme, json, --json, packages, monorepo, readme.md

## Expected Behavior

When using entryPointStrategy=packages, I would expect the --json output to include readme properties for the modules that have README.md files present.

## Actual Behavior

These readmes are present in the rendered HTML for the modules, but omitted from the json output.

## Steps to reproduce the bug

Repro: https://github.com/TypeStrong/typedoc-repros/pull/38

Manually:
1. Setup a project that uses entryPointStrategy=packages
	1a. For example, with entryPoints=['packages/*']
2. Ensure packages/a/README.md and packages/b/README.md exist, following this example
3. Verify the readme's are rendered in the HTML output, but not the --json output

## Environment

-   Typedoc version: 0.25.2
-   TypeScript version: 5.2.2
-   Node.js version: 19.9.0
-   OS: MacOS

JMBeresford:
Resolves #2499

--

This change attempts to fix the above by mimicking the behavior of the ProjectReflection json serialization, which properly outputs the json readme property.

I skimmed through the test suites, but could not find a good place to integrate a test for this at a glance. Let me know which existing tests should be updated (if needed) and where new test logic should be implemented and I'll dig into it, assuming this change in behavior is acceptable.

### DIFF
--- a/src/lib/models/reflections/declaration.ts
+++ b/src/lib/models/reflections/declaration.ts
@@ -311,6 +311,7 @@ export class DeclarationReflection extends ContainerReflection {
                 this.implementedTypes,
             ),
             implementedBy: serializer.toObjectsOptional(this.implementedBy),
+            readme: Comment.serializeDisplayParts(serializer, this.readme),
         };
     }
 
--- a/src/lib/models/reflections/declaration.ts
+++ b/src/lib/models/reflections/declaration.ts
@@ -321,14 +321,15 @@ export class DeclarationReflection extends ContainerReflection {
     ): void {
         super.fromObject(de, obj);
 
+        if (obj.readme) {
+            this.readme = Comment.deserializeDisplayParts(de, obj.readme);
+        }
+
         // This happens when merging multiple projects together.
         // If updating this, also check ProjectReflection.fromObject.
         if (obj.variant === "project") {
             this.kind = ReflectionKind.Module;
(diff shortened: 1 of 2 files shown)


src/lib/models/reflections/declaration.ts
```diff
@@ -311,6 +311,7 @@ export class DeclarationReflection extends ContainerReflection {
                 this.implementedTypes,
             ),
             implementedBy: serializer.toObjectsOptional(this.implementedBy),
+            readme: Comment.serializeDisplayParts(serializer, this.readme),
```
Gerrit0, line 314: Can you update the deserializer to handle the readme property unconditionally too?

Gerrit0 (changes requested):
I'm fine merging this without a test update -- I don't have a good way to test packages mode as thoroughly as I'd like yet.

src/lib/models/reflections/declaration.ts
JMBeresford: Done, I believe this to be sufficient -- the changes are working with my monorepo as expected.

Thanks also for the prompt feedback, as usual! 😄

Gerrit0 (approved):
Looks great, thanks!

PR #2511, opened 2024-02-29
Fix getReflectionsFromSymbolId from returning undefined elements

nbilyk:
Returning [undefined] causes downstream TypeErrors.

Example:
```
TypeError: Cannot read properties of undefined (reading 'kindOf')
    at ./node_modules/typedoc/dist/lib/models/types.js:688:25
    at Array.find (<anonymous>)
    at get reflection [as reflection]
```

### DIFF
--- a/src/lib/models/reflections/project.ts
+++ b/src/lib/models/reflections/project.ts
@@ -282,9 +282,13 @@ export class ProjectReflection extends ContainerReflection {
     getReflectionsFromSymbolId(symbolId: ReflectionSymbolId) {
         const id = this.symbolToReflectionIdMap.get(symbolId);
         if (typeof id === "number") {
-            return [this.getReflectionById(id)!];
+            const refl = this.getReflectionById(id);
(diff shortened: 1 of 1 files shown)


Gerrit0:
These assertions are in place because it shouldn't be possible for ids in symbolToReflectionIdMap to be invalid. If they are, that's a bug elsewhere. This hides the problem, but it's still broken.

nbilyk:
Gerrit, I thought that might be the case, although this commit does avoid what I was hitting.

I unfortunately don't have an isolated test case to provide, it happens when a library re-exports modules from another library that has type declarations.

Gerrit0:
Could you point me to it anyways? I might have enough time this weekend to dive into it

nbilyk:
Closing, I updated to 25.12 and no longer have this issue. Thanks whomever fixed this for me :)

Gerrit0:
Strange... nothing that I can remember should have changed this.

Issue #2518, opened 2024-03-07
Segoe UI font issue on Mac shows random letter/numbers in many places


docmattman:
## Search terms

Some text looks like gibberish with random letters and number on a Mac

## Expected Behavior

Text should show proper values

## Actual Behavior

Some output text shows garbled text

## Steps to reproduce the bug

View generated docs on a Mac that doesn't have `Segoe UI` font installed.  Lots of text in the docs will end up looking like this screenshot:

![Screenshot 2024-03-06 at 3 59 24 PM](https://github.com/TypeStrong/typedoc/assets/801209/ae380c2b-6a50-4438-8356-fa3f4740a8ae)

If I specifically install the `Segoe UI` font on my Mac (not super easy to find), then the text is fixed.  Perhaps it would be good to use an alternate default font that would work better cross-platform?

## Environment

-   Typedoc version: latest
-   TypeScript version: any
-   Node.js version: any
-   OS: Mac

Gerrit0:
What the... why didn't your browser fall back to the sans-serif fallback?

docmattman:
I was wondering the same thing.  I do have multiple colleagues on Macs that see the same behavior.  I think it mostly shows up in places with italic style fonts.

This is from Chrome on a Mac.

Gerrit0:
As someone who primarily works on a system not connected to the internet, I really like that TypeDoc today doesn't depend on any external resources...

I think a better fix is probably to switch to what GitHub is using, does this fix the issue for you?

```css
font-family: -apple-system,BlinkMacSystemFont,"Segoe UI","Noto Sans",Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji"
```

docmattman:
That makes sense.  To verify that change, I can plug that in to my local setup and check/confirm on Mac to make sure, but it seems like it should work fine.  I can post back later when I have a chance to check it.

Gerrit0:
That'd be great! I'd like to include that fix in the next release (tomorrow/Sunday) with the next TypeScript version support

docmattman:
Ok, I plugged in that font-family value and it all looks good on a Mac.  I made the change in the PR as well in case you want to merge it from there.  Thanks!

Gerrit0:
Awesome, thanks!

PR #2520, opened 2024-03-07
Change default font to be compatible with Mac

docmattman:
Segoe UI causes some text rendering issues on Mac.  Use Open Sans as a close replacement.

Closes: #2518


Gerrit0:
Thanks!

Issue #2557, opened 2024-04-26
Ensures select element has an accessible name.


msftedad:
**Test Environment:** 
OS: Windows 11
OS version: 23H2 ((OS Build 26052.1000)
Edge Version: 123.0.2380.1 (Official build) stable app, dev channel (64-bit)
Tool: AI for web/ windows
Typedoc version: v0.25.13
 
**Steps to Reproduce:**   
1. Open [TypeDoc - v0.25.13](https://typedoc.org/api/) in edge browser.
2. Run accessibility Insights tool and observe the issue.
 
**Actual result:**
Select element does not have an accessible name at the theme dropdown.

**Expected Result:**
Ensures select element has an accessible name.

![221470_main](https://github.com/TypeStrong/typedoc/assets/93735775/d33f4b40-33ba-4f02-b826-6f74d653cc32)
![221470](https://github.com/TypeStrong/typedoc/assets/93735775/30bb1530-2f27-48f1-a9d4-ce7562350528)
![fastpass](https://github.com/TypeStrong/typedoc/assets/93735775/f11384e6-7437-4577-aab2-9ec40b904d88)
![accessible name](https://github.com/TypeStrong/typedoc/assets/93735775/5823e672-2db1-4592-a37a-58e32eb60321)

Gerrit0:
Fixed with TypeDoc 0.26, which is releasing 2024/06/21

PR #2558, opened 2024-04-27
✔️ name attribute added for select tag

Nil2000:
This PR is with reference to #2557 
Hope this resolves.


Gerrit0:
This doesn't appear to solve the issue, since there isn't any `<label>` with that name for it to reference... I *think* the right approach here is to use aria-labeledby to reference the header

Nil2000:
Hi, @Gerrit0 I introduced `<label>` and made it associated with the `name` in `<select>`. Can you verify if this is fine?

Gerrit0:
MDN says that `<label><h3>` [is a bad idea](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#headings), it looks like the theme "label-heading" ought not be a header element, and just be a label to me... which probably means the same applies to the member visibility heading... slightly more involved here than I expected!

Nil2000:
Then it's better to use `<label>` and define a class having the same property of `<h4>`. I believe it's just the size that's require here.

Gerrit0:
I manually merged this into the beta branch with [56f92c5](https://github.com/TypeStrong/typedoc/commit/56f92c5f2f77a7ad5da58fef13ae8602cd825762), thanks!

Issue #2559, opened 2024-04-27
Suggestion: Custom footer string


nicfv:
## Search Terms

custom, footer, string, hide, generator

## Problem

By using the `--hideGenerator` option, the message "created by TypeDoc" is removed, but the footer is still there. It might be nice to include an option to customize what is written in the footer instead.

## Suggested Solution

`--footer "<span class="...">should support html</span>"`

Gerrit0:
Sure, seems like an easy customization to allow

Aryakoste:
Hello @Gerrit0. I would like to work on this issue.

Gerrit0:
Go for it!

xuhdev:
There's already a plugin for this: https://typedoc-extra-footer.8hob.io/

mayank1513:
> There's already a plugin for this: https://typedoc-extra-footer.8hob.io/

Please don't avoid adding any features because there is a plugin. Plugins don't work with all versions of Typedoc

xuhdev:
I don't oppose adding this feature to typedoc, (although architecture-wise I think typedoc should leave these kinds of specialized theme customization options to plugins, even as builtin ones.)

However, as the author, I think the behavior defined in the plugin above is better designed. Specifically,

- If the custom footer and `--hideGenerator` are both specified, don't hide the custom footer. Instead, add the custom footer after the generator text.
- Add a default style if the customer footer is plain text. This permits users with not much HTML and CSS knowledge to use without struggling making the footer look reasonable.

What do you think?

mayank1513:
I am more interested in adding markdown pages to the documentation.

As of now, I just employed a quick hack - https://react18-tools.github.io/react18-global-store/

But, I believe that this is a very important part of documentation and should be supported by the library.

Gerrit0:
In general, I agree with @xuhdev, the more specialized functionality I can push to plugins, the easier to maintain TypeDoc itself stays, which makes it easier to maintain. I also mostly agree with the builtin plugin approach, though unfortunately it is messier than I'd like due to other constraints (options are all declared in one place for TypeDoc so that http://typedoc.org/schema.json can be easily generated, in 0.26 option descriptions are translated, and I'd really like to not spread the translatable strings all over the place to make it easier for translators)

That said, having worked places where getting plugins for approved software was a nightmare, I think it's important that it be easy to customize TypeDoc without requiring a plugin for everything. This is why the `--customCss` option exists, despite being easily implemented with a plugin.

[...]

PR #2562, opened 2024-04-28
Custom footer string when hideDeclaration is true 

Aryakoste:
Fixes https://github.com/TypeStrong/typedoc/issues/2559


src/lib/output/themes/default/partials/footer.tsx
```diff
@@ -1,18 +1,23 @@
...
         <footer>
             {context.hook("footer.begin")}
-            {hideGenerator || (
-                <p class="tsd-generator">
-                    {"Generated using "}
-                    <a href="https://typedoc.org/" target="_blank">
-                        TypeDoc
-                    </a>
-                </p>
+            {customFooter ? (
```
Gerrit0, line 10: Hmm... it's unfortunate that there are now two options to control the generated footer... Something I probably want to change during the next round of breaking changes.

src/lib/utils/options/sources/typedoc.ts
```diff
@@ -404,6 +404,11 @@ export function addTypeDocOptions(options: Pick<Options, "addDeclaration">) {
         help: "Do not print the TypeDoc link at the end of the page.",
         type: ParameterType.Boolean,
     });
+    options.addDeclaration({
+        name: "footer",
+        help: "Set custom footer when hideGenerator option is set to true",
```
Gerrit0, line 409: I think it would be better to name this `customFooterHtml` to be consistent with the `customCss` option, also to hint that this option is just raw HTML that will be injected into the page

Gerrit0 (changes requested):

Gerrit0:
Thanks!

PR #2574, opened 2024-05-15
fix: Resolve the name of parameters referring to default exports

tristanzander:
Hello! My team recently discovered some strange behavior with TypeDoc resolving the type name of a class as `default` instead of the name of the class. I discovered that this happens when the following constraints are met:

1. A file that's not included in the documentation exports a default function/class.
3. Another file exports that default export as a non-default member.
4. A file that is included in documentation references that type in the parameter of a function, a property of a class, etc.

Note: This can only happen if the function/class is declared using `export default class` or `export default function`, since TypeScript will resolve the type back to any normal `export class` or `export function` with the correct name.

I noticed that TypeScript itself resolves the parameter's symbol type as the name `default` instead of the name of the type. I'm unsure if this is correct, so I applied conditional logic to help handle this particular case.

Here's a visual example from our generated docs:
image
...where the correct type should be `DataTable`.

Please let me know if a better solution is more desirable 🙏

### DIFF
--- a/src/lib/models/types.ts
+++ b/src/lib/models/types.ts
@@ -907,11 +907,39 @@ export class ReferenceType extends Type {
         return new ReferenceType(name, target, project, name);
     }
 
+    /**
+     * In certain cases, TypeScript returns the name `default` for the name of a type that is defined with
+     * the format `export default class` or `export default function`. This method checks for that case and returns the
+     * declaration of the export instead of the name `default`.
+     */
+    private static getNameForDefaultExport(
+        symbol: ts.Symbol,
+    ): string | undefined {
+        if (
+            symbol.name !== "default" ||
+            symbol.valueDeclaration === undefined
+        ) {
+            return;
+        }
+
+        const name = ts.getNameOfDeclaration(symbol.valueDeclaration);
+
+        if (!name) {
+            return;
+        }
+
+        if (ts.isIdentifier(name)) {
+            return name.text;
+        }
+    }
+
     static createSymbolReference(
         symbol: ts.Symbol,
         context: Context,
         name?: string,
     ) {
+        name ??= this.getNameForDefaultExport(symbol);
+
         const ref = new ReferenceType(
             name ?? symbol.name,
             new ReflectionSymbolId(symbol),
--- a/src/lib/models/types.ts
+++ b/src/lib/models/types.ts
@@ -907,39 +907,11 @@ export class ReferenceType extends Type {
         return new ReferenceType(name, target, project, name);
     }
 
-    /**
-     * In certain cases, TypeScript returns the name `default` for the name of a type that is defined with
-     * the format `export default class` or `export default function`. This method checks for that case and returns the
-     * declaration of the export instead of the name `default`.
-     */
-    private static getNameForDefaultExport(
-        symbol: ts.Symbol,
-    ): string | undefined {
-        if (
-            symbol.name !== "default" ||
-            symbol.valueDeclaration === undefined
-        ) {
-            return;
-        }
-
-        const name = ts.getNameOfDeclaration(symbol.valueDeclaration);
-
-        if (!name) {
-            return;
-        }
-
-        if (ts.isIdentifier(name)) {
-            return name.text;
-        }
-    }
-
     static createSymbolReference(
         symbol: ts.Symbol,
         context: Context,
         name?: string,
     ) {
-        name ??= this.getNameForDefaultExport(symbol);
-
         const ref = new ReferenceType(
             name ?? symbol.name,
             new ReflectionSymbolId(symbol),
--- a/src/test/issues.c2.test.ts
+++ b/src/test/issues.c2.test.ts
@@ -15,6 +15,7 @@ import {
     ProjectReflection,
     QueryType,
     ReferenceReflection,
+    ReferenceType,
     ReflectionKind,
     ReflectionType,
     SignatureReflection,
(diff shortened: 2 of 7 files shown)


src/test/issues.c2.test.ts
```diff
@@ -91,6 +92,21 @@ describe("Issue Tests", () => {
...
+        const usesDefaultExport = query(project, "usesDefaultExport");
+        const sig = usesDefaultExport.signatures?.[0];
+        ok(sig, "Missing signature for usesDefaultExport");
+        const param = sig.parameters?.[0];
+        ok(param, "Missing parameter");
+        equal(param.name, "param", "Incorrect parameter name");
+        const paramType = param.type as ReferenceType | undefined;
+        ok(paramType, "Parameter type is not a reference type or undefined");
+        equal(paramType.type, "reference", "Parameter is not a reference type");
+        equal(paramType.name, "DefaultExport", "Incorrect reference name");
```
Gerrit0, line 106: I was expecting to see `Default` here, `DefaultExport` is definitely not the right thing to show to the user as that name is internal to the module where the symbol is declared, `Default` also matches what TypeScript shows when hovering over the type.

Also, please stick this test at the end of the file, it's intended to be in order of the issue number so it's easy to find tests

src/lib/models/types.ts
```diff
@@ -907,11 +907,39 @@ export class ReferenceType extends Type {
         return new ReferenceType(name, target, project, name);
     }
 
+    /**
+     * In certain cases, TypeScript returns the name `default` for the name of a type that is defined with
+     * the format `export default class` or `export default function`. This method checks for that case and returns the
+     * declaration of the export instead of the name `default`.
+     */
+    private static getNameForDefaultExport(
```
Gerrit0, line 915: I think this is the wrong spot for this check, I'd recommend instead including it in `referenceConverter.convertType`.

Doing this in the reference converter means that we'll use the name used by the exported function rather than whatever internal name is used in the module where the symbol is declared, which is what we already do when converting type nodes.

```diff
diff --git a/src/lib/converter/types.ts b/src/lib/converter/types.ts
index 1360c137..6471cb73 100644
--- a/src/lib/converter/types.ts
+++ b/src/lib/converter/types.ts
@@ -744,9 +744,14 @@ const referenceConverter: TypeConverter<
             return ref;
         }
 
+        const name = ts.isIdentifier(node.typeName)
+            ? node.typeName.text
+            : node.typeName.right.text;
+
         const ref = ReferenceType.createSymbolReference(
             context.resolveAliasedSymbol(symbol),
             context,
+            name,
         );
         if (type.flags & ts.TypeFlags.Substitution) {
             // NoInfer<T>
```

src/test/issues.c2.test.ts
```diff
@@ -91,6 +92,21 @@ describe("Issue Tests", () => {
         );
     });
 
+    it("#2574", () => {
+        const project = convert();
+        const usesDefaultExport = query(project, "usesDefaultExport");
+        const sig = usesDefaultExport.signatures?.[0];
+        ok(sig, "Missing signature for usesDefaultExport");
```
Gerrit0, line 99: There's a helper for this:

```suggestion
        const sig = querySig(project, "usesDefaultExport");
```

src/test/issues.c2.test.ts
```diff
@@ -91,6 +92,21 @@ describe("Issue Tests", () => {
...
+        const param = sig.parameters?.[0];
+        ok(param, "Missing parameter");
+        equal(param.name, "param", "Incorrect parameter name");
+        const paramType = param.type as ReferenceType | undefined;
+        ok(paramType, "Parameter type is not a reference type or undefined");
+        equal(paramType.type, "reference", "Parameter is not a reference type");
```
Gerrit0, line 105: `as` doesn't check anything, so this is misleading, you can just check that the type's `type` member is `reference` to know it is a reference type.

```suggestion
        const paramType = param.type;
        equal(paramType?.type, "reference", "Parameter type is not a reference type");
```

Gerrit0 (changes requested):

src/test/issues.c2.test.ts
tristanzander: > I was expecting to see Default here, DefaultExport is definitely not the right thing to show to the user as that name is internal to the module where the symbol is declared, Default also matches what TypeScript shows when hovering over the type.

See, the weird thing is that TypeScript's `checker.getTypeFromSymbolAtLocation()` function returns the underlying `DefaultExport` in this instance, so I figured that's just how it was expected to show up. I'll see what I can do to get it to behave correctly, but I was having a lot of trouble getting that to work.

tristanzander: > Default also matches what TypeScript shows when hovering over the type.

When hovering over the type, it does show `Default` as the name and export. When navigating to that type, the LSP seems to bring me to the `DefaultExport` implementation, not the `export { default as Default }` statement at the file that is actually imported. It seems that TypeScript's behavior is set to resolve all references deeply to the base type implementation rather than showing the reference type from the re-export.

Gerrit0: Hmm... I guess that result is reasonable then

tristanzander:
@Gerrit0 I tried to find a way to resolve those symbols as their re-exported name but I couldn't find I decent solution that didn't break other tests. This seems to be the way that type checker wants to resolve those types. Do you have any ideas for how I would determine if a given type was re-exported to apply this logic?

tristanzander:
I am noticing that the `declaration.type.id` and `type.id` are different in this example. That might be a way to determine if the types are different, but I'm not sure how generics come into play here

src/test/issues.c2.test.ts
tristanzander: Think I should log this as an issue on [microsoft/TypeScript](https://github.com/microsoft/TypeScript)?

Gerrit0:
I incorrectly assumed that making the change to derive the name in the type converter would just make that work - given that the type converter is getting the resolved type, I think it's reasonable to leave the behavior as is

tristanzander:
@Gerrit0 Any chance I could get a re-review on this soon?

Gerrit0:
I meant to merge it last weekend, unfortunately it fell off my list. I'll definitely get it in next weekend!

Gerrit0:
Thanks!

PR #2580, opened 2024-05-30
Add partial Korean translations

XiNiHa:
This PR adds partial Korean translations only with strings used in the output documentation.


Gerrit0:
Thank you!

Issue #2581, opened 2024-06-01
Feature request: display warning and error count like eslint.


bladerunner2020:
## Problem

TypeDoc doesn't show the number of warnings and errors (though it does calculate them!). If a project has a lot of warnings (as it is in our case), it would be very convenient to see the number of warnings to ensure that the count goes down over time and not up.

## Suggested Solution

I suggest displaying the warning and error count like ESLint does. Possible outputs:

- if there are only warnings: `[warning] Found 0, error(s), 70 warning(s). `
- if there are errors: `[error] Found 1, error(s), 70 warning(s).`

Considering that TypeDoc already counts errors and warnings, it should only take a couple of lines to add this feature.

I don't mind to make a PR.

Gerrit0:
Seems like a reasonable request, go for it!

Gerrit0:
Included in TypeDoc 0.26, which is releasing 2024/06/21

PR #2583, opened 2024-06-01
Added summary on warnings and errors (#2581)

bladerunner2020:
Added a summary message about warnings and errors if any (see #2581) in the following functions:
- generateDocs
- generateJson

Added tests.


src/lib/application.ts
```diff
@@ -525,6 +525,14 @@ export class Application extends ChildableComponent<
         const start = Date.now();
         out = Path.resolve(out);
         await this.renderer.render(project, out);
+
+        // Generate a message summarizing the number of errors and warnings if there are any
```
Gerrit0, line 529: I think this is the wrong place to put this. TypeDoc has a `--watch` option, which will result in `generateDocs` and `generateJson` being called many times. If a user has both options turned on, then with this implementation they may get two warnings/errors for each compilation.

I think the way I'd prefer to handle this is to add a public `logRunSummary` (better name?) method on `Application` which is called in `cli.ts` in the appropriate places

Gerrit0: This also has the benefit of making the tests for this feature much less hacky - I've very intentionally avoided mocking pieces of TypeDoc in tests (almost) everywhere so far, they tend to lead to testing implementation details+easily broken tests

Gerrit0 (changes requested):
Thanks for the PR! I have a couple concerns

src/lib/application.ts
bladerunner2020: > I think this is the wrong place to put this. TypeDoc has a `--watch` option, which will result in `generateDocs` and `generateJson` being called many times. If a user has both options turned on, then with this implementation they may get two warnings/errors for each compilation.
> 
> I think the way I'd prefer to handle this is to add a public `logRunSummary` (better name?) method on `Application` which is called in `cli.ts` in the appropriate places

That seems reasonable. I'll take a look at it.

bladerunner2020:
I moved `logRunSummary` to `cli.ts` and agree that it's a better place for it. I also removed the test file as the tests are no longer necessary due to the simplicity of this solution. However, I did not add `logRunSummary` to the Application; I left it as a function in `cli.ts`.

Gerrit0 (approved):

Gerrit0:
Thanks!

Issue #2567, opened 2024-05-04
TypeDoc 0.26 Beta


Gerrit0:
TypeDoc 0.26 is now in beta! :tada:

Please try it out and report any issues here or in new issues:

```bash
npm install --save-dev typedoc@beta
```

The full release will be made on 2024-06-21.

This release includes support for TypeScript 5.5 in addition to two large features

### External Markdown Pages

It has been a highly requested feature for TypeDoc to support including additional markdown pages beyond just the project's readme. In fact the original issue requesting them, #247, is the oldest issue still open. There have been a few external plugins over the years which added this, then broke with TypeDoc updates, but it's now coming to TypeDoc proper!

This feature falls into two parts:

1. Project level documents
2. Reflection level documents

Project level documents can be added with the `--projectDocuments` option. These documents will be added as children of the root reflection object, so are a good fit for guides on how to use the library and other top level pages that should be included in your docs.

Reflection level documents can be added as children to declarations with the `@document` tag. These are intended for referencing lower level documents which are relevant when viewing documentation for that item. Note: "top level" declarations is a somewhat tricky term. The initial implementation of this feature only supports

The markdown pages added may include YAML formatted frontmatter. TypeDoc will check the frontmatter for `title`, `group`, and `category` properties. The `title` property will set the name of the document used, while `group` and `category` are treated like the `@group` and `@category` tags on a normal reflection.

Note: This frontmatter **must** begin and end with `---` on lines by itself. TypeDoc's frontmatter extraction uses this to determine when the block ends.

Example:

```ts
/**
 * @document relative/path/to/document.md
 */
namespace Foo {}
```

```md
---
title: Document name
group: Guides
---

This guide will show...
```

### Localization Support

TypeDoc has always been an English project, without any real support for other languages. This version sets up the infrastructure to support changing rendered text and TypeDoc's log messages to other languages. See [internationalization.md](https://github.com/TypeStrong/typedoc/blob/beta/internal-docs/internationalization.md) for details. Translation PRs are welcome!

### Important Breaking Changes

-   Dropped support for Node 16
-   Moved from marked to markdown-it for parsing markdown. TypeDoc has been on an old version of Marked for a while now to avoid breaking changes they made which turned parsing from a synchronous approach to potentially async. Supporting async code in TypeDoc's renderer is currently infeasible. Markdown-it also supports plugins which extend markdown and can be used within TypeDoc to add custom rendering of syntax within comments.
[...]

PR #2592, opened 2024-06-15
fix translations

steve02081504:
add more keys for https://github.com/TypeStrong/typedoc/issues/2567#issuecomment-2169046237
and some fixes


Gerrit0:
GitHub didn't pick it up as I rebased, but this was merged with 3eb73cb55c0c931f6fd5fb1231d4daebef17151c, thanks!

I made a couple changes after merging:

1. `theme_member_protected` etc. are from the `ReflectionFlag` enum, which now has a dedicated section, which highlighted some additional missing values
2. `tag_alias` is not in TypeDoc's default list of valid tags, so should not have a translation in the default translation objects. If you use it in your project, you can add translations for it using the `locales` object in your typedoc config:

```json
{
  "locales": {
    "zh": {
      "tag_alias": "別名"
    }
  }
}
```

(`@alias` is not supported by TypeDoc, in that TypeDoc does not support renaming exports arbitrarily, as specified by JSDoc, so I don't want to add it to the config)

PR #2598, opened 2024-06-18
add `theme_loading` key

steve02081504:


steve02081504:
I think maybe it would be easier for us to store the translation data in json to write scripts to automatically follow up the translations like https://github.com/SillyTavern/SillyTavern-i18n/blob/main/generate.py
what do you think?

Gerrit0:
Automatic translations are problematic. I intentionally avoided them when first building TypeDoc's system because I have seen it generate utterly wrong text (even with my two years of a foreign language in high school), and that was in a simpler context.

I reached out to some people I know, who confirmed my suspicions about using this for TypeDoc:

> I've been a translator for a few projects in the past, including the TS docs, and I can confirm this is indeed far from ideal
texts written in English usually use common vocabulary, where lots of words can have different meanings depending on the context
unlike "richer" languages like French, German, or even Chinese where concepts more often than not have a dedicated word or group of words
which leaves less room for interpretation / guessing from the context
also lots of technical words often don't have a proper translation, sometimes a word-for-word translation from English, or an obscure word that has decided upon by some language institution
👉  machine translation is possible,
but only for languages where enough bilingual content exists
and really depends on the kind of translator used (Google Translate vs Deepl vs some other AI stuff)

It's not called out explicitly anywhere in the internationalization docs, but I probably should add some wording that asks people not to submit translations if all they are doing is running keys through google translate without a good understanding of the language they are submitting translations for. I don't have the expertise to judge the translations for accuracy, so am entirely reliant on contributors for that.

One example where as a non-Chinese speaker I wouldn't be comfortable including Google translate based definitions is the text:

> Specify the default visibility for builtin filters and additional filters according to modifier tags

Which Google translates as:

> 根据修饰符标签指定内置过滤器和附加过滤器的默认可见性

When translated back to English, Google gives

> Specifies the default visibility of built-in filters and additional filters according to the modifier tag

Which implies there is *one* modifier tag. This isn't correct, so either the translation to Chinese is wrong, the translation back to English is wrong, or there is some hint in the Chinese translation that implies plural tags which Google didn't pick up, but would be obvious to a native speaker.

PR #2625, opened 2024-07-03
fix sidebar highlighting when links share path but not hash

garrett-hopper:
Fixes an issue where many items in the sidebar are highlighted at the same time. (Shared pathname, different hashes)

Less than ideal because it uses a page refresh to rebuild the sidebar
Perhaps it could instead add `current` to the active link and also somehow find the old link to remove `current` from?
Feel free to modify as you see fit.

### DIFF
--- a/src/lib/output/themes/default/assets/typedoc/Navigation.ts
+++ b/src/lib/output/themes/default/assets/typedoc/Navigation.ts
@@ -98,13 +98,13 @@ function addNavText(
         if (location.pathname + location.hash === a.pathname + a.hash) {
             a.classList.add("current");
         }
-        a.addEventListener("click", function(e) {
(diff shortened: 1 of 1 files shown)


Gerrit0:
I am very confused by this... could you share a docs site where this happens? The left navigation which this updates only includes individual pages, not hashes within the page.

garrett-hopper:
image

Turns out this is caused by a double re-export

`index.ts`
```typescript
export * as test from './test'
export * from './test'
```

`test/index.ts`
```typescript
export const test1 = 1
export const test2 = 1
```

Gerrit0:
Ah, I see! The re-exports shouldn't be showing up in the left hand page's navigation at all.

garrett-hopper:
Thanks! 😁

Issue #2628, opened 2024-07-07
Line endings on windows break frontmatter in `projectDocuments` CLRF


bukowa:
I don't know how much time I have wasted on debugging line endings on Windows in many projects. It's funny, for example, in Docker where if you have entrypoint.sh with CLRF it will simply tell you that "entrypoint file not found". I found a similar error in the markdown plugin for typedoc, I always come across it wherever I go. `CLRF` - I will probably get this tattooed :D

## Search terms
Line endings on windows break frontmatter in `projectDocuments` CLRF

## Expected Behavior
It works.

## Actual Behavior
It's broken. `quick_start` has CLRF below:

![9f5e5e3d8933a885af58fce8cf46ea2c](https://github.com/TypeStrong/typedoc/assets/38087302/0d0d7500-e755-4504-ad3c-1e9ac3a13693)
![badbad](https://github.com/TypeStrong/typedoc/assets/38087302/0d75158f-85b8-4b63-a428-0e58edcdb166)

## Steps to reproduce the bug
I added files here: https://gist.github.com/bukowa/e9623517b7e73f2494369af98185a13f#file-quick-start-html-L54
If necessary I can probably build a full example.

## Environment

-   Typedoc version: `0.26.3`
-   TypeScript version: `5.5.3`
-   Node.js version: `v22.4.0`
-   OS: `Windows 10`

Gerrit0:
Yep. That's it!

Gerrit0:
Fixed with #2631

PR #2631, opened 2024-07-09
fix: frontmatter parsing in windows (CLRF)

bukowa:
fixes https://github.com/TypeStrong/typedoc/issues/2628

It's not pretty but it works. ^^

### DIFF
--- a/src/lib/converter/comments/parser.ts
+++ b/src/lib/converter/comments/parser.ts
@@ -217,7 +217,10 @@ export function parseCommentString(
     if (firstBlock.text.startsWith(`---${lineBreak}`)) {
         const end = firstBlock.text.indexOf(`${lineBreak}---${lineBreak}`);
         if (end !== -1) {
-            const yamlText = firstBlock.text.slice(`---${lineBreak}`.length, end);
+            const yamlText = firstBlock.text.slice(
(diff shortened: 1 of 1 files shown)


Gerrit0 (changes requested):
I think we can make this a bit nicer + more robust by using a regex to extract the frontmatter.

If we use `firstBlock.text.match(/^(---\r?\n)([\s\S]+)(\r?\n---\r?\n)/)`, we'll either have a returned object with how much we need to skip when reporting error positions (`match[1].length`), the yaml text to parse (`match[2].length`) and the new start position of the first block (`match[0].length`), or no object, in which case there wasn't frontmatter. Thoughts?

This will also handle the issue where a file has mixed line endings, as some people don't normalize those...

Gerrit0:
I spent a bit more time looking at this... and turns out that earlier assumption was bad because of https://github.com/eemeli/yaml/issues/127, I ended up pulling your branch to play around with how to fix that, and fixed it another way..

Gerrit0:
I wasn't thinking, so rebased when looking at this locally instead of merging... merged with:

21cf5ae976c93d0ec0c1c55d5a5a8b2d3d6034be
ef0f26ee20a868b2198f1ab259101a943e64989e

PR #2658, opened 2024-08-07
add ref noopener to link to typedoc.org

felabr:
The link to typedoc.org in the footer of generated pages could be considered an unsafe 3rd party link vulnerable to tabnabbing and might be flagged in security audits. Adding `ref="noopener noreferrer"` will prevent that.


Gerrit0:
Any security audit which flags this is an indication that the auditor hasn't kept up to date with the specification. Since 2021, `target="_blank"` has implied `rel="noopener"`. See https://github.com/whatwg/html/issues/4078 for details.

felabr:
We managed to talk the auditors out of it. I'm a bit surprised myself :)
There's no reason to do anything here then.
Thanks anyway

PR #2664, opened 2024-08-12
Add support for "none" category

### LINKED ISSUE #2656: How to organize navigation (the left sidebar) by category without the remainder being placed in a default "Other" category?
## Search terms

navigation, category, default category

## Question
I'm trying to use the "navigation" > "includeCategories" option in typedoc.json to organize the left nav bar. When I use the "@category" tag within my doc comments, the reflections are correctly placed under those categories in the navigation, but everything else is placed under a default category called "Other". How do I disable this default categorization and allow any uncategorized reflections to just remain where they are?

I'm not sure how clear my question is, so here is an example: Suppose I have a module named colors.ts which has 6 interfaces named `Red`, `Orange`, `Yellow`, `Green`, `Blue`, and `Purple` and 1 class named `Color`. By default, once the documentation is generated, all 7 would be listed straight under "colors". If I then applied "@category Warm" to the first 3 interfaces and "@category Cool" to the latter 3 interfaces in their doc comments, I _want_ the navigation layout to look something like this:

- colors
    - C Color
    - Warm
        - I Red
        - I Orange
        - I Yellow
    - Cool
        - I Green
        - I Blue
        - I Purple

Instead, it looks like this:
- colors
    - Other 
        - C Color
    - Warm
        - I Red
        - I Orange
        - I Yellow
    - Cool
        - I Green
        - I Blue
        - I Purple

nick-ls:
Resolves #2656

```md
---
title: Syntax Highlighting
category: none
---
```
With `navigation.includeCategories` being set to true, this will place the markdown file at the top level of navigation:

image

### DIFF
--- a/src/lib/output/themes/default/DefaultTheme.tsx
+++ b/src/lib/output/themes/default/DefaultTheme.tsx
@@ -383,21 +383,11 @@ export class DefaultTheme extends Theme {
             }
 
             if (parent.categories && shouldShowCategories(parent, opts)) {
-                const noneCategory = parent.categories.find((x) => x.title === "none");
-                const otherCategories = parent.categories.filter((x) => x.title !== "none");
-
-                const mappedOthers = filterMap(otherCategories, toNavigation);
-
-                if (noneCategory) {
-                    const noneMappedChildren = filterMap(noneCategory.children, toNavigation);
-                    return [...noneMappedChildren, ...mappedOthers];
-                }
-
-                return mappedOthers;
+                return filterMapWithNoneCollection(parent.categories);
             }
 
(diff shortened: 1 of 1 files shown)


Gerrit0:
Awesome! I thought it was going to be a pretty easy update. Could you make the same addition to the groups section of the navigation builder?

nick-ls:
```ts
/**
 * Calculates the square root of a number.
 *
 * @group None
 */
export function sqrt(x: number): number;
```

Setting the group to `none` with `navigation.includeGroups` now places that group at the top level of navigation.

image

Gerrit0:
Thanks!

PR #2693, opened 2024-08-29
Added support for `@abstract` tag

### LINKED ISSUE #2692: Add support for `@abstract` tag
## Search Terms

jsdoc abstract

## Problem

The engine warns when the `@abstract` tag is used although it is a valid  jsdoc tag.

## Suggested Solution

Add it to the official allow list.

waynemwashuma:
Fixes #2692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -   Added `markdownLinkExternal` option to treat `http[s]://` links in markdown documents and comments as external to be opened in a new tab, #2679.
 -   Added `navigation.excludeReferences` option to prevent re-exports from appearing in the left hand navigation, #2685.
+-   Added support for the `@abstract` tag, #2692.
 
 ### Bug Fixes
 
--- a/src/lib/converter/plugins/CommentPlugin.ts
+++ b/src/lib/converter/plugins/CommentPlugin.ts
@@ -192,6 +192,15 @@ export class CommentPlugin extends ConverterComponent {
             comment.removeModifier("@interface");
         }
 
+        if (comment.hasModifier("@abstract")) {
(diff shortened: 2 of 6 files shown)


Gerrit0:
I think if typedoc is to support this tag, it should actually support it, and make it mark the reflection abstract. There are examples in CommentPlugin which do this already. 

I also think this should be marked as a modifier tag, not a block tag

waynemwashuma:
Currently when building docs, the abstract class is marked as such.
Screenshot 2024-08-30 05-13-49
I thought that this means it is internally supported or is there something else to be done?
Currently have no explicit plugins for typedoc unless the CommentPlugin is an internal one and set on by default.

Ive also moved the tag to modifier type.

Gerrit0:
The CommentPlugin is an internal one, as you suspected.

TypeDoc has some default rendering rules for tags. If you did `@someRandomTag` typedoc would produce a header with `Some Random Tag`

I don't want to just rely on the default rendering here as TypeDoc has real handling for TypeScript's `abstract` keyword, and I think `@abstract` should be an alternative for projects using JS rather than being slightly different.

waynemwashuma:
Ok, i might be able to connect the typescript implementation to the jsdoc javascript one but im not completely conversant with the codebase.

If one goes ahead to implement this,how will it be done?

Also,if this is not within your plans, ill leave it here.

Gerrit0:
All that should be necessary is adding another case for `@abstract` like there is for `@private` here:

https://github.com/TypeStrong/typedoc/blob/43f1dc20540b9bc6af1125db6cf0ed6deffccc00/src/lib/converter/plugins/CommentPlugin.ts#L195-L196

Gerrit0:
Thanks!

Issue #2717, opened 2024-09-27
TypeDoc displays a superfluous type on number constants


Zamiell:
Consider the following code:

```ts
const SECOND_IN_MILLISECONDS = 1000;
```

TypeDoc will generate a page with the following:

![image](https://github.com/user-attachments/assets/4b5fb08a-4ee4-43cd-91a2-93aa1979ec92)

Here, the type of "1000" is superfluous/confusing/noisy/distracting. Thus, I propose that when TypeDoc determines that the type is exactly the same thing as the value, then it should omit printing out the type.

Corso02:
Hi @Zamiell and @Gerrit0.
I managed to fix this issue. 
There were 2 tests that were failing right after I forked and cloned this repo. No other tests are failing after I implemented my solution. 

I used the same code that is mentioned in the description of this issue, and here is how the docs are generated with my solution in place. 
![image](https://github.com/user-attachments/assets/1ebedc5a-d05b-4ff6-ade4-e14bc1aec4e8)

Can I submit Pull Request?

Gerrit0:
Please do!

PR #2721, opened 2024-09-29
TypeDoc displays a superfluous type on number constants implementation

Corso02:
This PR is implementing the enhancement mentioned in #2717. 
When the type is exactly the same thing as the value, the type is not printed to docs.

### DIFF
--- a/src/lib/output/themes/default/partials/member.declaration.tsx
+++ b/src/lib/output/themes/default/partials/member.declaration.tsx
@@ -28,31 +28,33 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
     }
 
     const visitor = { reflection: renderTypeDeclaration };
-    
-     /** Fix for #2717. If type is the same as value the type is omited */
-    function shouldRenderType(){
-        if(props.type && props.type.type === "literal"){
+
+    /** Fix for #2717. If type is the same as value the type is omited */
+    function shouldRenderType() {
+        if (props.type && props.type.type === "literal") {
             const typeObject = props.type.toObject();
             const value = typeObject.value;
-            if(!value) { // should be unreachable
+            if (!value) {
+                // should be unreachable
                 return true;
             }
-            if(typeof value === "object") {
+            if (typeof value === "object") {
                 return true;
             }
             const reflectionTypeString: string = value.toString();
             let defaultValue = props.defaultValue!;
-            if(defaultValue){
+            if (defaultValue) {
                 // If the default value is string and it's wrapped in ' in the code, the value is wrapped in " and vice-versa
-                if( (defaultValue[0] === '"' && defaultValue[defaultValue.length - 1] === '"') ||
+                if (
+                    (defaultValue[0] === '"' && defaultValue[defaultValue.length - 1] === '"') ||
                     (defaultValue[0] === "'" && defaultValue[defaultValue.length - 1] === "'")
                 ) {
                     defaultValue = defaultValue.slice(1, -1);
                 }
             }
-            
-            if( reflectionTypeString === defaultValue.toString() ) {
-                return false;    
+
+            if (reflectionTypeString === defaultValue.toString()) {
+                return false;
             }
             return true;
         }
--- a/src/lib/output/themes/default/partials/member.declaration.tsx
+++ b/src/lib/output/themes/default/partials/member.declaration.tsx
@@ -29,34 +29,16 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
 
     const visitor = { reflection: renderTypeDeclaration };
 
-    /** Fix for #2717. If type is the same as value the type is omited */
+    /** Fix for #2717. If type is the same as value the type is omitted */
     function shouldRenderType() {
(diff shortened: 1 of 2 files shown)


src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
 
     const visitor = { reflection: renderTypeDeclaration };
 
+    /** Fix for #2717. If type is the same as value the type is omited */
+    function shouldRenderType() {
+        if (props.type && props.type.type === "literal") {
+            const typeObject = props.type.toObject();
+            const value = typeObject.value;
+            if (!value) {
+                // should be unreachable
```
Gerrit0, line 38: This isn't true, `value = 0` is falsy, it could also be a literal `false`

src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
 
     const visitor = { reflection: renderTypeDeclaration };
 
+    /** Fix for #2717. If type is the same as value the type is omited */
```
Gerrit0, line 32: Typo, "omitted"

src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
...
+            const value = typeObject.value;
+            if (!value) {
+                // should be unreachable
+                return true;
+            }
+            if (typeof value === "object") {
+                return true;
+            }
+            const reflectionTypeString: string = value.toString();
+            let defaultValue = props.defaultValue!;
```
Gerrit0, line 45: This assertion is unsafe, `defaultValue` may very well be undefined here, which would result in a crash due to using `defaultValue.toString()` below. Given that if it is `undefined` you want that comparison to be false anyways, I'd recommend just getting rid of the `toString` and the non-null assertion here.

src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
...
+    /** Fix for #2717. If type is the same as value the type is omited */
+    function shouldRenderType() {
+        if (props.type && props.type.type === "literal") {
+            const typeObject = props.type.toObject();
+            const value = typeObject.value;
+            if (!value) {
+                // should be unreachable
+                return true;
+            }
+            if (typeof value === "object") {
```
Gerrit0, line 41: I don't think this is the logic we want. This is the case for bigint literals. TypeDoc splits them up as they can't be JSON serialized, but in this case we'd want to check that it is different.

src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
 
     const visitor = { reflection: renderTypeDeclaration };
 
+    /** Fix for #2717. If type is the same as value the type is omited */
+    function shouldRenderType() {
+        if (props.type && props.type.type === "literal") {
+            const typeObject = props.type.toObject();
```
Gerrit0, line 35: I'd prefer to not call `toObject` here. That is intended for serialization, and this code ought to work with the non-serialized object instance. You might find `props.type.toString()` useful, however.

src/lib/output/themes/default/partials/member.declaration.tsx
```diff
@@ -29,12 +29,44 @@ export function memberDeclaration(context: DefaultThemeRenderContext, props: Dec
...
+                // should be unreachable
+                return true;
+            }
+            if (typeof value === "object") {
+                return true;
+            }
+            const reflectionTypeString: string = value.toString();
+            let defaultValue = props.defaultValue!;
+            if (defaultValue) {
+                // If the default value is string and it's wrapped in ' in the code, the value is wrapped in " and vice-versa
```
Gerrit0, line 47: Ew... so this breaks if there are any escape sequences in the string... I'm okay ignoring that edge case for now.

Gerrit0 (changes requested):

Corso02:
Hi @Gerrit0, thank you for your code review. I'll look into fixing the issues mentioned in the coming days. Would it be possible to assign me the #2717 issue? Thank you so much for your attention and participation.

Corso02:
I managed to fix most of the issues, below I have some examples of what I looked into.

Code: ```const falseVal = false;```
Output:
image

Code: ```const zero = 0;```
Output:
image

Code: ```const escaped_str_end = "Testing with escaped at the end\n"```
Output:
image
(This works the same with escaped character at the start or in the middle)


I have one question.
Consider this code:
```const bigNumber = 9007199254740991n;```
I get the output below:
image

Is this the correct output? Or should it be different? Thanks in advance.

Gerrit0 (approved):

PR #2733, opened 2024-10-07
Added headings option to control optional headings, changed all headings to h1, #2729

### LINKED ISSUE #2729: Better page heading behavior
## Search Terms

readme index heading

## Problem

Issue 1:
All other pages, excluding modules.html (see #2728), have an h1 heading, while the readme index.html page has an h2 heading. This causes an inconsistent look.

Issue 2:
Many (probably most) typescript projects start their README.md files with a project name heading, which leads to redundant headings when combined with the readme index.html page title.

## Suggested Solution

Issue 1:
Change the index.html heading from h2 to h1.

Issue 2 Option 1:
Add an `omitReadmeHeading` boolean option that skips rendering the page title heading.

Issue 2 Option 2:
Add an `omitReadmeHeading` boolean option that checks if the README.md file starts with a heading that matches the project name and removes it if found. This would have the added benefit of simplifying the `On This Page` section.

Issue 2 Option 3:
Same as option 2 but have the `omitReadmeHeading` option default to `true`.

Issue 2 Option 4 (preferred):
Check if the README.md file starts with a heading that matches the project name and remove it if found by default. This would have the added benefits of simplifying the `On This Page` section and not adding more options.

mrfigg:
Resolves #2729.

### DIFF
--- a/src/lib/output/themes/default/partials/header.tsx
+++ b/src/lib/output/themes/default/partials/header.tsx
@@ -2,19 +2,39 @@ import { classNames, getDisplayName, hasTypeParameters, join } from "../../lib";
 import { JSX } from "../../../../utils";
 import type { DefaultThemeRenderContext } from "../DefaultThemeRenderContext";
 import type { PageEvent } from "../../../events";
-import { type Reflection, ReflectionKind } from "../../../../models";
+import type { Reflection } from "../../../../models";
 
 export const header = (context: DefaultThemeRenderContext, props: PageEvent<Reflection>) => {
-    const HeadingLevel = props.model.isProject() ? "h2" : "h1";
+    const opts = context.options.getValue("headings");
+
+    // Don't render on the index page or the class hierarchy page
+    // We should probably someday render on the class hierarchy page, but currently breadcrumbs
+    // are entirely dependent on the reflection hierarchy, so it doesn't make sense today.
+    const renderBreadcrumbs = props.url !== "index.html" && props.url !== "hierarchy.html";
+
+    // Titles are always rendered on DeclarationReflection pages and the modules page for the project.
+    // They are also rendered on the readme + document pages if configured to do so by the user.
+    let renderTitle: boolean;
+    let titleKindString = "";
+    if (props.model.isProject()) {
+        if (props.url === "index.html" && props.model.readme?.length) {
+            renderTitle = opts.readme;
+        } else {
+            renderTitle = true;
+        }
+    } else if (props.model.isDocument()) {
+        renderTitle = opts.document;
+    } else {
+        renderTitle = true;
+        titleKindString = " " + context.internationalization.kindSingularString(props.model.kind);
+    }
+
     return (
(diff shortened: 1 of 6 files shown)


src/lib/output/themes/default/partials/header.tsx
```diff
@@ -5,12 +5,13 @@ import type { PageEvent } from "../../../events";
...
-    const HeadingLevel = props.model.isProject() ? "h2" : "h1";
+    const opts = context.options.getValue("headings");
+
     return (
         <div class="tsd-page-title">
             {!!props.model.parent && <ul class="tsd-breadcrumb">{context.breadcrumb(props.model)}</ul>}
-            {!props.model.isDocument() && (
-                <HeadingLevel class={classNames({ deprecated: props.model.isDeprecated() })}>
+            {(!props.model.isProject() || opts.project) && (!props.model.isDocument() || opts.document) && (
+                <h1 class={classNames({ deprecated: props.model.isDeprecated() })}>
```
Gerrit0, line 14: We should make the change to h1 in hierarchy.tsx as well

src/lib/output/themes/default/partials/header.tsx
```diff
@@ -5,12 +5,13 @@ import type { PageEvent } from "../../../events";
...
 export const header = (context: DefaultThemeRenderContext, props: PageEvent<Reflection>) => {
-    const HeadingLevel = props.model.isProject() ? "h2" : "h1";
+    const opts = context.options.getValue("headings");
+
     return (
         <div class="tsd-page-title">
             {!!props.model.parent && <ul class="tsd-breadcrumb">{context.breadcrumb(props.model)}</ul>}
-            {!props.model.isDocument() && (
-                <HeadingLevel class={classNames({ deprecated: props.model.isDeprecated() })}>
+            {(!props.model.isProject() || opts.project) && (!props.model.isDocument() || opts.document) && (
```
Gerrit0, line 13: Hmmmm.... I've realized that this is actually kind of problematic due to the readme. Both the index and the root modules pages are rendered with the project as the model.... and it's probably somewhat confusing that changing the project option will affect both of the pages... 

Not quite sure how we ought to handle this yet... going to sleep on it.

mrfigg: Yea, I agree, not sure how that should be handled. We could add a `headings.readme` flag and then in `header.tsx` use something like the following:

```tsx
//...
{
    (
        !props.model.isProject() ||
        (props.url === "index.html" && opts.readme) ||
        (props.url !== "index.html" && opts.project)
    ) &&
    (!props.model.isDocument() || opts.document) &&
    (<h1...> ... </h1>)
}
// ...
```

Gerrit0: On second thought... the project page is really just another module, so this check should probably just be for index.html, I think we should always include the title on the modules page.

Gerrit0:
Thanks!

Issue #2726, opened 2024-10-05
Support <text> element in SVGs


lriggle-strib:
## Search Terms

svg, icons, text vs path

## Problem

The icons used to represent the various types have very small letters and are harder to identify for those with poor eyesight.

## Suggested Solution
Support the `<text>` element inside SVGs, allowing theme builders to replace the `partials/icon.tsx` file within their theme to use an actual text letter. Doing this allows the font/color/size to be set to something more accessible and easier to identify for those needing support.

Doing this would require updating `utils/jsx.elements.ts` with the following:
- add this interface:
    ```ts
    export interface JsxTextElementProps
      extends JsxSvgCoreProps,
        JsxSvgStyleProps,
        JsxSvgPresentationProps {
      x?: string | number;
      y?: string | number;
    }
    ```
- include `text: JsxTextElementProps;` in the `IntrinsicElements` interface.

Doing this will allow theme builders to update the icon paths with text elements in an included `partials/icon.tsx` file making the icons easier to read.

A side-by-side comparison of updated icons, manually updated in the generated javascript file.
| Before  | After |
| ------ | ----- |
| <img width="162" alt="Screenshot 2024-10-05 at 11 08 28 AM" src="https://github.com/user-attachments/assets/1625775f-6f10-4d21-b129-fa0af14202db"> | <img width="171" alt="Screenshot 2024-10-05 at 11 09 49 AM" src="https://github.com/user-attachments/assets/b97bb91e-702c-4591-9eeb-93a53254a0b2"> |

---

### OPTIONAL addition
Update the `partials/icon.tsx` file within the default theme to use `<text>` elements by default.

The icon for Module would go from:
```tsx
    [ReflectionKind.Module]() {
        return kindIcon(
            <path
[...]

PR #2736, opened 2024-10-08
2726 svg text icons

lriggle-strib:
Resolves: https://github.com/TypeStrong/typedoc/issues/2726

## Description
Update SVG icons to use `<text>` elements for the letters instead of pathing the letter shapes. This allows size/font/etc to be adjusted in accordance with the styling/theme being developed, so icons are easier to read/identify and are more accessible.

Out of the box, icons now look like this:
Screenshot 2024-10-08 at 10 42 11 AM

The definition for an icon goes from looking like:
```ts
    [ReflectionKind.Accessor]: () =>
        kindIcon(
            <path
                d="M8.85 16L11.13 7.24H12.582L14.85 16H13.758L13.182 13.672H10.53L9.954 16H8.85ZM10.746 12.76H12.954L12.282 10.06C12.154 9.548 12.054 9.12 11.982 8.776C11.91 8.432 11.866 8.208 11.85 8.104C11.834 8.208 11.79 8.432 11.718 8.776C11.646 9.12 11.546 9.544 11.418 10.048L10.746 12.76Z"
                fill="var(--color-text)"
            />,
            "#FF4D4D",
            true,
        ),
```

to
```ts
    [ReflectionKind.Accessor]: () =>
        kindIcon(
            <text
                fill="var(--color-icon-text)"
                font-family="var(--icon-font-family)"
                font-size="var(--icon-font-size)"
                font-weight="var(--icon-font-weight)"
                font-style="var(--icon-font-style)"
                x="6.5"
                y="17.5"
            >
                A
            </text>,
            "var(--color-ts-accessor)",
            true,
        ),
```

## Additional Changes
- Font family/size/weight/style are adjustable on each text icon
    - css properties for `--icon-font-family`, `--icon-font-size`, `--icon-font-weight`, `--icon-font-style` added to control those values
- Outline colors on multiple icons set to the `--color-ts-xyz` value rather than a hard coded color code
    - `--color-ts-xyz` properties that did not use the same color code as the hard coded icon updated to use that code
- Fill color on icons changed from `--color-text` to `--color-icon-text` to allow the icons to be managed independently of the text on the page.
- The following properties were being used in the stylesheet but not defined. Their definitions have been added
    - `--color-comment-tag-text`, `--color-comment-tag`, and `--color-ts-project`
- Use of `--color-ts-type-literal` removed, as there are comments earlier in the stylesheet explicitly stating it is not being defined for a reason
- Removed extra `white-space` property in `pre` tag style definition.
- Removed extraneous `0` from the end of the `margin-top` property in `.site-menu` style definition
- Css property for `--color-ts-reference` defined and used in icon definition

### DIFF
--- a/src/lib/output/themes/default/partials/icon.tsx
+++ b/src/lib/output/themes/default/partials/icon.tsx
@@ -238,7 +238,7 @@ export const icons: Record<
             >
                 R
             </text>,
-            "#FF4D82", // extract into a CSS variable potentially?
+            "var(--color-ts-reference)",
             true,
         ),
     [ReflectionKind.SetSignature]() {
--- a/src/lib/output/themes/default/partials/icon.tsx
+++ b/src/lib/output/themes/default/partials/icon.tsx
@@ -19,6 +19,15 @@ const kindIcon = (letterPath: JSX.Element, color: string, circular = false) => (
     </svg>
 );
 
+const textIcon = (letter: string, color: string, circular = false) =>
+    kindIcon(
+        <text fill="var(--color-icon-text)" x="50%" y="50%" dominant-baseline="central" text-anchor="middle">
+            {letter}
+        </text>,
+        color,
+        circular,
+    );
+
 export function buildRefIcons<T extends Record<string, () => JSX.Element>>(
     icons: T,
     context: DefaultThemeRenderContext,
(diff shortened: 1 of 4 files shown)


src/lib/utils/jsx.elements.ts
```diff
@@ -1169,3 +1170,16 @@ export interface JsxUseElementProps
     width?: string | number;
     height?: string | number;
 }
+
+/**
+ * Properties permitted on the `<text>` element.
+ *
+ * Reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use
```
Gerrit0, line 1177: This isn't the right link

src/lib/output/themes/default/partials/icon.tsx
```diff
@@ -52,50 +52,88 @@ export const icons: Record<
...
-                d="M8.85 16L11.13 7.24H12.582L14.85 16H13.758L13.182 13.672H10.53L9.954 16H8.85ZM10.746 12.76H12.954L12.282 10.06C12.154 9.548 12.054 9.12 11.982 8.776C11.91 8.432 11.866 8.208 11.85 8.104C11.834 8.208 11.79 8.432 11.718 8.776C11.646 9.12 11.546 9.544 11.418 10.048L10.746 12.76Z"
-                fill="var(--color-text)"
-            />,
-            "#FF4D4D",
+            <text
+                fill="var(--color-icon-text)"
+                font-family="var(--icon-font-family)"
+                font-size="var(--icon-font-size)"
+                font-weight="var(--icon-font-weight)"
+                font-style="var(--icon-font-style)"
```
Gerrit0, line 60: I'm worried about exposing all of this with the existing setup. Exposing it this way may be convenient for the one or two people who just want to tweak this, but it implies that changing it will for at least some inputs result in reasonable output... which it won't.

Different font family = likely different width characters, so it won't be centered properly anymore, especially with the hardcoded x/y coordinates... even the existing set of font families likely won't quite play.
Different font size = Almost certainly won't fit properly.
Different font weight = Okay, this one is probably fine... but why? To expose everything? Making everything infinitely tweakable via CSS is very scary to me as essentially any change to the CSS is potentially breaking someone.
Different font style = Ditto...

Changing the text/background/border colors, sure, the rest I'm going to need some convincing for...

src/lib/output/themes/default/partials/icon.tsx
```diff
@@ -106,36 +144,63 @@ export const icons: Record<
...
+                x="10"
+                y="18"
+            >
+                I
```
Gerrit0, line 156: This doesn't appear to be quite centered to me.

image

src/lib/output/themes/default/partials/icon.tsx
```diff
@@ -146,28 +211,52 @@ export const icons: Record<
...
+            <text
+                fill="var(--color-icon-text)"
+                font-family="var(--icon-font-family)"
+                font-size="var(--icon-font-size)"
+                font-weight="var(--icon-font-weight)"
+                font-style="var(--icon-font-style)"
+                x="7"
+                y="17.5"
+            >
+                T
```
Gerrit0, line 258: This doesn't appear quite centered

image

src/lib/output/themes/default/partials/icon.tsx
Gerrit0: It appears that SVG has a way to just say "center this text"

```html
<text x="50%" y="50%" dominant-baseline="central" text-anchor="middle">I</text>
```

src/lib/output/themes/default/partials/icon.tsx
lriggle-strib: Sure thing, my goal was to make sure people had the ability to adjust those settings to ensure they could be made as accessible as possible, but I can bring it back. We're doing a lot of things at work right now with accessibility practices, and I may have gone overboard.

src/lib/output/themes/default/partials/icon.tsx
lriggle-strib: > It appears that SVG has a way to just say "center this text"

There is, but the way things are constructed currently, that doesn't work. With no other changes, when I set the x/y properties to 50%, they render like this. I'm not really sure why.
Screenshot 2024-10-10 at 2 35 09 PM

Gerrit0: That's because of the dominant-baseline and text-anchor properties - without them, the icons don't use the center of the letter as the rendered center

Gerrit0:
Thanks! I decided to address the comments I left as I want to push a release before avoiding touching a computer for a few days.

lriggle-strib:
@Gerrit0 Sure thing, i can continue to work on the placement of the letters in an effort to make it less brittle. I know there's improvement to be made with it.

PR #2749, opened 2024-10-14
Implemented client side hierarchy & expandable hierarchy section, #2744

### LINKED ISSUE #2744: Expandable Hierarchy Section
## Search Terms
hierarchy

## Problem

Having to go to a separate page (`hierarchy.html`) to view the full hierarchy of a declaration is inconvenient, and frankly the fact that `hierarchy.html` does not cleanly fit into the site navigation just gets under my skin (in a pet peeve sense).

## Suggested Solution

Make the hierarchy section on a declaration page expandable.

Here's a working example of how to do this with only html & css: [Codepen](https://codepen.io/mrfigg/pen/NWQbpjY)

mrfigg:
Resolves #2744.

Feedback & changes welcome.

### DIFF
--- a/src/lib/output/plugins/HierarchyPlugin.ts
+++ b/src/lib/output/plugins/HierarchyPlugin.ts
@@ -1,14 +1,28 @@
 import * as Path from "path";
 import { Component, RendererComponent } from "../components";
 import { RendererEvent } from "../events";
-import { writeFile } from "../../utils";
+import { JSX, writeFile } from "../../utils";
 import { DefaultTheme } from "../themes/default/DefaultTheme";
 import { gzip } from "zlib";
 import { promisify } from "util";
-import { type Reflection } from "../../models";
+import {
+    DeclarationReflection,
+    ReferenceType,
+    ReflectionKind,
+    type Reflection,
+} from "../../models";
 import { UrlMapping } from "../models/UrlMapping";
 
 const gzipP = promisify(gzip);
+
+export interface HierarchyElement {
+    html: string;
+    text: string;
+    path: string;
+    kind: ReflectionKind;
+    parents?: ({ path: string } | { html: string; text: string })[];
+    children?: ({ path: string } | { html: string; text: string })[];
+}
 
 @Component({ name: "hierarchy" })
 export class HierarchyPlugin extends RendererComponent {
@@ -39,7 +53,51 @@ export class HierarchyPlugin extends RendererComponent {
 
         const context = theme.getRenderContext(pageEvent);
 
-        const hierarchy = context.getHierarchy();
+        const hierarchy = (
+            event.project.getReflectionsByKind(
+                ReflectionKind.ClassOrInterface,
+            ) as DeclarationReflection[]
+        )
+            .filter(
+                (reflection) =>
+                    reflection.extendedTypes?.length ||
+                    reflection.extendedBy?.length,
+            )
+            .map((reflection) => ({
+                html: JSX.renderElement(
+                    context.type(
+                        ReferenceType.createResolvedReference(
+                            reflection.name,
+                            reflection,
+                            reflection.project,
+                        ),
+                    ),
+                ),
+                // Full name should be safe here, since this list only includes classes/interfaces.
+                text: reflection.getFullName(),
+                path: reflection.url!,
+                kind: reflection.kind,
+                parents: reflection.extendedTypes?.map((type) =>
+                    !(type instanceof ReferenceType) ||
+                    !(type.reflection instanceof DeclarationReflection)
+                        ? {
+                              html: JSX.renderElement(context.type(type)),
+                              text: type.toString(),
+                          }
+                        : { path: type.reflection.url! },
+                ),
+                children: reflection.extendedBy?.map((type) =>
+                    !(type instanceof ReferenceType) ||
+                    !(type.reflection instanceof DeclarationReflection)
+                        ? {
+                              html: JSX.renderElement(context.type(type)),
+                              text: type.toString(),
+                          }
+                        : { path: type.reflection.url! },
+                ),
+            }));
+
+        if (!hierarchy.length) return;
 
         const hierarchyJs = Path.join(
             event.outputDirectory,
--- a/src/lib/output/plugins/HierarchyPlugin.ts
+++ b/src/lib/output/plugins/HierarchyPlugin.ts
@@ -17,7 +17,6 @@ export interface HierarchyElement {
     html: string;
     text: string;
     path: string;
-    kind: ReflectionKind;
     parents?: ({ path: string } | { html: string; text: string })[];
     children?: ({ path: string } | { html: string; text: string })[];
 }
@@ -66,7 +65,6 @@ export class HierarchyPlugin extends RendererComponent {
                 // Full name should be safe here, since this list only includes classes/interfaces.
                 text: reflection.getFullName(),
                 path: reflection.url!,
-                kind: reflection.kind,
                 parents: reflection.extendedTypes?.map((type) =>
                     !(type instanceof ReferenceType) ||
                     !(type.reflection instanceof DeclarationReflection)
@@ -88,6 +86,12 @@ export class HierarchyPlugin extends RendererComponent {
             }));
 
         if (!hierarchy.length) return;
+
+        hierarchy.forEach((element) => {
+            if (!element.parents?.length) delete element.parents;
+
+            if (!element.children?.length) delete element.children;
+        });
 
         const hierarchyJs = Path.join(
             event.outputDirectory,
--- a/src/lib/output/plugins/HierarchyPlugin.ts
+++ b/src/lib/output/plugins/HierarchyPlugin.ts
@@ -1,6 +1,6 @@
 import * as Path from "path";
 import { Component, RendererComponent } from "../components";
-import { RendererEvent } from "../events";
+import { PageEvent, RendererEvent } from "../events";
 import { JSX, writeFile } from "../../utils";
 import { DefaultTheme } from "../themes/default/DefaultTheme";
 import { gzip } from "zlib";
@@ -9,9 +9,7 @@ import {
     DeclarationReflection,
     ReferenceType,
     ReflectionKind,
-    type Reflection,
 } from "../../models";
-import { UrlMapping } from "../models/UrlMapping";
 
 const gzipP = promisify(gzip);
 
@@ -43,15 +41,7 @@ export class HierarchyPlugin extends RendererComponent {
     private async buildHierarchy(event: RendererEvent) {
         const theme = this.owner.theme as DefaultTheme;
 
-        const [_, pageEvent] = event.createPageEvent<Reflection>(
-            new UrlMapping<Reflection>(
-                "assets/hierarchy.js",
-                event.project,
-                () => "",
-            ),
-        );
-
-        const context = theme.getRenderContext(pageEvent);
+        const context = theme.getRenderContext(new PageEvent(event.project));
 
         const hierarchy = (
             event.project.getReflectionsByKind(
--- a/src/lib/output/plugins/HierarchyPlugin.ts
+++ b/src/lib/output/plugins/HierarchyPlugin.ts
@@ -1,0 +1,59 @@
+import * as Path from "path";
+import { Component, RendererComponent } from "../components";
+import { RendererEvent } from "../events";
+import { writeFile } from "../../utils";
+import { DefaultTheme } from "../themes/default/DefaultTheme";
+import { gzip } from "zlib";
+import { promisify } from "util";
+import { type Reflection } from "../../models";
+import { UrlMapping } from "..";
+
+const gzipP = promisify(gzip);
+
+@Component({ name: "hierarchy" })
+export class HierarchyPlugin extends RendererComponent {
+    override initialize() {
+        this.owner.on(RendererEvent.BEGIN, this.onRendererBegin.bind(this));
+    }
+
+    private onRendererBegin(_event: RendererEvent) {
+        if (!(this.owner.theme instanceof DefaultTheme)) {
+            return;
+        }
+
+        this.owner.preRenderAsyncJobs.push((event) =>
+            this.buildHierarchy(event),
+        );
+    }
+
+    private async buildHierarchy(event: RendererEvent) {
+        const theme = this.owner.theme as DefaultTheme;
+
+        const [_, pageEvent] = event.createPageEvent<Reflection>(
+            new UrlMapping<Reflection>(
+                "assets/hierarchy.js",
+                event.project,
+                () => "",
+            ),
+        );
+
+        const context = theme.getRenderContext(pageEvent);
+
+        const hierarchy = context.getHierarchy();
+
+        const hierarchyJs = Path.join(
+            event.outputDirectory,
+            "assets",
+            "hierarchy.js",
+        );
+
+        const gz = await gzipP(Buffer.from(JSON.stringify(hierarchy)));
+
+        await writeFile(
+            hierarchyJs,
+            `window.hierarchyData = "data:application/octet-stream;base64,${gz.toString(
+                "base64",
+            )}"`,
+        );
+    }
+}
--- a/src/lib/output/plugins/HierarchyPlugin.ts
+++ b/src/lib/output/plugins/HierarchyPlugin.ts
@@ -6,7 +6,7 @@ import { DefaultTheme } from "../themes/default/DefaultTheme";
 import { gzip } from "zlib";
 import { promisify } from "util";
 import { type Reflection } from "../../models";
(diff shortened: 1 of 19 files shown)


src/lib/output/themes/default/assets/typedoc/Hierarchy.ts
```diff
@@ -0,0 +1,339 @@
...
+                anchor.href = baseUrl + currentBranch.path!;
+                item.append(anchor);
+
+                anchor.innerHTML += `<svg width="20" height="20" viewBox="0 0 24 24" fill="none" class="tsd-kind-icon"><use href="#icon-${currentBranch.kind!}"></use></svg>`;
+                anchor.innerHTML += `<span>${currentBranch.text}</span>`;
+            } else if (item.classList.contains("tsd-hierarchy-target")) {
+                item.innerHTML += `<span>${currentBranch.text}</span>`;
+            } else {
+                item.innerHTML += currentBranch.html.replace(
+                    /(?<=<a [^>]*(?<=\s)href=")(?![a-zA-Z]+:\/\/)([^"]*)(?="[^>]*>)/g,
```
Gerrit0, line 319: That's... no... that's not going to be okay. It'd be better to set the innerHTML, then query for anchor elements and update the href attribute.

src/lib/output/plugins/HierarchyPlugin.ts
```diff
@@ -0,0 +1,59 @@
...
+                event.project,
+                () => "",
+            ),
+        );
+
+        const context = theme.getRenderContext(pageEvent);
+
+        const hierarchy = context.getHierarchy();
+
+        const hierarchyJs = Path.join(
```
Gerrit0, line 44: It'd be nice to skip writing this if it is empty, and also skip loading a script to handle it.

src/lib/output/themes/default/templates/reflection.tsx
```diff
@@ -39,7 +39,7 @@ export function reflectionTemplate(context: DefaultThemeRenderContext, props: Pa
             {hasTypeParameters(props.model) && <> {context.typeParameters(props.model.typeParameters)} </>}
             {props.model instanceof DeclarationReflection && (
                 <>
-                    {context.hierarchy(props.model.typeHierarchy)}
+                    {context.hierarchy(props.model)}
```
Gerrit0, line 42: If this change is made it has to wait for 0.27 due to being a breaking change

src/lib/output/themes/default/templates/hierarchy.tsx
```diff
@@ -1,46 +1,17 @@
...
-    return (
-        <>
-            <h2>{context.i18n.theme_class_hierarchy_title()}</h2>
-            {getHierarchyRoots(props.project).map((root) => (
-                <ul class="tsd-full-hierarchy">{fullHierarchy(context, root)}</ul>
-            ))}
-        </>
+            <ul class="tsd-hierarchy tsd-full-hierarchy">
+                <li class="tsd-hierarchy-item">
+                    <span>{context.i18n.theme_loading()}</span>
```
Gerrit0, line 12: This page should not need JS, including the full hierarchy in a place where users without it enabled can see it is useful.

Gerrit0 (commented):
Some initial comments, without having played with it much locally yet

src/lib/output/plugins/HierarchyPlugin.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        this.owner.preRenderAsyncJobs.push((event) =>
+            this.buildHierarchy(event),
+        );
+    }
+
+    private async buildHierarchy(event: RendererEvent) {
+        const theme = this.owner.theme as DefaultTheme;
+
+        const [_, pageEvent] = event.createPageEvent<Reflection>(
```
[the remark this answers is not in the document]
mrfigg: Wow, I did that at the start and forgot to go back to do it properly. Egg on my face. I'll get started on fixing this soon, but it might take me a bit to get to it.

src/lib/output/themes/default/DefaultTheme.tsx
```diff
@@ -214,7 +225,7 @@ export class DefaultTheme extends Theme {
             urls.push(new UrlMapping("index.html", project, this.indexTemplate));
         }
 
-        if (getHierarchyRoots(project).length) {
+        if (this.application.options.getValue("includeHierarchySummary") && getHierarchyRoots(project).length) {
```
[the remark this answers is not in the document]
mrfigg: I admit it might be rare, but I've had a small project that had a single class that involved a hierarchy. Having an entire page just for two lines that already appeared on both of the linked pages seemed like bloat to me.

src/lib/output/themes/default/assets/typedoc/Hierarchy.ts
```diff
@@ -0,0 +1,339 @@
...
+            baseUrl,
+            !targetPath,
+        )!;
+
+        list.classList.add("tsd-full-hierarchy");
+
+        container.append(list);
+    });
+
+    if (!targetPath && window.location.hash) {
```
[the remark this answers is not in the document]
mrfigg: Currently the "view full" link includes a hash to the interface/class on the `hierarchy.html` page. To emulate that behavior for content that was loaded dynamically this work around is required, but if I rework the PR to have `hierarchy.html` be static as seems to be the desire in a comment below it can easily be removed.

src/lib/output/themes/default/assets/typedoc/Hierarchy.ts
mrfigg: That's totally fair and an easy fix. I might be too used to abusing regex, I didn't even think about it.

src/lib/output/plugins/HierarchyPlugin.ts
mrfigg: Smart.

src/lib/output/themes/default/templates/reflection.tsx
mrfigg: Would adding a second argument be considered a breaking change? That could work.

src/lib/output/themes/default/templates/hierarchy.tsx
mrfigg: I was under the assumption that TypeDoc always requires JavaScript, as the entire navigation section is non-functional without it.

Assuming that was the case, this change was made to improve size on disk (otherwise the same data would be in both `hierarchy.html` and `assets/hierarchy.js`) and to make the hierarchy trees match. Currently interface hierarchies with multiple roots get partially rendered multiple times, where as the new rendering logic renders the entire tree once.

It would be possible to duplicate the rendering logic and make the `hierarchy.html` page static while fixing multi-root hierarchies, but that would still unnecessarily increase the file size if JavaScript is still required..

src/lib/output/plugins/HierarchyPlugin.ts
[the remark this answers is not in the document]
mrfigg: Wait, I remember why I did this initially, although I do still want to find a better way. `getHierarchy` relies on the `type` partial to render the `html` property for each element, which takes a `DefaultThemeRenderContext` argument. The only way I found to create a full `DefaultThemeRenderContext` instance was to fake a `PageEvent`. I'll look for an alternative, but do you know of a better way to handle this?

As for `getHierarchy` being on the context, very true, that should be easy enough to fix.

src/lib/output/themes/default/templates/reflection.tsx
Gerrit0: If the second argument is required, yes, that's also a breaking change. You shouldn't need it as `context.page` is the `PageEvent` for the current page

src/lib/output/themes/default/templates/hierarchy.tsx
Gerrit0: While the user experience is significantly degraded, it's still possible to navigate through the docs, the one link that exists in the navigation takes you to a page which lists all of its children, so it can be browsed via that method.

It will make the generated docs slightly larger, but including that information twice is acceptable, 500x times, less so :)

src/lib/output/plugins/HierarchyPlugin.ts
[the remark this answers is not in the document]
Gerrit0: Ah, ew.. that's very unfortunate. I'm going to say don't worry about it for now.  I'll look for a better way to do this when considering breaking changes for 0.27.

src/lib/output/themes/default/assets/typedoc/Hierarchy.ts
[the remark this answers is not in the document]
mrfigg: Removed

src/lib/output/themes/default/templates/reflection.tsx
mrfigg: Fixed

src/lib/output/themes/default/templates/hierarchy.tsx
mrfigg: Changed

src/lib/output/plugins/HierarchyPlugin.ts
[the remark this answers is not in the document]
mrfigg: Sounds good to me, I took another look around and didn't find a good alternative that wouldn't be a breaking change.

Gerrit0:
This gets really confused with multiple inheritance, as is possible with interfaces:

```ts
class Foundation {}

export interface Base extends Foundation {}
export interface Branch extends Base {}

export interface Base2 {}

export interface Combined extends Branch, Base2 {}
```

image

mrfigg:
Yea, I agree it's a bit confusing, but I'd argue it's more to do with an UL not being well suited to multi-parent trees in general. That's actually working as intended. Actually you came across one of my test cases, example 3. Is there an alternate layout that you would find more intuitive?

I came up with 3 rules for myself, but I'm open to changing them.
1. Children of multiple parents should appear below their last parent both vertically and horizontally. (examples 1 & 2)
2. Parents should appear above their top-most child both vertically and horizontally. (examples 3 & 4)
3. Parents & children should be sorted alphabetically.

I'm guessing we want that last one tweaked.

<details>
<summary>Example 1:</summary>

```ts
/*
Graph:
╔═══╗                         ╔═══╗
║ A ║                         ║ G ║
╚══╤╝                         ╚╤══╝
   └──┐                     ┌──┘
     ╔╧══╗     ╔═══╗     ╔══╧╗
     ║ B ║     ║ D ║     ║ F ║
     ╚══╤╝     ╚╤═╤╝     ╚╤══╝
        └──┐ ┌──┘ └──┐ ┌──┘
          ╔╧═╧╗     ╔╧═╧╗
          ║ C ║     ║ E ║
          ╚═══╝     ╚═══╝

Desired Tree:
╔═══╗
║ A ║
╚═╦╤╝
  ║└────┐
  ║    ╔╧══╗
  ╠════╣ B ║
  ║    ╚══╤╝
  ║       └──────┐
  ║    ╔═══╗     │
  ╚════╣ D ║     │
       ╚╤╦╤╝     │
        │║└────┐ │
        │║    ╔╧═╧╗
        │╚════╣ C ║
        │     ╚═══╝
╔═══╗   │
║ G ║   └────────┐
╚═╦╤╝            │
  ║└────┐        │
  ║    ╔╧══╗     │
  ╚════╣ F ║     │
       ╚═╦╤╝     │
         ║└────┐ │
         ║    ╔╧═╧╗
         ╚════╣ E ║
              ╚═══╝
*/

export interface E1A {}
export interface E1B extends E1A {}
export interface E1C extends E1B, E1D {}
export interface E1D {}
export interface E1E extends E1D, E1F {}
export interface E1F extends E1G {}
export interface E1G {}
```
</details>

<details>
<summary>Example 2:</summary>

```ts
/*
Graph:
╔═══╗     ╔═══╗          ╔═══╗
║ A ║     ║ C ║          ║ F ║
╚══╤╝     ╚╤══╝          ╚══╤╝
   └──┐ ┌──┘                └──┐
     ╔╧═╧╗                    ╔╧══╗
     ║ B ║                    ║ G ║
     ╚═╤╤╝                    ╚╤╤═╝
       ││         ┌────────────┘│
       │└─────────┼──┐ ┌────────┘
       └────────┐ │  │ │
               ╔╧═╧╗╔╧═╧╗
               ║ D ║║ E ║
               ╚═══╝╚═══╝

Desired Tree:
╔═══╗
║ A ║
╚══╤╝
   └──────┐
╔═══╗     │
║ C ║     │
╚═╦╤╝     │
  ║└────┐ │
  ║    ╔╧═╧╗
  ╚════╣ B ║
       ╚═╤╤╝
╔═══╗    ││
║ F ║    │└────────┐
╚═╦╤╝    └───────┐ │
  ║└────┐        │ │
  ║    ╔╧══╗     │ │
  ╚════╣ G ║     │ │
       ╚╤╦╤╝     │ │
        │║└────┐ │ │
        │║    ╔╧═╧╗│
        │╠════╣ D ║│
        │║    ╚═══╝│
        └╫─────┐ ┌─┘
         ║    ╔╧═╧╗
         ╚════╣ E ║
              ╚═══╝
*/

export interface E2A {}
export interface E2B extends E2A, E2C {}
export interface E2C {}
export interface E2D extends E2B, E2G {}
export interface E2E extends E2B, E2G {}
export interface E2F {}
export interface E2G extends E2F {}
```
</details>

<details>
<summary>Example 3:</summary>

```ts
/*
Graph:
                    ╔═══╗
                    ║ E ║
                    ╚╤══╝
                  ┌──┘
               ╔══╧╗
               ║ D ║
               ╚╤══╝
             ┌──┘
╔═══╗     ╔══╧╗
║ A ║     ║ C ║
╚══╤╝     ╚╤══╝
   └──┐ ┌──┘
     ╔╧═╧╗
     ║ B ║
     ╚═══╝

Desired Tree:
╔═══╗
║ 1 ║
╚═╦═╝
  ║    ╔═══╗
  ╚════╣ 2 ║
       ╚═╦═╝
         ║    ╔═══╗
         ╚════╣ A ║
              ╚══╤╝
╔═══╗            │
║ E ║            │
╚═╦╤╝            │
  ║└────┐        └──────┐
  ║    ╔╧══╗            │
  ╚════╣ D ║            │
       ╚═╦╤╝            │
         ║└────┐        │
         ║    ╔╧══╗     │
         ╚════╣ C ║     │
              ╚═╦╤╝     │
                ║└────┐ │
                ║    ╔╧═╧╗
                ╚════╣ B ║
                     ╚═══╝
*/

export interface E3A {}
export interface E3B extends E3A, E3C {}
export interface E3C extends E3D {}
export interface E3D extends E3E {}
export interface E3E {}
```
</details>

<details>
<summary>Example 4:</summary>

```ts
/*
Graph:
╔═══╗          ╔═══╗
║ A ║          ║ D ║
╚═╤╤╝          ╚╤══╝
  │└───────┐ ┌──┘
  │       ╔╧═╧╗
  └───┐   ║ C ║
      │   ╚╤══╝
      │ ┌──┘
     ╔╧═╧╗
     ║ B ║
     ╚═══╝

Desired Tree:
╔═══╗
║ A ║
╚═╤╤╝
  │└────────────┐
  └──────┐      │
╔═══╗    │      │
║ D ║    │      │
╚═╦╤╝    │      │
  ║└────┐│      │
  ║    ╔╧╧═╗    │
  ╚════╣ C ║    │
       ╚═╦╤╝    │
         ║└────┐│
         ║    ╔╧╧═╗
         ╚════╣ B ║
              ╚═══╝
*/

export interface E4A {}
export interface E4B extends E4A, E4C {}
export interface E4C extends E4A, E4D {}
export interface E4D {}
```
</details>

Gerrit0:
This is probably why the original implementation of this block didn't try to render everything... it predates me.

Hmm... my main concern with this implementation is that it runs into an issue where it isn't easy to tell from the hierarchy which classes belong to which parts of the tree. For example in Test Case 1 the tree really makes it look like `D` inherits from `A`

I'm not actually sure there's any better option than what you've done besides rendering an actual graph, which I don't want to maintain, so don't want to go there... going to think about this a bit more, but will probably just go with it as is.

Gerrit0:
After entirely to much deliberation, I've decided that the right way to handle this is:

1. Define the full hierarchy as starting at hierarchy roots, and including all children.
2. If a child has already appeared in the page and has children, bail out early. If JS is enabled, add a button to expand it in place
3. The "Expand" button on individual class/interface pages will simply include all hierarchy lists which contain the current reflection.

This isn't perfect, like the implementation here it means that showing the entire relevant hierarchy on a page accepts dropping any type arguments provided to parent classes but it avoids being completely unable to tell what members on the tree are children of others. (Preserving type arguments is a rather awful mess as really type parameters specified at the lower level should be passed through parent type parameters... but then if there's some conditional type or indexed type that ought to be handled specially... NO!), 

Figuring this out required a fair amount of experimentation, so I have this half-done on a local branch based on yours, hopefully I can get it merged in tomorrow

Issue #2597, opened 2024-06-17
Decouple `Renderer` class from HTML output


Gerrit0:
TypeDoc was originally built with the idea of HTML rendering, without consideration for other output types. Later on, TypeDoc gained support for rendering a project to JSON, and later still rendering to markdown (via [typedoc-plugin-markdown](https://github.com/typedoc2md/typedoc-plugin-markdown)). It would be neat if TypeDoc could be easily extended via plugins to render documentation in other formats (e.g. LaTeX -> PDF) without hacks.

Today, the markdown plugin [hijacks](https://github.com/typedoc2md/typedoc-plugin-markdown/blob/50f0b685154139bc2a110dfc14728ff4af3c1f2d/packages/typedoc-plugin-markdown/src/index.ts#L45-L63) the renderer to support markdown, which prevents users with the plugin activated from producing both HTML and markdown, but has the benefit of letting users just add the plugin, and suddenly get markdown where their HTML was previously being generated. Furthermore, despite markdown rendering not using it, this method still requires that users of it pay the 250+ms price of loading syntax highlighting.

The way it seems that this ought to work is that TypeDoc's renderer shouldn't care about HTML/JSON/Markdown/Tex/whatever, but instead deal with an interface which applies to all of them.

For a first cut at this, I propose:

```ts
export interface OutputOptions {
    type: string; // defined output type, html, json, etc.
    path: string;
}

export class Renderer extends EventDispatcher<RendererEvents> {
    /**
     * Define a new output that can be used to render a project to disc.
     */
    defineOutput(
        name: string,
        output: new (app: Application) => Output<MinimalDocument, {}>,
    ): void;

    /**
     * Render the given project reflection to all user configured outputs.
     */
    async writeOutputs(project: ProjectReflection): Promise<void>;

    /**
     * Render the given project with the provided output options.
     */
    async writeOutput(
        project: ProjectReflection,
        output: OutputOptions,
    ): Promise<void>;
}

export interface MinimalDocument {
    /** Path relative to the user specified output directory */
    filename: string;
}

/**
 * Base class of all output types.
 *
 * 0-N outputs may be enabled by the user. When enabled, the {@link Renderer} will construct
 * and instance of the requested class and use it to write a project to disc. The output class
 * will then be deleted; in watch mode, this means the class may be constructed many times.
 *
 * The renderer will first call {@link Output.getDocuments} which will be used to list the files
 * to be written. Each document returned will be passed to the {@link Output.render} function
[...]

PR #2769, opened 2024-11-10
Create dedicated html outputShortcut

tgreyuk:
@Gerrit0 Thanks for the great work on the beta release. I have been working against it to prepare typedoc-plugin-markdown when it lands. I hope this PR makes sense but let me know what you think.

## Summary / Purpose

While updating typedoc-plugin-markdown to properly utilize the "outputs" implementation  as per https://github.com/TypeStrong/typedoc/issues/2597 (which is fantastic by the way) I have stumbled upon a bit on an issue.

The issue is that the "html" outputShortcut is bound to the "out" declaration. By reading the above thread I think the initial intention was perhaps to rename the option? 

> "_The --out option will be replaced with a --html option._". 

I understand that this is a significant breaking change, so I see the importance of preserving the out option. Ideally, though, I would prefer to avoid introducing a plugin breaking change of requiring a "markdown" option to be specified (instead of the generic "out" option) in order to generate the markdown docs. As it stands there is no way around this.

## Proposal

I propose to preserve the existing "out" option as a basic path declaration and to introduce a dedicated "html" outputShortcut. To the end user this will have no effect, as the Outputs class will assume that no shortcut has been specified and apply the html output as the default output. 

However this also means that I can create a markdown output and set it as the default and preserve the existing behaviour of the "out" option.

ie something like this:

```
app.options.addDeclaration({
    name: 'markdown',
    outputShortcut: 'markdown',
    help: (i18n) => i18n.help_markdown(),
    type: ParameterType.Path,
    hint: ParameterHint.Directory,
    defaultValue: './docs',
  });

app.outputs.addOutput('markdown', async (out, project) => {
    await customMarkdownRenderer(out, project)
  });

  app.outputs.setDefaultOutput(() => {
    return {
      name: 'markdown',
      path: app.options.getValue('out'),
    };
  });
```

Then when outputs are specified in config, the dedicated shortcuts will be picked up so the existing behaviour is preserved here too:

```
 "outputs": [
        {
            "name": "html",
            "path": "../html-docs"
        },
        {
            "name": "markdown",
            "path": "../markdown-doc",
        }
    ],
```

## Changes

The changes are pretty simple:

- removed outputShortcut key from the "out" option.
- added a dedicated "html" outputShortcut
- updated help text (put the translation through chat gpt so not sure of the accuracy)

### DIFF
--- a/src/lib/internationalization/locales/en.cts
+++ b/src/lib/internationalization/locales/en.cts
@@ -221,7 +221,10 @@ export = {
         "If a symbol is exported multiple times, ignore all but the first export",
     help_externalSymbolLinkMappings:
         "Define custom links for symbols not included in the documentation",
-    help_out: "Specify the location the documentation should be written to",
+    help_out:
+        "Specify the location the documentation for the default output should be written to",
+    help_html:
+        "Specify the location where the html documentation should be written to. For the default theme, this is equivalent to the out option",
     help_json:
         "Specify the location and filename a JSON file describing the project is written to",
     help_pretty:
--- a/src/lib/internationalization/locales/jp.cts
+++ b/src/lib/internationalization/locales/jp.cts
@@ -234,7 +234,9 @@ export = localeUtils.buildIncompleteTranslation({
         "シンボルが複数回エクスポートされた場合、最初のエクスポート以外はすべて無視されます。",
     help_externalSymbolLinkMappings:
         "ドキュメントに含まれていないシンボルのカスタムリンクを定義する",
-    help_out: "ドキュメントを書き込む場所を指定します",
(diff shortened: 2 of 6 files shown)


Gerrit0:
Thanks, this is a good idea! However, the implementation doesn't quite work, as now `typedoc --out docs --json docs.json` will only produce JSON output.

Making `--out` dedicated to the default output makes the output class slightly simpler. The `setDefaultOutput` method can be replaced with one which just sets the default output name rather than a callback. Doing this will prevent accidental misuse by plugins which get a different option than `out`.

I'm also going to revert the changes to the non-english locales -- my policy there is not to include machine generated translations as I've seen bad output from them before... which I thought I'd put somewhere obvious, but apparently have not.

PR #2770, opened 2024-11-11
Introducing TypeDoc Guru on Gurubase.io

kursataktas:
Hello team,

I'm the maintainer of [Anteon](https://github.com/getanteon/anteon). We have created Gurubase.io with the mission of building a centralized, open-source tool-focused knowledge base. Essentially, each "guru" is equipped with custom knowledge to answer user questions based on collected data related to that tool.

I wanted to update you that I've manually added the [TypeDoc Guru](https://gurubase.io/g/typedoc) to Gurubase. TypeDoc Guru uses the data from this repo and data from the [docs](https://typedoc.org) to answer questions by leveraging the LLM.

In this PR, I showcased the "TypeDoc Guru", which highlights that TypeDoc now has an AI assistant available to help users with their questions. Please let me know your thoughts on this contribution.

Additionally, if you want me to disable TypeDoc Guru in Gurubase, just let me know that's totally fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Documentation generator for TypeScript projects.
 
 [![CI](https://github.com/TypeStrong/typedoc/workflows/CI/badge.svg)](https://github.com/TypeStrong/typedoc/actions)
-[![NPM Version](https://img.shields.io/npm/v/typedoc?color=33cd56&logo=npm)](https://www.npmjs.com/package/typedoc)
+[![NPM Version](https://img.shields.io/npm/v/typedoc?color=33cd56&logo=npm)](https://www.npmjs.com/package/typedoc) [![Gurubase](https://img.shields.io/badge/Gurubase-Ask%20TypeDoc%20Guru-006BFF)](https://gurubase.io/g/typedoc)
 
 ## Documentation
 


Gerrit0:
I'm going to decline this for a few reasons:

1. AI documentation is highly unlikely to produce ideal results, or even be accurate in many ways. It doesn't know how to say "I don't know" or "that isn't possible".

    how can I write a plugin to exclude private members -- hallucinated an `exclude` method, wrote *entirely* too much text (repeated problem)
    be brief, how to remove inherited members from typedoc -- hallucinated an option, should have recommended a plugin
    how can I combine multiple documentation sites -- useful-ish, covers packages, but doesn't mention the merge strategy, which is what I wanted.
    how can I rename a module -- Recommends `module Foo`... which is syntax that has been discouraged for *years* and isn't how modules are done today. Doesn't include `@module`
    how to generate both json and html output -- it recommends running twice rather than specifying `--out` and `--json` in one command
    how to document globals -- recommends an option which doesn't exist, doesn't mention that TypeDoc will document globals defined in a file if that file isn't a module and is provided as an entry point.

    The two answers that it got right in the first dozen or so questions I asked it was how to generate markdown and how to hide re-exports.

2. This is a really new service. It looks like it has been around for less than 2 months at this point. It hasn't proved its reliability, and like all AI services right now is highly suspect as some companies are pouring money into this without it being clear how costs will be recouped.

3. I don't want to advertise a service which I haven't personally found useful. 

Also: A Discord account joined the TypeScript discord back in September, advertised this service once, and has since been deleted. That's really suspicious from a "is something bad going on here" standpoint.

---

This does appear to be able to come up with reasonable answers to questions which have an answer that boils down to "use this option", and does surprisingly well when asked to do some simple things including creating some tiny plugins, so if you want to keep it up, that's fine. If this is still around in a few years, has become a reasonably well known tool that people link to, and hasn't introduced some crazy business model that makes me want to have nothing to do with it, I'll be open to revisiting linking to it.

kursataktas:
Thanks @Gerrit0 for the detailed review. I'll update this thread once I believe your concerns are addressed.

PR #2776, opened 2024-11-17
Update Chinese translations

XeroAlpha:
I will regularly submit PRs to update the Chinese translations, so there is no need to mention this PR in the changelog when the submitted translations correspond to the original text added in the current version.

Additionally, should [this line](https://github.com/XeroAlpha/typedoc/blob/2b9ee59e1664f8c818230d1c1b9a42056a6ed478/src/lib/converter/plugins/MergeModuleWithPlugin.ts#L74) be `tempRefl = tempRefl.parent!;`? Since `refl` and `refl.parent` keep unchanged inside the `while` scope.

### DIFF
--- a/src/lib/internationalization/locales/zh.cts
+++ b/src/lib/internationalization/locales/zh.cts
@@ -20,8 +20,8 @@ export = localeUtils.buildIncompleteTranslation({
         "对于监视模式，entryPointStrategy 必须设置为 resolve 或 expand",
     found_0_errors_and_1_warnings: "发现 {0} 个错误和 {1} 个警告",
 
-    // output_0_could_not_be_generated
-    // output_0_generated_at_1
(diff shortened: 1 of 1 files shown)


Gerrit0:
Thanks!

> I will regularly submit PRs to update the Chinese translations, so there is no need to mention this PR in the changelog when the submitted translations correspond to the original text added in the current version.

I do want to recognize your contributions, so will at least continue adding a thanks line :)

> Additionally, should [this line](https://github.com/XeroAlpha/typedoc/blob/2b9ee59e1664f8c818230d1c1b9a42056a6ed478/src/lib/converter/plugins/MergeModuleWithPlugin.ts#L74) be tempRefl = tempRefl.parent!;? Since refl and refl.parent keep unchanged inside the while scope.

Good catch! Yes, it absolutely should be

PR #2791, opened 2024-11-30
Closes #2790 allow png favicon

### LINKED ISSUE #2790: PNG format for favicon
## Problem

Thank you for adding the "favicon" option, that was the only reason I used the `typedoc-plugin-extras` plugin. However it is laking the support of PNG files which is fairly standard.

## Suggested Solution

Allow .png images as well as .ico and .svg.

mistic100:
Solves #2790 Adds "png" to the list of allowed extensions for the favicon option

I didn't see unit tests related to this option, did I missed them ?

~~TODO : change the chinese messages~~

### DIFF
--- a/src/lib/internationalization/locales/en.cts
+++ b/src/lib/internationalization/locales/en.cts
@@ -268,8 +268,7 @@ export = {
         "Path to the readme file that should be displayed on the index page. Pass `none` to disable the index page and start the documentation on the globals page",
     help_cname:
         "Set the CNAME file text, it's useful for custom domains on GitHub Pages",
-    help_favicon:
-        "Path to a .ico, .png or .svg file to include as the site icon",
+    help_favicon: "Path to favicon to include as the site icon",
     help_sourceLinkExternal:
         "Specifies that source links should be treated as external links to be opened in a new tab",
     help_markdownLinkExternal:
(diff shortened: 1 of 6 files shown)


src/lib/internationalization/locales/en.cts
```diff
@@ -388,7 +388,8 @@ export = {
         "hostedBaseUrl must start with http:// or https://",
     useHostedBaseUrlForAbsoluteLinks_requires_hostedBaseUrl:
         "The useHostedBaseUrlForAbsoluteLinks option requires that hostedBaseUrl be set",
-    favicon_must_be_ico_or_svg: "Favicon file must be either a .ico or .svg",
+    favicon_must_be_ico_or_svg:
```
Gerrit0, line 391: This is getting long enough that I think it should be made a template: `favicon_must_have_one_of_the_following_extensions_0`

Gerrit0 (changes requested):
You didn't miss any unit tests, HTML output testing is all done by hand right now as I haven't found a decent way to test it which doesn't take forever to run, doesn't cost a fortune, and isn't a nightmare to maintain.

src/lib/internationalization/locales/en.cts
mistic100: done

Gerrit0:
Thanks!

Issue #2794, opened 2024-12-02
Replace fetch(data) with local data usage


mikalai-snap:
## Search Terms

DecompressionStream, CSP, gzip, data URL, strict CSP, compressed JSON, fetch

## Problem

The current implementation uses a `data:` URL for the compressed search index, which relies on decompressing it with `DecompressionStream`. This approach conflicts with stricter Content Security Policy rules, particularly when `connect-src 'self'` is enforced, as `data:` URLs are not allowed in such scenarios. This makes it harder to adopt the tool in environments with strict CSP policies.

## Suggested Solution

Consider embedding the compressed search index as an inline Base64 string within the script and decompressing it directly in the browser using `DecompressionStream`. This avoids the need for `data:` URLs while maintaining compression benefits.

Here is an example of how that can be done:
```ts
// search.ts
window.searchData = "gzip+base64 string"; // as currently done, but without data:application/octet-stream;base64, prefix

// main.js
async function decompressAndParseData(base64Data) {
    const binaryData = Uint8Array.from(atob(base64Data), c => c.charCodeAt(0));
    const blob = new Blob([binaryData]);
    const decompressedStream = blob.stream().pipeThrough(new DecompressionStream("gzip"));
    const decompressedText = await new Response(decompressedStream).text();
    return JSON.parse(decompressedText);
}

const searchData = decompressAndParseData(window.searchData);
```

Gerrit0:
PR welcome!

mikalai-snap:
> PR welcome!

Submitted! https://github.com/TypeStrong/typedoc/pull/2797

PR #2797, opened 2024-12-03
Replaced fetch() with direct data access

mikalai-snap:
Addresses https://github.com/TypeStrong/typedoc/issues/2794

This PR updates the handling of the search index, navigation, and hierarchy data by removing the use of the Fetch API.  This retains compression benefits while ensuring full compatibility with strict CSP environments.


Gerrit0:
Very nice, thank you!

PR #2812, opened 2024-12-14
chore: remove internal comment from jsdoc

twlite:
Removes the following comment from jsdoc

> Changes need to happen here at some point. I think we should follow ESLint's new config system eventually: [https://eslint.org/blog/2022/08/new-config-system-part-1/](vscode-file://vscode-app/c:/Users/USER/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

image


Gerrit0:
Seems reasonable, thanks!

PR #2816, opened 2024-12-22
support regions in @includeCode

### LINKED ISSUE #2814: Support regions and line numbers within `@includeCode` tag reference
## Search terms

@knodes/typedoc-plugin-code-blocks, typedoc@v0.23, typedoc@v0.27, @codeblock

## Question

Back in typedoc 0.23, I could use the `@knodes/typedoc-plugin-code-blocks` plugin to include bits of code from other files into markdown files which would end up as doc pages thanks to `typedoc-plugin-pages`.

For example, consider a source file `examples/basic-usage.ts`:
```ts
// #region import
import sum from '../src/sum'
// #endregion import

console.log(
  // #region functionCall
  sum(3, 4)
  // #endregion functionCall
)
```
and a markdown file `pages/basic-usage.md`:
```md
# Sum

## Basic usage
### import the function
{@codeblock examples/basic-usage.ts#import}

### call the function
{@codeblock examples/basic-usage.ts#functionCall}
```
Then, in typedoc.json, I would have:
```json
{
  "pluginPages": {
    "pages": [{
      "name": "Docs",
      "children": [{
        "name": "Basic Usage",
        "source": "basic-usage.md"
      }]
    }]
  }
}
```

Well, it seems the `@knodes/typedoc-plugin-code-blocks` plugin isn't compatible with more recent versions of typedoc…

Is it still possible do achieve something like this, perhaps using a different plugin? I tried using `typedoc-plugin-include-example`, but that seems only to work in typescript files and not in markdown files.

The reason I want this is because I want to include thoughtfully composed guides in my docs (hence the markdown) which include code samples that are taken directly from example files which I actually run in my tests. This ensures that the code samples in the docs never fall out of sync with the actual code and are all tested.

shawninder:
This PR adds support for regions in `@includeCode` tags, meaning you can insert part of a file as a code block by referencing a region of that file.

## Example

**sourceFile.ts**
```ts
console.log("Code we don't want inserted")
// #region regionName
console.log('Code we want to insert')
// #endregion regionName
console.log("More code we don't want inserted")
```

**documents/some-page.md**
```md
Here is an example:
{@includeCode src/sourceFile.ts#regionName}
This is useful for reasons.
```

**output:**

----
Here is an example:
```ts
console.log('Code we want to insert')
```
This is useful for reasons.

----

## Notes
- Once merged, this will close #2814

@Gerrit0 if this basic strategy seems OK, I would add the following to this PR before merging:

## TO DO

- [x] Get a simple example working with Typescript
- [x] Use the appropriate region syntax for different languages as documented in the VS Code documentation for [Folding](https://code.visualstudio.com/docs/editor/codebasics#_folding)
- [x] Make sure small variants don't break things (like using `//#region NAME` instead of `// #region NAME`)
- [x] Support regions for Markdown files in the `@include` tag as well
- [x] Produce useful (translated) errors where appropriate:
  - [x] when a region is specified but cannot be found
  - [x] when a region is found but empty (perhaps just a warning for this one?)
  - [x] when a region is found more than once in a file
- [x] Add support for line-number syntax
- [ ] Add tests to validate success, error and warning paths related to these new features
- [x] Document the above additions in `site/tags/include.md`
- [x] Use these new features in the example app (`/example`)
- [x] Add an entry in the Changelog describing the changes brought about by this PR

Thoughts? Questions? Advice?

### Updates

- [x] Check all new regular expressions against the [RE DOS checker](https://devina.io/redos-checker)
- [x] Lint


Gerrit0:
Yes, this seems reasonable. Might want to split the logic out into its own function at this point.

> Add support for line-number syntax

Is there any case in which this would be a better option than using a region? It seems like an objectively worse UX to me...

shawninder:
> Might want to split the logic out into its own function at this point.

I'll probably be extracting a function to its own file for this, indeed.

> Is there any case in which [line-number syntax] would be a better option than using a region? It seems like an objectively worse UX to me...

Agreed. I'm just thinking of the case where you want to include code from a file you don't control or in which you can't add comments to define regions (think JSON, for example, which does't support comments). In this case the line-numbers seem like the only text-based approach (as opposed to using a language-aware AST) that could work…

Can I let you decide if it's worth including?

Gerrit0:
Ah, that's a very fair point, line numbers is probably worth it at some point

shawninder:
The holidays are over so I'm back at it Today, but I think I already made a mistake. I tried pulling the latest changes before pushing mine but this PR is still showing more than my own changes…

I'm not sure how to make the PR clean from here… Help?

---

Other questions and comments about my actual code changes:
- I'm using regular expressions to support whitespace in the region tags, is that OK? Do I need to guard them somehow to avoid regular expression related attack vectors?
- I tried my best at naming things in line with the naming convention I think I'm seeing in the package, but I don't feel like the names I found are particularly good, especially the names of the translated error messages. Ideas?
- I'm having trouble finding existing tests for the `@include` and `@includeCode` tags. These would be helpful before I write my own tests for the new features. Any guidance would be appreciated.
- I added new English error messages. Do I need to translate them? French is the only language I could translate to unassisted…
- I created new functions in `src/lib/converter/plugins/IncludePlugin.ts` and a new file called `src/lib/converter/utils/regionTagREsByExt.ts`. Am I following the existing strategy closely enough? I could move the new functions to that new file (at the cost of extra parameters to these functions) if that's better…
- I'm using a controversial double-ternary expression in `src/lib/converter/plugins/IncludePlugin.ts` Would you prefer I use different syntax?

---

Thanks for any and all feedback. Looking forward to wrapping this up and marking it ready for official review and eventual merging :)

Gerrit0:
> I'm not sure how to make the PR clean from here… Help?

Not exactly sure how this happened... it looks like some kind of rebase gone wrong as Git thinks you went and modified every commit. If you run `git fetch upstream` and `git rebase -i upstream/master`, `:wq`, then force push, your changes should be re-applied on top of current master.

> I'm using regular expressions to support whitespace in the region tags, is that OK? Do I need to guard them somehow to avoid regular expression related attack vectors?

This is fine, I usually toss new ones into https://devina.io/redos-checker to check

> I tried my best at naming things in line with the naming convention I think I'm seeing in the package, but I don't feel like the names I found are particularly good, especially the names of the translated error messages. Ideas?

This looks pretty reasonable to me, I'm considering going the same route as TypeScript for these names someday which will make them automatically generated, but for now it makes sense. The names are completely subject to change anyways.

> I'm having trouble finding existing tests for the `@include` and `@includeCode` tags. These would be helpful before I write my own tests for the new features. Any guidance would be appreciated.

It looks like I didn't write unit tests specifically for this as they were used in TypeDoc's site so effectively tested there. There's a test that was added later (search for 2800 in issues.c2.test.ts)

> I added new English error messages. Do I need to translate them? French is the only language I could translate to unassisted…

Don't worry about adding translations, they'll just use English until someone who speaks other languages natively adds a translation.

> I created new functions in src/lib/converter/plugins/IncludePlugin.ts and a new file called src/lib/converter/utils/regionTagREsByExt.ts. Am I following the existing strategy closely enough? I could move the new functions to that new file (at the cost of extra parameters to these functions) if that's better…

I probably would have left the region tag regexes in the IncludePlugin file, fine where they are though. I tend to only split things out to other files if they are useful in more than one place.

> I'm using a controversial double-ternary expression in src/lib/converter/plugins/IncludePlugin.ts Would you prefer I use different syntax?

Looks fine to me!

I'll take a closer look at the code tomorrow, off to spend more time with family :)

src/lib/converter/plugins/IncludePlugin.ts
```diff
@@ -108,8 +150,264 @@ export class IncludePlugin extends ConverterComponent {
...
+        file: string,
+        ext: string,
+        textPart: string,
+        text: string,
+        target: string,
+    ) {
+        const regionTagsList = regionTagREsByExt[ext];
+        let found: string | false = false;
+        for (const [startTag, endTag] of regionTagsList) {
+            const start = text.match(startTag(target));
```
Gerrit0, line 165: We should regex-escape `target` here, there's an `escapeRegExp` function in `src/lib/utils/general.ts`

Gerrit0 (commented):
Apologies for the delay here! Time has a way of slipping away... not going to merge this weekend either, but it's first on my list for next weekend. This is looking really good to me. I especially like the changes you made to the docs! It seems to work well in the playing around I've done so far.

Scutua (approved):

src/lib/converter/plugins/IncludePlugin.ts
shawninder: Good point. Done in my latest commit.

shawninder:
> Apologies for the delay here! Time has a way of slipping away... not going to merge this weekend either, but it's first on my list for next weekend.

No worries @Gerrit0, I am in no rush :)

About the conflicts in CHANGELOG.md, do you want me to remove my changes to the changelog or merge and fix or anything else?

Gerrit0:
I can fix it when merging :) I'll probably look at making a few unit tests as well

Gerrit0:
I finally got around to pulling this down, writing some unit tests, and just playing with it. Found a couple issues, but it's all merged to my `dev` staging branch, will go to master as soon as I'm ready to push a release (I want #2831 merged too), hopefully this time it will *actually* be this weekend.

phoneticallySAARTHaK:
@Gerrit0 hey, please don't make a release yet, I'd like to include the sidebar changes as well, since it might change some selectors again, and I don't wanna get cursed by theme maintainers for repeated breaking changes

Gerrit0:
... shoot, I forgot about themes, there were only two custom ones for 0.27 published last I looked, neither of which would have been affected by your changes. Now it looks like there are more which will be affected, so that rewrite will have to come with a minor version bump.

phoneticallySAARTHaK:
Alright, and btw, it won't be a rewrite. I'll just put the sidebar in a dialog.

rauschma:
Not sure if it’s worth the added complexity but you could allow multiple arguments for `@includeCode`. Their contents would be merged into a single code block:

``````ts
/**
 * This function returns an iterable where each item was produced by applying
 * `mapperFn` to an item of `iterable`.
 * 
 * @example
 * {@includeCode
 *   ../../test/async-usage_test.ts#imports,
 *   ../../test/async-usage_test.ts#example-map,
 * }
 */
``````

That would give you more flexibility w.r.t omitting lines inside a range of lines.

**Comment by Gerrit0** [comment]
That's definitely not worth it. If anything, I could see  `{@includeCode ../../test/async-usage_test.ts#imports,example-map}`, but allowing a single tag to import from multiple different files is not something I want to support.

**Comment by rauschma** [comment]
One more consideration (again, possibly not worth the complexity): optional removing of indents – taken from the `//#region` line (“dedenting”).

Example – someone using `@includeCode` for both regions in the following code:

```ts
//#region imports
import * as assert from 'assert/strict';
import { AsyncIterable } from '@rauschma/iterable/async';
const {toArray, fromIterable: fi} = AsyncIterable;
//#endregion imports

suite('async-usage_test.ts');

test('Using map()', async () => {
  //#region example-map
  assert.deepEqual(
    await toArray(AsyncIterable.map(x => x + x, fi(['a', 'b', 'c']))),
    ['aa', 'bb', 'cc']
  );
  //#endregion example-map
  assert.deepEqual(
    await toArray(AsyncIterable.map(x => x + x, fi([]))),
    []
  );
});
```

Not sure what a good syntax would be for switching dedenting on/off.

rauschma:
> I could see `{@includeCode ../../test/async-usage_test.ts#imports,example-map}`, but allowing a single tag to import from multiple different files is not something I want to support.

Makes sense! I’d use your suggested syntax. One alternative is: merging multiple `@includeCode` into a single code block if there is no content between them.

Gerrit0:
I'm not sure why GitHub thinks this isn't merged... but it is on master now, thanks for the feature!

PR #2831, opened 2025-01-20
Implement Search as ARIA spec compliant* dialog and combobox pattern

phoneticallySAARTHaK:
Pretty much a complete rewrite of the UI and UI logic for search.

UI Changes:
- fix height calculations in sectioning containers to remove extra scrollbars/overflows
- Fixes tab order of toolbar links
- Replaces search input with a button in toolbar, which in turn open a dialog, that has the search, which can be opened by clicking the button or pressing "/", or Ctrl+K
- Implement search as [combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/)  (and all the changes that comes along with it)
- Enabled virtualKeyboard API to adjust dialog height when keyboard opens in mobile (currently only chromium browsers support it)
- Removes the [`ready`](https://github.com/TypeStrong/typedoc/pull/2831/files#diff-cd1eb3d812c33ec99146bc893a6bec0e30a519c9b9a99e13709f03b2c7ed59c5L45) class from search element.
- Updat color variable to satisfy WCAG level AA

More details are in commit messages. I haven't updated the changelog, because I don't know what's relevant.

[Visual Regression script](https://github.com/TypeStrong/typedoc/blob/master/.github/CONTRIBUTING.md#visual-regression-testing) probably won't detect the change?

\**Mostly. There might be minor variations*


src/lib/output/themes/default/assets/bootstrap.ts
```diff
@@ -26,3 +26,8 @@ Object.defineProperty(window, "app", { value: app });
 initSearch();
 initNav();
 initHierarchy();
+
+if ("virtualKeyboard" in navigator) {
+    // @ts-ignore
```
phoneticallySAARTHaK, line 31: help! ts error

phoneticallySAARTHaK, line 31: help! ts error

phoneticallySAARTHaK: Source: https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API#detect_the_virtual_keyboard_geometry_using_javascript

src/lib/output/themes/default/assets/typedoc/components/Search.ts
```diff
@@ -33,130 +35,194 @@ interface SearchState {
...
+/**
+ * Populates search data into `state`, if available.
+ * Removes deault loading message
+ */
+async function updateIndex(state: SearchState, results: HTMLElement) {
     if (!window.searchData) return;
 
-    const data: IData = await decompressJson(window.searchData);
+    try {
+        const data: IData = await decompressJson(window.searchData);
```
phoneticallySAARTHaK, line 62: Not sure if a try-catch is practically needed here

src/lib/output/themes/default/assets/typedoc/Application.ts
```diff
@@ -20,6 +24,10 @@ window.translations ||= {
         "This member is normally hidden due to your filter settings.",
     hierarchy_expand: "Expand",
     hierarchy_collapse: "Collapse",
+    theme_search_index_not_available: "The search index is not available",
+    theme_search_no_results: "No results found",
+    theme_search_placeholder: "Search the docs",
```
phoneticallySAARTHaK, line 29: This placeholder string is not needed in js, should I make a commit to delete it?

src/lib/output/themes/default/assets/bootstrap.ts
Gerrit0: This is probably because the frontend JS doesn't have a proper tsconfig setup, this is fine for now

static/style.css
```diff
@@ -555,6 +548,58 @@
         border-left: 4px solid gray;
     }
 
+    button {
```
Gerrit0, line 551: This needs to be qualified -- it has made buttons within code blocks to copy the text nearly invisible

src/lib/output/themes/default/assets/typedoc/components/Search.ts
Gerrit0: It shouldn't be necessary, searchData is generated by TypeDoc and known to always be valid here

src/lib/output/themes/default/assets/typedoc/Application.ts
Gerrit0: Yes please, don't include useless strings

src/lib/output/themes/default/assets/typedoc/components/Search.ts
```diff
@@ -33,130 +35,194 @@ interface SearchState {
...
-        } else if (e.key === "ArrowDown") {
-            setCurrentResult(results, field, 1);
-            e.preventDefault();
+        if (resultCount === 0 || e.ctrlKey || e.metaKey || e.altKey) {
+            return;
+        }
+
+        // Get the visually focused element, if any
+        const currentId = field.getAttribute("aria-activedescendant");
+        const current = document.getElementById(currentId || "");
```
Gerrit0, line 143: This seems to cause console warnings when currentId is an empty string... seems like we ought to bail in that case?

Gerrit0 (changes requested):
A couple general issues:
1. With no search history, typing something and then clearing the search field results in the claim  "No Recent Searches"... which is rather confusing, I just searched for something! Didn't click on anything, but... maybe that's expected, but it greatly confused me for a bit. 
2. This introduced some contrast issues in the new search dialog (run chromium's lighthouse checks)

This is looking really awesome! It seems to behave significantly better on mobile and I like the change on desktop as well

phoneticallySAARTHaK:
Regarding the recent searches message, yup, totally agree, I think even left a comment about in the code that it's actually search results. I picked up it up from [remix.run](https://remix.run/docs/en/main), and I'm confused what to show there, "no recent... *What?*"

Gerrit0:
Would it be better to just not show anything there? It's a different dialog size, but just showing a search bar the first time it is opened doesn't seem completely wrong to me.

I also just realized that storing IDs is not necessarily the way we should go about that. IDs are somewhat likely to change between rebuilds of the site as they are dependent on conversion order, so if the very first class TypeDoc converts gets a new property, all the IDs will end up shifted.

phoneticallySAARTHaK:
Yup, I mainly added for cosmetic UX purposes. It's not important and I already had doubts about it (that's why it's the last commit). 

Without anything it feels.. empty. I'll get some quick design advice on this, otherwise I'll just remove it.

And yes I tested the change of IDs, I still added it on the premises that *atleast it won't crash*.

> - This implementation prevents broken links, but refs might change, changing the suggestions

It was the simplest way to achieve it, and I don't wanna make it complicated, so I'll probably just remove it. But I don't think occasional changing suggestions is a big deal. But after all, it is a bug. 

So here's my confusion:

- last search is mainly for UX, to not have a tiny search dialog
- The message "no recent search" is a bit confusing, it should probably be changed.
- it could be a useful feature, but current implementation has a *minor* bug.
- That shouldn't be a big issue, and a lot of people probably wouldn't even notice.

Tradeoffs everywhere.. What are your thoughts about it?

src/lib/output/themes/default/assets/typedoc/components/Search.ts
phoneticallySAARTHaK: Weird, it doesn't in chromium (brave)

src/lib/output/themes/default/assets/typedoc/components/Search.ts
```diff
@@ -198,11 +209,11 @@ function updateResults(
...
 
-    resultCount = res.length;
-
-    if (res.length === 0) {
-        const item = createStateEl(window.translations.theme_search_no_results);
-        results.appendChild(item);
+    if (res.length === 0 && searchText) {
+        const message =
+            window.translations.theme_search_no_results_found_for +
+            ` "<strong>${searchText}</strong>"`;
```
Gerrit0, line 215: searchText should be escaped here, this introduces an XXS issue.

src/lib/internationalization/locales/en.cts
```diff
@@ -522,4 +521,7 @@ export = {
         "This member is normally hidden due to your filter settings.",
     theme_hierarchy_expand: "Expand",
     theme_hierarchy_collapse: "Collapse",
+    theme_search_index_not_available: "The search index is not available",
+    theme_search_no_results_found_for: "No results found for", // for <search query>
```
Gerrit0, line 525: It would be better to implement this with a placeholder like the other ones, and fill the placeholder with `{0}` when writing to JS, so the frontend can do the same. This forces the search query to be last, which might not make sense for some languages.

phoneticallySAARTHaK: You mean like this?

`theme_search_no_results_found_for: "No results found for {0}` 

what should I pass as the argument in the Asset plugin?

https://github.com/TypeStrong/typedoc/pull/2831/files#diff-0a222463dab95a2fe2f10289f37d4fdb1d535f88a26bd90398d964a433c42fdbR43

phoneticallySAARTHaK: You mean like this?

`theme_search_no_results_found_for: "No results found for {0}` 

what should I pass as the argument in the Asset plugin?

https://github.com/TypeStrong/typedoc/pull/2831/files#diff-0a222463dab95a2fe2f10289f37d4fdb1d535f88a26bd90398d964a433c42fdbR43

Gerrit0: Yes, exactly, and then pass in `"{0}"` in the Asset plugin and do a `.replace` rather than `+` to create the message in the frontend

src/lib/output/themes/default/assets/typedoc/components/Search.ts
phoneticallySAARTHaK: Ironic how I emphasized "innerHTML" in the doc comment and still forgot to escape.

src/lib/output/themes/default/assets/typedoc/components/Search.ts
```diff
@@ -187,110 +233,81 @@ function updateResults(
...
             name = `<span class="parent">
-                ${boldMatches(row.parent, searchText)}.</span>${name}`;
+                ${highlightMatches(row.parent, searchText)}.</span>${name}`;
         }
 
         const item = document.createElement("li");
+        item.id = `tsd-search:${optionsIdCounter}-${i}`;
+        item.role = "option";
+        item.ariaSelected = "false";
         item.classList.value = row.classes ?? "";
```
phoneticallySAARTHaK, line 257: Figure out how to override filter styles to not affect search results

Gerrit0:
I've merged this into the beta branch

src/lib/output/themes/default/assets/typedoc/components/Search.ts
phoneticallySAARTHaK: I've used a static method, I doubt if that's going to cause any issue with inheritance. 
Otherwise I'm done with this PR

Gerrit0:
Thank you!

Issue #2675, opened 2024-08-19
Watch Inclusions


Metaxona:
## Search terms

Watch

## Expected Behavior

I expected watch to watch the files changes on

- typedoc.json
- config.projectDocuments files
- readme -> when not none

## Actual Behavior

changes made to the said files included above does not re-execute the build unless i quit watch mode and re-execute the command again

## Steps to reproduce the bug

if this is not a normal behavior tag me on this issue so i can provide a repo that recreates this bug, if this is the normal behavior
converting this to a feature request would be the next choice

## Environment

-   Typedoc version: ^0.26.5
-   TypeScript version: ^5.5.4
-   Node.js version: v20.11.0
-   OS: Ubuntu 22.04.4 LTS x86 (Specifics: Ubuntu Server)

Gerrit0:
This is expected as watch mode just uses typescript's watch with a hook to generate docs, there isn't any support for watching other changes. This is essentially the same limitation as what is preventing #1772

Metaxona:
so currently there is no way to solve it?

Gerrit0:
Correct, it isn't supported yet

pjeby:
For anyone encountering this issue, you can work around it using the `onchange` package from npm, and a build script like this:

```json
"watch-docs": "onchange --await-write-finish 3000 -i --kill \"*.md\"  \"guides/**/*.md\" \"typedoc.config.*\" \"typedoc/*\" -- typedoc --watch",
```

Of course you need to change the patterns and command line to something that makes sense for your project.  (e.g. the `typedoc/*` part is to catch changes to my custom CSS, and you might not even have such a directory.)

The `--await-write-finish` bit says to wait 3 seconds after a change to these files to make sure you're not changing something else, to avoid continually killing and restarting the process (e.g. if you have a build step that writes to your .md files or custom.css.)

shawninder:
If this isn't fixable, maybe a quick note about it in the docs would help?

Gerrit0:
There is a note - https://typedoc.org/documents/Options.Other.html#watch

> Note: This mode will only detect changes to files watched by the TypeScript compiler. Changes to other files (`README.md`, imported files with `@include` or `@includeCode`) will not cause a rebuild.

pjeby:
I did a little experimentation with the TS compiler API, and it has a way to add files to watch and get a callback when the files change.  

```diff
diff --git a/src/lib/application.ts b/src/lib/application.ts
[...]

PR #2836, opened 2025-01-26
Watch documents, readme, and CSS (#2675)

pjeby:
This PR is a proof-of-concept for enhanced watch support, covering:

- Project README
- Project documents
- Custom CSS file
- Custom JS file
- Config files
- `@include`/`@includeCode` files
- package.json
- Anything registered by plugins as a file to watch

with both soft and hard restarts.

It does *not* currently support:

- Reloading changed plugins
- Reloading modules imported or require'd from within config files
- Reloading ESM config files on Windows

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -552,7 +552,7 @@ export class Application extends AbstractComponent<
             this.logger.verbose(
                 `Watching ${nicePath(path)}, shouldRestart=${shouldRestart}`,
             );
-            if (!this.watchers.has(path)) return;
+            if (this.watchers.has(path)) return;
             this.watchers.set(
                 path,
                 host.watchFile(
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -506,6 +506,7 @@ export class Application extends AbstractComponent<
 
         let successFinished = true;
         let currentProgram: ts.Program | undefined;
+        let watches: ts.FileWatcher[] = []
 
         const runSuccess = () => {
             if (!currentProgram) {
@@ -528,6 +529,24 @@ export class Application extends AbstractComponent<
                     return;
                 }
                 const project = this.converter.convert(entryPoints);
+                watches.forEach(w => w.close())
+                watches = []
+                const lastProgram = currentProgram;
+                const addWatch = (path: string) => void watches.push(host.watchFile(path, () => {
+                    if (!currentProgram) {
+                        currentProgram = lastProgram
+                        if (successFinished) runSuccess()
+                    }
+                }))
+                for(const d of project.documents || []) {
+                    const path = project.files.getReflectionPath(d)
+                    if (path) addWatch(path)
+                }
+                const css = this.options.getValue("customCss")
+                if (css) addWatch(css)
+                const js = this.options.getValue("customJs")
+                if (js) addWatch(js)
+                if (project.readmeFile) addWatch(project.readmeFile)
                 currentProgram = undefined;
                 successFinished = false;
                 void success(project).then(() => {
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -231,7 +231,9 @@ export class Application extends AbstractComponent<
         readers.forEach((r) => app.options.addReader(r));
         app.options.reset();
         app.setOptions(options, /* reportErrors */ false);
-        await app.options.read(new Logger());
+        await app.options.read(new Logger(), undefined, (path) =>
+            app.watchConfigFile(path),
+        );
         app.logger.level = app.options.getValue("logLevel");
 
         await loadPlugins(app, app.options.getValue("plugin"));
@@ -265,7 +267,9 @@ export class Application extends AbstractComponent<
     private async _bootstrap(options: Partial<TypeDocOptions>) {
         this.options.reset();
         this.setOptions(options, /* reportErrors */ false);
-        await this.options.read(this.logger);
+        await this.options.read(this.logger, undefined, (path) =>
+            this.watchConfigFile(path),
+        );
         this.setOptions(options);
         this.logger.level = this.options.getValue("logLevel");
         for (const [lang, locales] of Object.entries(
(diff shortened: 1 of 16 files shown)


pjeby:
At this point I've figured out how to watch and reload basically anything that isn't an esm module, and have created a new draft of this PR with support for plugins to `app.watchFile()` during conversion or rendering, and that lets config readers track which files they looked for, etc., so that the watch can do a full restart when configuration files change or are added/removed etc.

If you're interested I can update this with that version or do a separate PR.  It's not hugely complicated but it currently touches 12 files, even if most of them only touch a few lines.  (And it changes the interface to `OptionsReader.read()` to add a callback to track options files that were read or whose existence was checked for.)

Let me know what you think.

Gerrit0:
Please do! This is similar to how I thought it might turn out someday.

The last time I looked at this I was thinking that an `app.fs` property which all (non-rendering) filesystem access goes through would make it easier to implement this.

pjeby:
Well, I suppose you *could* do that, but I really only needed to munge a few utilities.  I'll clean up what I've got and push it.

I did run into a couple of minor issues.  It looks like `_bootstrap()` and `bootstrapWithPlugins()` both call `.options.read()` and with my hot-reloading in effect for config files this causes them to be loaded/run twice.  I have an alternate reloading approach I could possibly use, but if there's not a reason to load the files twice maybe they shouldn't?  Also, `nicePath()` doesn't seem to work properly on Windows for absolute paths.

src/lib/cli.ts
```diff
@@ -76,7 +76,7 @@ async function run(app: td.Application) {
         app.convertAndWatch(async (project) => {
             app.validate(project);
             await app.generateOutputs(project);
-        });
+        }, main);
```
Gerrit0, line 79: This is an unfortunate API... tells me there's a better interface for this. I suspect it looks something like TS's WatchHost. This is probably fine for now, but is something I'll want to look at in the future.

src/lib/output/plugins/AssetsPlugin.ts
```diff
@@ -52,6 +52,7 @@ export class AssetsPlugin extends RendererComponent {
 
         if (this.customCss) {
             if (existsSync(this.customCss)) {
+                this.application.watchFile(this.customCss);
```
Gerrit0, line 55: So in findTsConfigFile, we end up calling watchFile for each checked path... whether or not it was read. Is there a reason to only call watchFile if the path exists here?

src/lib/utils/options/readers/typedoc.ts
```diff
@@ -83,10 +90,19 @@ export class TypeDocReader implements OptionsReader {
...
             try {
-                // On Windows, we need to ensure this path is a file path.
-                // Or we'll get ERR_UNSUPPORTED_ESM_URL_SCHEME
-                const esmPath = pathToFileURL(file).toString();
-                fileContent = await (await import(esmPath)).default;
+                if (process.platform === "win32") {
+                    // Node on Windows doesn't support the `?` trick for
+                    // cache-busting, so we need to use require()
+                    delete require.cache[require.resolve(file)];
+                    const mod = require(file);
```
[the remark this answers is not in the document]
pjeby: Oh darn.  I thought that worked on node 18 but apparently I was testing that with node 20.

src/lib/output/plugins/AssetsPlugin.ts
pjeby: Yeah, my bad.  I think I was just originally thinking "it's not looking for a file, these are config options that name a specific  file".  D'oh.

src/lib/utils/options/readers/typedoc.ts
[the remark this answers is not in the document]
pjeby: Changed to try require() on Windows and catch the ERR_REQUIRE_ESM.

src/lib/output/plugins/AssetsPlugin.ts
pjeby: Fixed in latest push.

src/lib/application.ts
```diff
@@ -506,16 +545,72 @@ export class Application extends AbstractComponent<
...
         let successFinished = true;
         let currentProgram: ts.Program | undefined;
+        let lastProgram = currentProgram;
+        let restarting = false;
+
+        this._watchFile = (path: string, shouldRestart = false) => {
+            this.logger.verbose(
+                `Watching ${path}, shouldRestart=${shouldRestart}`,
+            );
+            this.watchers.get(path)?.close();
```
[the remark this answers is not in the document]
pjeby: No, you caught it, I just also caught it and pushed a fix while you were reviewing I think.  :grin:

src/lib/cli.ts
pjeby: So, another idea I came up with for this was to change convertAndWatch to an async function that returns either true or false, where true means "restart watching" and false means "exit with an options error".

Combine that with a fork-and-loop in `bin/typedoc`, and voila: reloading config works fine because you're basically restarting the process.  All the cache-busting stuff goes away, too.  And the watch option doc loses all the caveats about indirect imports and so on.  Basically, if a file is denoted as shouldRestart, then it's a real restart of the process, not just a soft restart.  (Slower, of course, but with no caveats.)

On an unrelated note, I found a bug in my commit to fix the bug you mentioned earlier.  Will send it along later, along with the fork-to-reload version, if you think that's a better idea than the current API.

Gerrit0: I think that forking method is probably going to be better in the long run, please do!

pjeby: Pushed.  I haven't tested it extensively, barely more than a smoke test or two, but it handles the restarts and is interruptible with ^C and seems to shut down all its children (well, all one of them) properly.  Basically the main program only imports child_process so it doesn't seem any slower to start, either.  Let me know what you think.

Gerrit0: This seems much nicer, thank you! I'll play with it tomorrow and hopefully make a release then

Gerrit0:
This works amazingly well, thank you @pjeby!

Gerrit0:
Well, I spoke too soon... will hopefully have time to investigate it this weekend, but in case you have time sooner -- running with `--watch` in https://gitlab.com/hesxenon/ce-docs-how-to/-/tree/main results in constant rebuilds.

pjeby:
Uh, that project loops with typedoc 0.27.3 and 0.27.6 stock versions, so I don't think it's anything to do with this PR.  It's TypeScript that's looping the watch for whatever reason.

Gerrit0:
Oh that's fun, apparently TS doesn't like it when TypeDoc overrides some compiler options...

PR #2839, opened 2025-01-31
Move jp to ja and set jp as alias of ja

tats-u:
https://heistak.github.io/your-code-displays-japanese-wrong/
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#accessibility_concerns

`lang` attribute value for Japanese is `ja`, not `jp`.
`ja` is a language code stipulated in ISO-639-1 and `jp` is a *country* code.


Gerrit0:
Thanks!

PR #2843, opened 2025-02-02
Deprecate "jp"

tats-u:
- Hide "jp" from available languages
- Add Japanese messages for those who chose "jp" instead of "ja" (you can remove them after it turns to be unnecessary)
- Tweak message for unsupported languages (doesn't support → can't provide translations)

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -297,6 +297,11 @@ export class Application extends AbstractComponent<
             this.logger.info(
                 "You can define/override local locales with the `locales` option, or contribute them to TypeDoc!" as TranslatedString,
             );
+} else if (this.lang === "jp") {
+            this.logger.warn(
+                // Only Japanese see this. Meaning: "jp" is going to be removed in the future. Please designate "ja" instead.
+                "「jp」は将来削除されます。代わりに「ja」を指定してください。" as TranslatedString,
+            );
         }
 
         if (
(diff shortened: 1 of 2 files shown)


src/lib/application.ts
```diff
@@ -286,16 +286,28 @@ export class Application extends AbstractComponent<
...
             );
-            this.logger.info(
-                ("The supported languages are:\n\t" +
-                    this.internationalization
-                        .getSupportedLanguages()
-                        .join("\n\t")) as TranslatedString,
-            );
-            this.logger.info(
-                "You can define/override local locales with the `locales` option, or contribute them to TypeDoc!" as TranslatedString,
+            if (this.lang !== "jp") {
```
Gerrit0, line 291: Adding a warning for use of `jp` seems reasonable to me, but this over complicates things. It should still be supported for now and listed in the supported languages list.

tats-u: This is not run _now_ because jp.cts _still_ exists.
It will be run once you remove jp.cts.

Gerrit0: Which I won't until the next minor release. I don't want to include dead code in planning for a future deprecation.

tats-u: I see. How about commenting it out instead of removing it?

Gerrit0: No, commented out code is still dead code that shouldn't exist, it can be brought back if necessary in the future. When jp is removed, I will likely choose not to keep this message. It will be documented as a breaking change in the changelog, and the warning will have been around for several months at that point.

tats-u: I understand your plan.

Gerrit0:
Thanks!

PR #2846, opened 2025-02-08
Default theme: Tag-based visibilityFilter supports signatures

XeroAlpha:
# Changes

1. Introduces the capability to apply visibility filters based on (modifier) tags at the method signature, accessor signature, and index signature levels.

2. If visibility of every signature is affected by the same tag, this influence is now elevated to the method or accessor itself. This change ensures that when all signatures of a method/accessor are filtered out (hidden), the method/accessor title will also be hidden accordingly. This modification addresses the issue where method/accessor titles were still visible even though all their signatures were hidden.

3. Previously, the visibility of index signature container element was determined by whether the class they belonged to met the filter criteria. This behavior has been altered so that these container elements are no longer influenced by the filter settings.

# Known Issues

- Index signatures do not support "Inherited" and "External" filters. This limitation arises because the parent `DeclarationReflection` of an index signature `SignatureReflection` is the class it belongs to, instead of any method or property `DeclarationReflection`. Class `DeclarationReflection` is not `Inheritable`, since it cannot be a member of class/interface.

### DIFF
--- a/src/lib/output/themes/default/DefaultTheme.tsx
+++ b/src/lib/output/themes/default/DefaultTheme.tsx
@@ -503,37 +503,37 @@ export class DefaultTheme extends Theme {
 }
 
 function getReflectionClasses(reflection: Reflection, filters: Record<string, boolean>) {
-    const classes: string[] = [];
+    const classes = new Set<string>();
 
     // Filter classes should match up with the settings function in
     // partials/navigation.tsx.
     for (const key of Object.keys(filters)) {
         if (key === "inherited") {
             if (reflection.flags.isInherited) {
-                classes.push("tsd-is-inherited");
+                classes.add("tsd-is-inherited");
             }
         } else if (key === "protected") {
             if (reflection.flags.isProtected) {
-                classes.push("tsd-is-protected");
+                classes.add("tsd-is-protected");
             }
         } else if (key === "private") {
             if (reflection.flags.isPrivate) {
-                classes.push("tsd-is-private");
+                classes.add("tsd-is-private");
             }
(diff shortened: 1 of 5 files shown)


src/lib/output/themes/default/DefaultTheme.tsx
```diff
@@ -537,6 +534,18 @@ function getReflectionClasses(
...
+                const ownSignatures = reflection.getNonIndexSignatures();
+                // Check methods and accessors, find common tags, elevate
+                if (
+                    ownSignatures.length &&
+                    ownSignatures.every(
+                        (refl) =>
+                            refl.comment?.hasModifier(key as `@${string}`) || refl.comment?.getTag(key as `@${string}`),
+                    )
+                ) {
+                    classes.push(toStyleClass(`tsd-is-${key.substring(1)}`));
```
Gerrit0, line 547: This might add duplicates if there are multiple signatures (e.g. something with overloads)

Gerrit0 (changes requested):

Gerrit0:
Thanks!

src/lib/output/themes/default/DefaultTheme.tsx
XeroAlpha: Duplicates are possible, but not due to multiple signatures.

For example, if a reflection meets both `reflection.flags.isInherited` and `reflection.comment.hasModifier("@inherited")`, the same class `tsd-is-inherited` will be added twice.

```
<section class="tsd-panel tsd-member tsd-is-inherited tsd-is-inherited">

</section>
```

Issue #2853, opened 2025-02-15
Unresolved externals go to a page called `undefined`, and unhelpful warnings are generated


pjeby:
## Search terms

excludeExternals, Failed to resolve link in comment, undefined

## Expected Behavior

When there is a `{@link}` to an reference from an external module, and excludeExternals is enabled, no link should be generated in the documentation, and ideally no warning should be output since there is nothing that can be done to stop the warning except remove the link (which is still useful to have within e.g. VSCode, even if it's not being put into the docs).

## Actual Behavior

Typedoc generates the link with an href of `"undefined"`, and outputs a warning that the reference can't be resolved.

## Steps to reproduce the bug

As described above - {@link} to something exported from an external module.

Version 0.27.5 and up produces undefined links, 0.27.4 and earlier do not.

Versions prior to 0.26.7 do not output "[warning] Failed to resolve link to "whatever" in comment for whatever", while later versions do.

## Environment

-   Typedoc version: 0.26.7 and up for the warning, 0.27.5 and up for the `"undefined"` hrefs
-   TypeScript version: 5.2.2
-   Node.js version: 18.20.2
-   OS: Windows 10

pjeby:
It looks like 0.27.5 removed validity checking for `url` (in `MarkedPlugin.tsx`), and the fix for #2681  in 0.26.7 doesn't take into account the excludeExternals flag.  I'll work up a PR for both.

PR #2854, opened 2025-02-15
Fix `undefined` hrefs for excluded external refs

pjeby:
See #2853.  This fixes the broken links part by putting back in the undefined checks that were removed in 0.27.5.


Gerrit0:
Thanks!

Issue #2860, opened 2025-02-18
Wrong option help text


XeroAlpha:
```
> typedoc --help

 --lang                                       Sets the language to be used in generation and in TypeDoc's messages
 --suppressCommentWarningsInDeclarationFiles  Sets the language to be used in generation and in TypeDoc's messages

 --cascadedModifierTags                       Modifier tags which TypeDoc should recognize when parsing comments
 --modifierTags                               Modifier tags which TypeDoc should recognize when parsing comments
```

[suppressCommentWarningsInDeclarationFiles](https://github.com/TypeStrong/typedoc/blob/0bad04c55cb421ee9ea3a87ea75c31ca62f21cd2/src/lib/utils/options/sources/typedoc.ts#L746)
[cascadedModifierTags](https://github.com/TypeStrong/typedoc/blob/0bad04c55cb421ee9ea3a87ea75c31ca62f21cd2/src/lib/utils/options/sources/typedoc.ts#L809)

PR #2871, opened 2025-02-25
Update Chinese translations for beta

XeroAlpha:
Also fixes wrong description for `cascadedModifierTags` (#2860)


Gerrit0:
Thanks!

PR #2872, opened 2025-02-25
Add support for disabling local storage usage

jsmith2-coveo:
### Changes
***
Introduces ability to toggle usage of local storage from components and settings elements through `dataset` functionality of document similar to that of theme. UI checkbox added to settings menu that defaults to being hidden to allow for toggle of this feature via UI, otherwise accomplished through changes to the `document.documentElement.dataset` attributes directly.

#### Rationale Behind Changes:
***
Certain functional cookie requirements require that local storage not be used for legal compliance reasons. As such, having the ability to toggle the use of local storage for components based on users' cookie preferences is a priority.


site/development/local-storage.md
```diff
@@ -0,0 +1,27 @@
...
+**Note:** Enabling this feature will clear the local storage of the contents.
+
+## Programmatically
+
+Local storage can be disabled by setting the `disableLocalStorage` attribute in the `dataset` property
+of the `document.documentElement`.
+
+Specifically, use: `document.documentElement.dataset.disableLocalStorage`.
+
+**Note:** This attribute expects a boolean value, defaulting to `false`.
```
Gerrit0, line 21: I'm really not a fan of this, requiring the mutation observer feels incredibly hacky to me.

I'd rather define a global `TypeDoc` object set up by the frontend that exposes a function to disable local storage.

src/lib/output/themes/default/assets/typedoc/components/Accordion.ts
```diff
@@ -25,7 +29,16 @@ class AccordionImpl {
         for (const acc of this.accordions) {
             acc.open = open;
         }
-        storage.setItem(this.key, open.toString());
+        if (!this.disableStorage) {
+            storage.setItem(this.key, open.toString());
+        }
```
Gerrit0, line 34: I'm really not a fan of how this option has propagated to **every** instance of saving something to storage. A better implementation would instead be to have the storage module perform the correct action based on whether the option was set or not, so this option doesn't leak into every instance of saving state.

src/lib/output/themes/default/layouts/default.tsx
```diff
@@ -95,6 +95,7 @@ export const defaultLayout = (
             {context.hook("body.begin", context)}
             <script>
                 <Raw html='document.documentElement.dataset.theme = localStorage.getItem("tsd-theme") || "os";' />
+                <Raw html='document.documentElement.dataset.disableLocalStorage = "false";' />
```
Gerrit0, line 98: This should not be included in the default layout because that's the default. There's no point in duplicating the default and making every page bigger

Gerrit0 (changes requested):

phoneticallySAARTHaK:
> Certain functional cookie requirements require that local storage not be used for legal compliance reasons. As such, having the ability to toggle the use of local storage for components based on users' cookie preferences is a priority.

Is there a source for this? As per my (limited) knowledge, cookie preferences privacy concerns are for user tracking, I'm not sure if saving component state comes under that.

I'm also not sure if all websites must have a consent form or similar regardless of its use.

site/development/local-storage.md
jsmith2-coveo: I totally agree, I've updated the PR with a modified list of changes adding a `TypeDoc` object with exposed functionality to enable and disable local storage (for toggling).

src/lib/output/themes/default/layouts/default.tsx
jsmith2-coveo: Agreed, and removed with refactor.

jsmith2-coveo:
Saving of component toggle state, and theme via local storage would fall under functional cookie requirements, further details for which can be found [here](https://www.cookieyes.com/blog/functional-cookies/).

image

Gerrit0:
Other large documentation sites don't do anything like this, this seems like such a to-meet-compliance only feature, which makes me sad...

- https://doc.rust-lang.org/std/ - stores theme preferences like TypeDoc, no option to toggle
- https://docusaurus.io/docs - ditto. Links to https://opensource.fb.com/legal/cookie-policy/ for options, which recommends browser settings
- https://developer.mozilla.org/en-US/ ditto. Links to https://www.mozilla.org/en-US/privacy/websites/#cookies which says "You can accept or decline individual cookies in your preferences in the Tools/Options/Privacy history section", guessing that requires an account as the "Tools" link has no Options link under it for me.

src/lib/output/themes/default/assets/typedoc/Application.ts
```diff
@@ -13,9 +14,24 @@ declare global {
...
+            disableLocalStorage: () => void;
+            enableLocalStorage: () => void;
+        };
     }
 }
 
+window.TypeDoc ||= {
+    disableLocalStorage: () => {
+        storage.disable();
+        console.log("disable local storage option here");
```
Gerrit0, line 27: Why the console logs? Shouldn't be necessary...

src/lib/output/themes/default/assets/typedoc/utils/storage.ts
```diff
@@ -5,19 +5,41 @@
 export interface MinimalStorage {
     getItem(key: string): string | null;
     setItem(key: string, value: string): void;
+    removeItem(key: string): void;
+    clear(): void;
 }
 
 let _storage: MinimalStorage;
 
+const localStorageImpl: MinimalStorage = localStorage;
```
Gerrit0, line 14: This statement can throw on Safari in private mode, breaking the site. Just accessing the `localStorage` global is enough on some iOS versions.

src/lib/output/themes/default/assets/typedoc/utils/storage.ts
```diff
@@ -5,19 +5,41 @@
 export interface MinimalStorage {
     getItem(key: string): string | null;
     setItem(key: string, value: string): void;
+    removeItem(key: string): void;
+    clear(): void;
```
Gerrit0, line 9: These keys seem to be unused, remove?

src/lib/output/themes/default/assets/typedoc/utils/storage.ts
```diff
@@ -5,19 +5,41 @@
...
-export const storage = _storage;
+export const storage = {
+    getItem: (key: string) => _storage.getItem(key),
+    setItem: (key: string, value: string) => _storage.setItem(key, value),
+    removeItem: (key: string) => _storage.removeItem(key),
+    clear: () => _storage.clear(),
+    disable() {
+        localStorage.clear();
+        _storage = noOpStorageImpl;
+        console.log(_storage);
```
Gerrit0, line 39: Please don't keep console logs around in production code

Gerrit0 (changes requested):

src/lib/output/themes/default/assets/typedoc/utils/storage.ts
jsmith2-coveo: Since they're no longer used, refactored to be removed.

src/lib/output/themes/default/assets/typedoc/utils/storage.ts
jsmith2-coveo: I wrapped it back in the try/catch similar to how it was previously to prevent this, with the catch case being an assignment to the noOpStorageImpl

Gerrit0:
Thanks! I appreciate the minimal implementation

PR #2887, opened 2025-03-03
Markup/Styling improvements

phoneticallySAARTHaK:
- [ ] links in navigation take full width of the parent which makes it difficult to close the accordion unless clicked on the marker. ([DMT](https://github.com/TypeStrong/typedoc/discussions/2849) has fixed this)
- [x] Remove the empty anchor tag being used as an anchor target as it interferes with DOM tab order, and AT (Assistive Technologies)
- [ ] Contrast issues:
  - [ ] link color in dark theme
  - [ ] ~~alert color in light theme (Contrast ratio is 5, which is 0.5 more than the required value, but not good enough for low brightness)~~
  - [ ] Normal text color on accent color has poor contrast ratio
- [x] Align `summary` element contents
- [x] Remove `tabindex`, `aria-expanded`, `role` attributes from Index summary element
- [x] Add code blocks button and permalink focus styles to make them visible when being focused just like on hover


phoneticallySAARTHaK:
Is this a test case failing? I don't get any error in on the `test` run-script

Gerrit0:
Yes, maybe try merging in beta? I added some tests for the default theme rendering last night.

phoneticallySAARTHaK:
Oh, of course. PR changes theme output.

Gerrit0:
Changes so far LGTM

PR #2911, opened 2025-03-20
Update Chinese translations

XeroAlpha:


Gerrit0:
At some point I should probably look at some way to mark tags as showing up in the docs or not... some tags like `@enum` and all the tags added here are consumed when converting, so won't ever show up in the docs if used correctly.

scripts/sync_i18n.js
```diff
@@ -164,6 +165,50 @@ function listLocales() {
     return readdirSync("src/lib/internationalization/locales").map((n) => n.replace(/^(.+)\.\w+$/, "$1"));
 }
 
+function verifyTags() {
```
Gerrit0, line 168: What is this for...? There is a test at [src/test/utils/options/tsdoc-defaults](https://github.com/TypeStrong/typedoc/blob/master/src/test/utils/options/tsdoc-defaults.test.ts) which verifies the consistency between tsdoc.json and TypeDoc's lists of tags.

Gerrit0:
Thanks!

PR #2924, opened 2025-04-02
Use same chevron on index

bkeepers:
This is a very minor issue I noticed while working on a custom theme. Most of the site uses `chevronDown`, but for some reason the index uses `chevronSmall`. It looks weird when everything is collapsed and they don't match.

## Before

image

## After

Screenshot 2025-04-02 at 10 37 45 AM

### DIFF
--- a/src/lib/output/themes/default/partials/icon.tsx
+++ b/src/lib/output/themes/default/partials/icon.tsx
@@ -61,6 +61,7 @@ export interface Icons extends Record<ReflectionKind, () => JSX.Element> {
     checkbox(): JSX.Element;
     menu(): JSX.Element;
     search(): JSX.Element;
+    /** @deprecated */
     chevronSmall(): JSX.Element;
     anchor(): JSX.Element;
     folder(): JSX.Element;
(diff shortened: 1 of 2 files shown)


phoneticallySAARTHaK:
I was wondering the same, but then the index title is also small.

phoneticallySAARTHaK:
That icon is used only once, IIRC. Better to remove it entirely, and make the icon smaller with CSS if needed, IMO.

phoneticallySAARTHaK:
> I'd honestly rather have the index size follow the normal hierarchy rules than shrink the size with css - I've just never gotten around to making it do that

There's no special CSS targeting index title, It is using an `<h5>`, which makes it smaller due to default styling.

It would have produced an a11y error due to invalid heading order (as it does for hierarchy), but section "headings" are wrapped in a `<summary>` element, which has an implicit `role` of `button`, which means all content inside it becomes presentational. ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#summaries_as_headings))

bkeepers:
@Gerrit0 Want me to update this to make the index styles consistent?

Below is a screenshot of what I did in my custom theme (the index chevron is still different, but I was able to tweak it enough with CSS to make it less noticeable). If anyone is curious, here's [a draft of the site](https://opensoul.org/signalk-server/).

image

Gerrit0:
Go for it!

PR #2933, opened 2025-04-11
fix: Support `elementSummaries` for inlined union members

LekoArts:
Hello!

I've noticed a bug for a use case like this:

```ts
/**
 * @inline
 */
type TestReturn =
    /**
     * An apple a day keeps the doctor away.
     */
    | { kind: 'apple', isHealthy: true }
    /**
     * A donut a day keeps the doctor not away.
     */
    | { kind: 'sweets', isHealthy: false }

/**
 * Returns a random value
 */
export function test(): TestReturn {
    return Math.random() > 0.5
        ? { kind: 'apple', isHealthy: true }
        : { kind: 'sweets', isHealthy: false };
}
```

The `type.types` inside the signature doesn't receive the `elementSummaries` for the union type. I'm adding a failing test in this PR first to showcase the problem and show that it fails.

I'll spend some time to see if I can find the fix in this (so far unfamiliar) codebase. If not, I'll let you know 👍

### DIFF
--- a/src/test/converter2/issues/gh2933.ts
+++ b/src/test/converter2/issues/gh2933.ts
@@ -1,0 +1,21 @@
+/**
+ * @inline
+ */
+type TestReturn =
+    /**
+     * An apple a day keeps the doctor away.
+     */
+    | { kind: 'apple', isHealthy: true }
+    /**
+     * A donut a day keeps the doctor not away.
+     */
+    | { kind: 'sweets', isHealthy: false }
+
+/**
+ * Returns a random value
+ */
+export function test(): TestReturn {
+    return Math.random() > 0.5
+        ? { kind: 'apple', isHealthy: true }
+        : { kind: 'sweets', isHealthy: false };
+}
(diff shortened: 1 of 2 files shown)


LekoArts:
Mh, I do wonder, is the current behavior intentional because it's not possible?

Reading https://github.com/TypeStrong/typedoc/blob/4bf00b5f2557d59f352889fa5fbdd90ad3ec6d75/src/lib/models/types.ts#L1381-L1387 I'm wondering if what I want can even be achieved.

Gerrit0:
Yeah, this is intentional. Supporting comments on union elements only makes sense if TypeDoc is converting based on the type node rather than the actual type. This only happens in a few select cases, and only actually *needs* to happen for type aliases, so it is only valid to include it there.

TypeDoc doesn't support rendering the union comments anywhere except type alias types and I didn't want people expecting it to work in other locations, which is why that comment says what it does.

LekoArts:
Ok, thanks for enlightening me. Then I'll close it :)

PR #2941, opened 2025-04-23
Add localization for German

blutorange:
Adds localization for German. We want to generate our TypeDoc page in English and German, so I translated the messages.

For reference, I'm a native German speaker. German uses many English loan word, I used German equivalents where available.


Gerrit0:
Thank you!

PR #2972, opened 2025-07-07
Allow hyphenated JSDoc tags

jonathanhefner:
JSDoc tags may contain hyphens.  For example, `@TJS-type`, which is used by [`typescript-json-schema`](https://www.npmjs.com/package/typescript-json-schema).

---

~There are no existing tests for this function, but I can add a test for this change, if desired.  ([This test](https://github.com/TypeStrong/typedoc/blob/e8b6df3d83cbbe4010d5919ed72c6003c90036d1/src/test/utils-common/validation.test.ts#L35-L38) is close, but it is testing other functionality, not `isTagString` itself.)~

I stand corrected.  [This test](https://github.com/TypeStrong/typedoc/blob/master/src/test/utils/options/default-options.test.ts#L164) was [recently added](https://github.com/TypeStrong/typedoc/commit/1bf3217dbcf2e4a9ba3eb496296dd6d62db36d5f), explicitly disallowing hyphenated tags.  I'm unclear as to why though.  Is there a reason to reject hyphenated tags?

### DIFF
--- a/src/lib/converter/comments/blockLexer.ts
+++ b/src/lib/converter/comments/blockLexer.ts
@@ -263,7 +263,7 @@ function* lexBlockComment2(
                 if (lookahead !== pos + 1) {
                     while (
                         lookahead < end &&
-                        /[a-z0-9]/i.test(file[lookahead])
+                        /[a-z0-9-]/i.test(file[lookahead])
                     ) {
                         lookahead++;
                     }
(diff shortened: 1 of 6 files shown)


Gerrit0:
TypeDoc doesn't support hyphenated tags because TSDoc disallows it:

> Tag names start with an at-sign (@) followed by ASCII letters using "camelCase" capitalization.
> https://tsdoc.org/pages/spec/tag_kinds/

And no official JSDoc tag uses it - https://jsdoc.app/

In theory, I don't think I have an issue with adding hyphen to the permitted characters list -- this isn't enough though, the lexers in src/lib/converter/comments will also need to be updated.

jonathanhefner:
> the lexers in src/lib/converter/comments will also need to be updated.

Thanks for the pointer!  I've updated the lexers.

Gerrit0:
Thanks!

PR #2976, opened 2025-07-10
Handle relative links in `<source srcset>`

### LINKED ISSUE #2975: Relative links to images via <picture> <source srcset> do not work
## Search terms

Relative links images picture tag srcset attribute broken

## Expected Behavior

Given the following HTML in a Markdown document ([as per GitHub's recommendation](https://github.blog/developer-skills/github/how-to-make-your-images-in-markdown-on-github-adjust-for-dark-mode-and-light-mode/)):

```html
    <picture>
        <source media="(prefers-color-scheme: light)" srcset=".github/wagtail.svg">
        <source media="(prefers-color-scheme: dark)" srcset=".github/wagtail-inverse.svg">
        <img width="343" src=".github/wagtail.svg" alt="Wagtail">
    </picture>
```

Typedoc should copy both `wagtail.svg` and `wagtail-inverse.svg` to the `media` directory and render the HTML as:

```html
    <picture>
        <source media="(prefers-color-scheme: light)" srcset="media/wagtail.svg">
        <source media="(prefers-color-scheme: dark)" srcset="media/wagtail-inverse.svg">
        <img width="343" src="media/wagtail.svg" alt="Wagtail">
    </picture>
```

## Actual Behavior

Typedoc only copies `wagtail.svg` and renders the HTML like the following:

```html
    <picture>
        <source media="(prefers-color-scheme: light)" srcset=".github/wagtail.svg">
        <source media="(prefers-color-scheme: dark)" srcset=".github/wagtail-inverse.svg">
        <img width="343" src="media/wagtail.svg" alt="Wagtail">
    </picture>
```

## Steps to reproduce the bug

- Have a README with the above `<picture>` tag and files in the `.github` directory accordingly. You can grab them from https://github.com/wagtail/wagtail if needed.
- Run Typedoc

## Environment

- Typedoc version: 0.28.7
- TypeScript version: 5.6.3
- Node.js version: 22.11.0
- OS: macOS Sequoia 15.5

laymonage:
I tried this locally and it seems to work. Fixes #2975.

## Before

image

## After

image

### DIFF
--- a/src/lib/converter/comments/textParser.ts
+++ b/src/lib/converter/comments/textParser.ts
@@ -297,7 +297,7 @@ function checkReference(data: TextParserData): RelativeLink | undefined {
 }
 
 /**
- * Looks for `<a href="./relative">` and `<img src="./relative">`
+ * Looks for `<a href="./relative">`, `<img src="./relative">`, and `<source srcset="./relative">`
  */
 function checkTagLink(data: TextParserData): RelativeLink | undefined {
     const { pos, token } = data;
@@ -310,6 +310,11 @@ function checkTagLink(data: TextParserData): RelativeLink | undefined {
     if (token.text.startsWith("<a ", pos)) {
         data.pos += 3;
         return checkAttribute(data, "href");
+    }
+
+    if (token.text.startsWith("<source ", pos)) {
+        data.pos += 8;
+        return checkAttribute(data, "srcset");
     }
 }
 
--- a/src/test/comments.test.ts
+++ b/src/test/comments.test.ts
@@ -1636,6 +1636,38 @@ describe("Comment Parser", () => {
         );
     });
 
+    it("Recognizes HTML picture source srcset links", () => {
+        const comment = getComment(`/**
+        * <source media="(prefers-color-scheme: light)"  srcset="./test.png" >
+        * <source media="(prefers-color-scheme: dark)"  srcset="./test space.png"/>
+        * <source srcset="https://example.com/favicon.ico">
+        */`);
+
+        equal(
+            comment.summary,
+            [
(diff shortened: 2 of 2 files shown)


src/lib/converter/comments/textParser.ts
```diff
@@ -311,6 +311,11 @@ function checkTagLink(data: TextParserData): RelativeLink | undefined {
         data.pos += 3;
         return checkAttribute(data, "href");
     }
+
+    if (token.text.startsWith("<source ", pos)) {
+        data.pos += 8;
+        return checkAttribute(data, "srcset");
```
laymonage, line 317: The `<source>` tag supports:

- [`src` for `<audio>` or `<video>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source#src)
- [`srcset` for `<picture>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source#srcset)

Ideally we should support both, but I'm not familiar with how these functions work in the context of the parser. It seems that it mutates the `data` as it gets passed around, so I'm not sure how safe it is to e.g. store the original `pos` and call `checkAttribute` for `src` with the original `pos` after checking for `srcset`.

Gerrit0: Yeah, the amount of random state that this code needs to parse around is very unfortunate; it's gotten worse since the original implementation because markdown is *special*... I *think* that should be fine here; but I really need to revisit this at some point.

Unfortunately, doing that isn't good enough. `srcset` isn't just one `src`, it's a potentially comma separated list with additional metadata ([spec](https://html.spec.whatwg.org/multipage/images.html#srcset-attributes))

Gerrit0:
Thanks for this! I made a few updates to handle both src and srcset, and also to handle srcset with more than just a URL contained within it

PR #3035, opened 2025-11-12
Remove the @jsx tag from tsdoc.json as TSDoc now directly supports this tag.

iclanton:
See https://github.com/microsoft/tsdoc/pull/448

### DIFF
--- a/site/tags/typescript.md
+++ b/site/tags/typescript.md
@@ -13,6 +13,7 @@ will remove them from the generated documentation.
 - `@augments`
 - `@callback`
 - `@extends`
+- `@jsx`
 - `@type`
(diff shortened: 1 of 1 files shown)


site/tags/typescript.md
```diff
@@ -13,7 +13,6 @@ will remove them from the generated documentation.
 - `@augments`
 - `@callback`
 - `@extends`
```
Gerrit0: The wording here still correctly describes `@jsx`, it shouldn't be removed.

```suggestion
- `@extends`
- `@jsx`
```

Gerrit0:
Thanks!

Issue #3034, opened 2025-11-07
Custom theme icons are not used in the "On This Page" side nav


jonchardy:
## Search terms

plugin, theme, custom icon

## Expected Behavior

Custom theme icons should be shown in the sidebar as they are in the index and search results. 

## Actual Behavior

Default icons are used as the PageHeading interface doesn't store any information about icons from `getReflectionIcon`.

<img width="1204" height="334" alt="Image" src="https://github.com/user-attachments/assets/25028101-f56a-4d9b-a922-918c34905698" />

## Steps to reproduce the bug

See https://github.com/jonchardy/typedoc-icon-repro.

It seems as though more information needs to be stored on PageHeading for use in the navigation partial.
https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/themes/default/partials/navigation.tsx#L193

For example, the icon is saved for use in search results:
https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/plugins/JavascriptIndexPlugin.ts#L117
https://github.com/TypeStrong/typedoc/blob/master/src/lib/output/themes/default/assets/typedoc/components/Search.ts#L277

## Environment

- TypeDoc version: 0.28.14
- TypeScript version: 5.9.3
- Node.js version: 22.15.0
- OS: Windows 11

jonchardy:
PR submitted.

PR #3039, opened 2025-11-14
Fix: Custom theme icons are not used in the "On This Page" side nav

jonchardy:
Fixes https://github.com/TypeStrong/typedoc/issues/3034.

Uses `getReflectionIcon` instead of just looking up based on `ReflectionKind`.


Gerrit0:
Thank you!

PR #3048, opened 2025-12-09
feat(IncludePlugin): regions for tsx

a10nik:
Apparently including regions from TSX like `{@includeCode ./examples.tsx#component}` was not supported:

```
[error] @includeCode tag in MyComponent specified "./examples.tsx#component", but regions are not supported for this file extension.
```


a10nik:
Added jsx, mjs and cjs extensions as well while I'm at it

a10nik:
Tested a bit if vscode supports other region syntaxes inside JSX tags themselves, to see if we should add more regexes for JSX-like extensions. Looks like it doesn't.
image

Noticed another discrepance between vscode and IncludePlugin: IncludePlugin requires the closing comment to also have a regionName while it seems to be ignored in vscode:

image
This one is foldable no matter what the closing text is
image

Gerrit0:
Thanks!

The end region text discrepancy is unfortunate... I didn't implement the original feature, but I suspect the intent was to handle nested regions...

a10nik:
Actually, nested regions should work fine either way -- it's like bracket sequences. One doesn't need to tell each closing bracket which of the opening ones it refers to.

```
//#region a
//#region b

//#endregion (no way it can refer to region a, since b should close first)
```


I can only see its usefullness if we wanted to support partially overlapping regions like

```
//#region a
//#region b
//#endregion a
//#endregion b
```

However, I don't think it's important enough to open an issue about.

a10nik:
Btw when is the next release planned?

Gerrit0:
Unfortunately it'll likely be January; I've been very busy. NPM's decision to break literally everyone's CI automation arbitrarily and ignore everyone complaining is something I need to deal with fixing to get releases working again, and it isn't quite just a trivial fix without breaking beta/lts publishing. (Recurring 90 day breakage isn't something I have the patience to deal with, it needs to be fixed exactly once)

If you need features on master you should be able to install via a Git dependency.

PR #3056, opened 2025-12-23
Improved vertical positioning for svg text icons

kozi:
Replaced `dominant-baseline="central"` with `dy="0.35em"` to improve vertical positioning for svg text icons.

https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/dy

grafik


kentt9484-crypto (approved):

Gerrit0:
It's a real shame that asking the browser to do this using the knowledge it has of the font in use apparently has worse results than just hardcoding a random value that looks about right...

PR #3058, opened 2026-01-04
Normalize Paths When Resolving Entry Points to TypeScript Source Files

### LINKED ISSUE #3057: TypeDoc Fails to Resolve Entry Point Paths to TypeScript Source Files on Windows
## Search terms

Failed, resolve, entry, point, path, typescript, source.

## Expected Behavior

TypeDoc should automatically discover entry points according to the `exports` field in `package.json` and use `tsconfig.json` options to map the JavaScript files back to the original TypeScript source, as described in [the documentation](https://typedoc.org/documents/Options.Input.html).

## Actual Behavior

On Windows, each entry point results in a warning of `Failed to resolve entry point path {0} from package.json to a TypeScript source file`. This is because, in `src/lib/utils/entry-point.ts`, the JavaScript to TypeScript source map `jsToTsSource` is constructed with keys that are normalized via `normalizePath`, but when the map is later queried on line 90 (at time of writing), the `path` is not similarly normalized (forward slashes are used rather than backwards ones).

## Steps to reproduce the bug

1. On a Windows machine in a project that uses `exports`, execute `typedoc`. No configuration is needed.

## Environment

- TypeDoc version: 0.28.15.
- TypeScript version: 5.9.3.
- Node.js version: 25.2.1.
- OS: Windows 11.

Lakuna:
Fixes #3057.

### DIFF
--- a/src/lib/utils/entry-point.ts
+++ b/src/lib/utils/entry-point.ts
@@ -87,7 +87,7 @@ export function inferEntryPoints(logger: Logger, options: Options, programs?: ts
     for (const [name, path] of pathEntries) {
         // Strip leading ./ from the display name
         const displayName = name.replace(/^\.\/?/, "");
-        const targetPath = jsToTsSource.get(path) || resolveDeclarationMaps(path) || path;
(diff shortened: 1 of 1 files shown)


Gerrit0:
Thanks!

PR #3078, opened 2026-02-17
Validate file registry paths during `validate()` for `--emit none`

jonathanhefner:
**:warning: Disclaimer:** This is vibe-coded.  I did my best to steer Claude to follow the conventions of this project, and the result looks reasonable to me, but I am very open to feedback.  I would like to get this issue fixed, but I don't have a strong opinion on how it should be fixed.  If you would prefer me to file an issue instead, let me know.  For additional context, see https://github.com/modelcontextprotocol/typescript-sdk/pull/1546.

---

When `--emit none` is used with `treatWarningsAsErrors`, warnings about relative links pointing to directories were silently suppressed because the only `isFile()` check lived in `AssetsPlugin.onRenderEnd()`, which is skipped entirely when rendering is disabled.

This adds a `validateFilePaths()` validation pass that runs during `Application.validate()`, gated on the existing `validation.invalidPath` option. It iterates all registered file paths that lack a reflection association (preserving packages-mode directory links which map to reflections) and emits `validationWarning()` for non-file entries.

Because `validate()` runs before the `--emit none` gate in `cli.ts`, directory-pointing relative links now produce warnings regardless of emit mode.

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -41,6 +41,7 @@ import { fileURLToPath } from "url";
 import { createRequire } from "module";
 import { Outputs } from "./output/output.js";
 import { validateMergeModuleWith } from "./validation/unusedMergeModuleWith.js";
+import { validateFilePaths } from "./validation/filePaths.js";
 import { diagnostic, diagnostics } from "./utils/loggers.js";
 import { ValidatingFileRegistry } from "./utils/ValidatingFileRegistry.js";
 import { Internationalization } from "./internationalization/internationalization.js";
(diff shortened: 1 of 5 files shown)


Gerrit0:
The feature request makes sense!

Thank you for the disclosure, this looks very similar to what I'd have written. I have just one complaint - warnings will now be emitted twice since AssetsPlugin wasn't updated to remove the now redundant log.

This will slightly change behavior as previously warnings would not be emitted for reflections marked with `@hidden` which are removed from the docs before rendering. I think this is probably fine, other warnings will be printed for hidden reflections.

It's also worth mentioning that this will not necessarily make it so that `--emit none` shows all warnings which would be generated when rendering. The other warnings in AssetsPlugin.onRenderBegin are an obvious example, as are warnings in JavaScriptIndexPlugin, or warnings which are emitted during rendering by any plugins you have enabled.

jonathanhefner:
@Gerrit0 Sorry this fell through the cracks.  If I understood correctly, you just wanted the `else` branch removed here: https://github.com/TypeStrong/typedoc/pull/3078/changes/3f7880872fdfd807f064c5b0e768f3ebf25e9720 , is that correct?

Gerrit0:
Thanks! Yes, that's what I was after

PR #3090, opened 2026-04-02
Support --commentStyle triple-slash

marijnh:
Which acts like 'line', but ignores comment lines that don't start with (precisely) three slashes.


Gerrit0:
Thanks!

PR #3091, opened 2026-04-05
Add French translations for locales

hyosua:
This PR adds complete French translations for the TypeDoc interface, messages, and help descriptions. Verified locally with `--lang fr` and on the generated HTML.


Gerrit0:
Thank you!

PR #3072, opened 2026-02-08
feat(theme): add id attributes to group headings for linking

wotan-allfather:
Add id attributes and anchor icons to group headings in the default theme, allowing users to link directly to specific groups in the documentation.

This includes:
- h2 group headings in members.tsx (member groups)
- h2 group headings in moduleReflection.tsx (module groups)
- h3 group headings in index.tsx (index sections)

Example: modules.html#Runtime_Guards now links directly to that group.

Closes TypeStrong#3029

PR #3100, opened 2026-05-03
feat(theme): add id attribute to group/category section headings

### LINKED ISSUE #3029: Group headings in HTML output don't have id attributes for linking
### Search Terms

group heading id, group anchor, categorizeByGroup linking, group permalink, @group tag linking

### Expected Behavior

When using `@group` tags in JSDoc comments with `categorizeByGroup: true` in TypeDoc configuration, the generated group headings in the HTML output should have `id` attributes (similar to how individual symbols have them), allowing them to be directly linked to via fragment identifiers.

For example, a group heading like:
```html
<h2>Runtime Guards</h2>
```

Should be rendered as:
```html
<h2 id="Runtime_Guards">Runtime Guards</h2>
```

This would enable direct linking to specific groups in documentation (e.g., `modules.html#Runtime_Guards`).

### Actual Behavior

Group headings are rendered without `id` attributes, making it impossible to create stable, direct links to specific groups in the generated documentation.

Current output:
```html
<h2>Runtime Guards</h2>
```

Individual symbols within groups do have `id` attributes and can be linked directly, but the group headings themselves cannot.

### Steps to reproduce the bug

**Installed packages:**
- `typedoc`: 0.28.14
- `typescript`: 5.9.3
- `node`: 22.14.0

**Source code with `@group` tag:**
```typescript
/**
 * Checks if a value is not `undefined` or `null`.
 *
 * @group Runtime Guards
 *
 * @typeParam T The type when the value is defined
 *
 * @returns `true` if the value is neither `undefined` nor `null`
 */
export function isDefined<T>(value: T): value is NonNullable<T> {
  return value !== undefined && value !== null
}
```

**typedoc.json:**
```json
{
  "includeVersion": true,
  "sort": ["source-order"],
  "categorizeByGroup": true,
  "groupOrder": [
    "Interfaces",
    "Type Aliases",
    "Runtime Guards",
    "Value Guards",
    "Value Casts",
    "Structural Utilities",
    "Error Utilities",
    "*"
  ]
}
```

**tsconfig.json:**
```json
{
  "compilerOptions": {
    "baseUrl": "src",
    "outDir": "dist",
    "strict": true,
    "target": "ES2022",
    "lib": ["ES2022"],
    "module": "ES2022",
    "moduleResolution": "node",
    "esModuleInterop": true,
    "declaration": true,
    "declarationMap": true,
    "sourceMap": true,
    "skipLibCheck": true
  },
  "include": ["src"],
  "exclude": ["**/*.*.ts"]
}
```

**Running TypeDoc:**
```bash
npx typedoc
```

**Generated HTML (docs/modules.html):**
```html
<h2>Runtime Guards</h2>
<!-- No id attribute present -->
```

**Individual symbols do have ids:**
```html
<dt class="tsd-member-summary" id="isdefined">
  <span class="tsd-member-summary-name">
    <!-- ... -->
  </span>
</dt>
```

### Environment

- **TypeDoc version:** 0.28.14
- **TypeScript version:** 5.9.3
- **Node.js version:** 22.14.0
- **OS:** macOS (Darwin 24.6.0)

Metbcy:
Closes #3029.

### Summary

Adds an `id` attribute to the `<h2>` group/category section headings rendered by the default theme's `members` and `moduleReflection` partials so that section headers can be linked to directly via fragment URLs (e.g. `…/MyClass.html#methods`).

### Changes

- `src/lib/output/themes/default/partials/members.tsx` — slug `section.title` via `context.slugger.slug(...)` and pass it to the heading `<h2>` as `id`.
- `src/lib/output/themes/default/partials/moduleReflection.tsx` — same treatment for the module-reflection sections (skipped for the unnamed/none section, which has no title to slug).
- `CHANGELOG.md` — Unreleased note under Features.
- Snapshot specs regenerated with `pnpm rebuild_specs` (only the new `id` attribute is added to the `h2` tag — no other changes).

The slugger is the existing per-page `Slugger` instance, which already tracks heading collisions for member anchors, so section anchors will be deduplicated against everything else on the page automatically.

### Testing

- `pnpm lint` — clean.
- `pnpm build` — clean.
- `pnpm test` — 1046 passing.
- `pnpm rebuild_specs` — snapshot diff is exclusively the new `id` attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ title: Changelog
 ### Features
 
 - Introduced `generateOutputsBegin` and `generateOutputsEnd` events on `Application` for plugin use.
+- Group/category section headings (`<h2>`) in the default theme now include an `id` attribute so they can be linked to via fragment identifiers (e.g. `modules.html#runtime-guards`), #3029.
 
 ## v0.28.19 (2026-04-12)
 
--- a/src/lib/output/themes/default/partials/index.tsx
+++ b/src/lib/output/themes/default/partials/index.tsx
@@ -2,14 +2,22 @@ import { classNames, getMemberSections, isNoneSection, type MemberSection, rende
 import type { DefaultThemeRenderContext } from "../DefaultThemeRenderContext.js";
 import { i18n, JSX } from "#utils";
 import type { ContainerReflection } from "../../../../models/index.js";
+import { anchorIcon } from "./anchor-icon.js";
 
 function renderSection(
-    { urlTo, reflectionIcon, getReflectionClasses, markdown }: DefaultThemeRenderContext,
+    context: DefaultThemeRenderContext,
     item: MemberSection,
 ) {
+    const { urlTo, reflectionIcon, getReflectionClasses, markdown, slugger } = context;
+    const sectionAnchor = !isNoneSection(item) ? slugger.slug(item.title) : undefined;
     return (
         <section class="tsd-index-section">
-            {!isNoneSection(item) && <h3 class="tsd-index-heading">{item.title}</h3>}
+            {!isNoneSection(item) && (
+                <h3 class="tsd-index-heading tsd-anchor-link" id={sectionAnchor}>
+                    {item.title}
+                    {anchorIcon(context, sectionAnchor)}
+                </h3>
+            )}
             {item.description && (
                 <div class="tsd-comment tsd-typography">
                     <JSX.Raw html={markdown(item.description)} />
--- a/src/lib/output/themes/default/partials/members.tsx
+++ b/src/lib/output/themes/default/partials/members.tsx
@@ -18,12 +18,13 @@ export function members(context: DefaultThemeRenderContext, props: ContainerRefl
                 }
 
                 context.page.startNewSection(section.title);
+                const sectionAnchor = context.slugger.slug(section.title);
 
                 return (
                     <details class="tsd-panel-group tsd-member-group tsd-accordion" open>
                         <summary class="tsd-accordion-summary" data-key={"section-" + section.title}>
                             {context.icons.chevronDown()}
-                            <h2>
+                            <h2 id={sectionAnchor}>
                                 {section.title}
                             </h2>
                         </summary>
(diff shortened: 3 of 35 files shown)


Gerrit0:
Thank you! This is a good start, however these ID attributes are only usable if someone goes and figures out what they'll be with developer tools. We should also add an index link like #3072 did, before becoming abandoned.

Metbcy:
Thanks for the review @Gerrit0! Pushed dfab33f to address the feedback:

- Added the visible anchor icon + `tsd-anchor-link` class to the section `<h2>` in `members.tsx` and `moduleReflection.tsx`, mirroring #3072.
- Extended the same treatment to the `<h3>` section headings in `index.tsx` so the in-page index entries are also linkable. `None` sections remain unanchored.
- The per-page slugger naturally disambiguates collisions between the index entry and the later member section (e.g. `#methods` on the index, `#methods-1` on the heading).

Snapshots regenerated; full test suite (1046) + lint pass locally. Let me know if you'd like the index/members slugs swapped (members getting the bare slug, index the `-1` suffix) or any other tweak.

Gerrit0:
My apologies for the delay here, life has been busy.

The anchors do not seem to display correctly now within the page:

image

Otherwise, this looks good to me

Metbcy:
> My apologies for the delay here, life has been busy.
> 
> The anchors do not seem to display correctly now within the page:
> 
> image
> Otherwise, this looks good to me

No worries man, life gets busy. Let me dig in and see if I can fix the anchoring.

Metbcy:
Fixed in 21159e7. Root cause was a stale rule in `static/style.css`:

```css
.tsd-accordion-summary a { width: calc(100% - 1.5rem); }
```

That was written for the nav accordion's direct-child `<a>`, but the descendant selector also matched the new `<a.tsd-anchor-icon>` nested inside the `<h2>` in section summaries, stretching it to fill the row and wrapping it to a second line.

Tightened to a direct-child combinator (`.tsd-accordion-summary > a`) so the nav case is unaffected and the anchor icon falls back to its normal `.tsd-anchor-icon` styling (inline-flex, 1em SVG, hidden until hover).

Anchor `<a>` is now 24px wide (was 163px), heading 28.8px tall (was 57.6px, wrapped). 1046 tests passing.

Gerrit0:
Thanks!

Issue #3103, opened 2026-05-26
perf: tracking issue for several typedoc speedup PRs


StoneCypher:
# perf: tracking issue for several typedoc speedup PRs

> **Note to gerrit0** — first, thank you. I've relied on typedoc for roughly ten years now, across more projects than I can count, and it has always Just Worked. I hope this report and the accompanying PR stack are useful.

Yes, it's Claude slop.

## Summary

I ran a CPU + wall-clock profile of typedoc generating its own documentation. Several distinct, independent wins surfaced. None of them are large in isolation; collectively they look like a **~15-25 % reduction in wall-clock time** on a representative workload, with the biggest single piece coming from cold-start module-resolution rather than any specific phase.

This issue is the parent for an 8-PR stack covering:

1. **Async git** — `spawnSync("git", …)` in `SourcePlugin.onBeginResolve` blocks the conversion phase for ~673 ms self-time. Parallelizing the four independent git calls in `tryCreateRepositoryAsync` recovers most of that.
2. **Async parallel page writes** — the per-page `writeFileSync` inside `renderDocument` runs 271 times sequentially; sibling output plugins already use async `writeFile` + `Promise.all`.
3. **JSX string builder** — `renderElement` uses a per-call string accumulator pattern that allocates new strings on every `+=`. A shared `string[]` buffer joined once reduces both render-time work and GC pressure (3.2 % of total runtime today is in GC).
4. **Memoization caches** — markdown content cache (518 ms today in `markdown-it` calls), `normalizePath` already-normalized fast-path, `createSymbolIdImpl` symbol-keyed WeakMap.
5. **Bundle CLI with esbuild** — Node currently spends ~440 ms in module resolution (`detectModuleFormat`, `internalModuleStat`, `getPackageScopeConfig`, `cjs-module-lexer`) just to load typedoc's own JS modules. A single-file bundle for the CLI entry path eliminates this.

The PRs split into two independent groups. **Group A** (4 PRs) is the async-git refactor — sequential, because each PR consumes APIs added by the previous one. **Group B** (4 PRs) is the additional wins — mutually independent, can land in any order.

None of the changes break the plugin API contract (`EventDispatcher.trigger` is sync and ignores listener return values — that constraint is what shaped the deferral pattern used in PR-2 and PR-5).

## Why now / why this fix

[...]

PR #3104, opened 2026-05-26
perf(converter): add async git infrastructure

StoneCypher:
## What

Infrastructure: adds an async-git surface (`spawnAsync`, `gitAsync`, `gitIsInstalledAsync`, `GitRepository.tryCreateRepositoryAsync`, `RepositoryManager.getRepositoryAsync`) alongside the existing sync API in `src/lib/converter/utils/`. No existing in-tree caller is changed — the sync surface stays the authoritative path and continues to serve `SourcePlugin.onBeginResolve`.

## Why

This is the first PR of an 8-PR stack tracked by #3103. The CPU profile attributes ~673 ms of self-time per typedoc-on-typedoc run to blocking `child_process.spawnSync` calls in `SourcePlugin.onBeginResolve`. The follow-up PR will flip the consumer to use this new async surface — that's where the parallelism win lands. Splitting the infrastructure out keeps the load-bearing review focused on a smaller diff.

## How

- New file `src/lib/converter/utils/repository-async.ts` — `spawn`-based (no shell, no command-string parsing) Promise wrapper that mirrors the no-throw contract of the existing sync `git()` helper.
- `gitAsync` is a thin convenience over `spawnAsync` that hardcodes the executable to `git`. `gitIsInstalledAsync` is a cached one-shot detection.
- `GitRepository.tryCreateRepositoryAsync` fires its four independent git calls (`branch --show-current`, `rev-parse HEAD`, `remote get-url`, `ls-files -z`) in parallel via `Promise.all` instead of sequentially.
- `GitRepository`'s constructor is refactored to be data-only (three field assignments); the `ls-files` work moved into the sync `tryCreateRepository` factory itself to preserve existing behavior end-to-end.
- `RepositoryManager.getRepositoryAsync` caches in-flight promises; a cycle-guard in `getRepositoryFolderAsync` prevents the self-referential promise that would otherwise occur when \`repoDir === dir\`.

## Perf impact

N/A — infrastructure PR; no consumer flipped yet. The next PR in the stack carries the measurable wall-clock change.

## Test plan

- [x] Existing `Repository` test suite still passes (19 passing + 6 pending; the pending tests are pre-existing Windows symlink skips).
- [x] New `src/test/repository-async.test.ts` covers `spawnAsync` (success / non-zero exit / missing binary with `errorCode === ENOENT`) and `gitAsync` / `gitIsInstalledAsync` (success / cached behavior / cache reset for tests).
- [x] New cases added to `Repository.test.ts` cover `tryCreateRepositoryAsync` happy path, explicit `sourceLinkTemplate`, literal `gitRevision` pass-through, no-commits returns `undefined`, and the sync/async cross-check (\`getRepositoryAsync\` returns same repo as sync variant).

## Rollback

Revert this PR. The sync API in \`repository.ts\` is still the authoritative path; nothing in production code currently uses any of the new async exports.

## Stack context

Part of the async-git refactor stack tracked by #3103.

- Previous PR: none (first in stack)
- Next PR: \`perf(converter): resolve source URLs in parallel after conversion ends\` — flips \`SourcePlugin\` to consume this surface; carries the perf numbers

Blocks #3103

### DIFF
--- a/src/lib/converter/utils/repository-async.ts
+++ b/src/lib/converter/utils/repository-async.ts
@@ -38,8 +38,8 @@ export function spawnAsync(command: string, args: string[]): Promise<SpawnResult
         let stderr = "";
         child.stdout.setEncoding("utf-8");
         child.stderr.setEncoding("utf-8");
-        child.stdout.on("data", (chunk) => { stdout += chunk; });
-        child.stderr.on("data", (chunk) => { stderr += chunk; });
+        child.stdout.on("data", (chunk: string) => { stdout += chunk; });
+        child.stderr.on("data", (chunk: string) => { stderr += chunk; });
         child.once("error", (err: NodeJS.ErrnoException) => {
             resolve({ status: null, stdout, stderr, errorCode: err.code ?? "EUNKNOWN" });
         });
--- a/src/lib/converter/utils/repository-async.ts
+++ b/src/lib/converter/utils/repository-async.ts
@@ -1,12 +1,37 @@
 import { spawn } from "node:child_process";
 
-export interface GitSpawnResult {
+/**
+ * Result of a non-throwing async child-process spawn.
+ *
+ * `status: null` means the child failed to start (binary not found, permission denied, etc.);
+ * `errorCode` carries the underlying `NodeJS.ErrnoException` code in that case (e.g. `"ENOENT"`,
+ * `"EACCES"`, `"EMFILE"`, or `"EUNKNOWN"` when no code was provided).
+ * `status: number` is the child's exit code (zero or non-zero) from a normal close — `errorCode`
+ * is absent in that case.
+ */
+export interface SpawnResult {
     status: number | null;
     stdout: string;
     stderr: string;
+    errorCode?: string;
 }
 
-export function gitSpawn(command: string, args: string[]): Promise<GitSpawnResult> {
+/**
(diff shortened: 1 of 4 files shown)


Gerrit0:
I don't want to merge partially done work. This needs to be feature complete before I'll spend time on it - temporary dead code has a way of becoming forgotten dead code

StoneCypher:
there's nothing partially done here.  the reason i didn't push the other two prs is that there's a request for comment in the base

Gerrit0:
GitHub got confused by a rebase, this was merged with #3114; thank you for the contribution!

StoneCypher:
i'm really sorry i just left these in your lap.  it was a long month at work.  i started doing my open source rounds this morning and my plan was to finish them today, as i have today off for the fourth (but i imagine you're in the same boat)

it's just that i went for my own language first, then codemirror.  then i was going to do something in monarch, then it was on to you, then peggy.

i feel like i left a smoking pile in your lap and i feel genuinely bad about that

i should have finished these to your standards

Gerrit0:
No worries! Everyone has busy times, I also have had little time for open source in the past month or so; it's hard to make much progress if you only have like 10 minutes at the end of the day, I think the bundling took me like 7 different days before last night I finally had a couple hours to sit down and actually dig into what was going on.

OSS contributions _never_ come with an obligation; don't feel bad about leaving things!

PR #3109, opened 2026-05-27
perf(converter): resolve source URLs in parallel after conversion ends

StoneCypher:
## What

The load-bearing perf change of the async-git stack. `SourcePlugin.onBeginResolve` no longer calls git inline — it pushes source refs onto a deferred queue. `Application.convert()` then calls a new `Converter.resolveDeferredSourceUrls()` after `converter.convert()` returns, which awaits parallel git operations via the async surface added in #3104.

**Stacks on #3104.** Until that merges, this PR's diff includes #3104's commits; once #3104 lands, this diff narrows to just the consumer-flip changes.

## Why

CPU profile attributes ~673 ms of self-time per typedoc-on-typedoc run to blocking `child_process.spawnSync` in `SourcePlugin.onBeginResolve`. The four git operations per repo ran sequentially. After this PR they run in parallel via `Promise.all` inside `tryCreateRepositoryAsync` (from #3104), and the sequential blocking on the JS main thread goes away.

## How

1. `SourcePlugin.onBeginResolve` — the inner URL-resolving loop now only normalizes `source.fileName` and pushes onto a new `deferredSources: SourceReference[]` queue.
2. `SourcePlugin.resolveDeferredUrls()` — new async method that drains the queue using `gitIsInstalledAsync` + `getRepositoryAsync` (cache-hitting after first lookup per directory).
3. `Converter.resolveDeferredSourceUrls()` — new async forwarder; the `SourcePlugin` instance is now stored on `Converter`.
4. `Application.convert()` — awaits `this.converter.resolveDeferredSourceUrls()` after `converter.convert()` returns; logs the time.

**Lifecycle subtlety:** `EVENT_END` fires synchronously inside `Converter.convert()` before it returns. The original `onEnd` did `delete this.repositories`, which would nuke the `RepositoryManager` before `resolveDeferredUrls` (called after `Application.convert()` reaches its await) could use it — every snapshot test would have failed with empty URLs. The cleanup of `this.repositories` moved into `resolveDeferredUrls`; `fileNames.clear()` stays in `onEnd`. A comment explains.

## Perf impact

**Measured on the reference machine (Windows 11, Node 24.11, NVMe SSD), workload `node bin/typedoc --options .config/typedoc.json`, 5 runs each back-to-back:**

\`\`\`
Condition                       run1    run2    run3    run4    run5    median    min
─────────────────────────  ───────  ───────  ───────  ───────  ───────  ──────  ─────
master (sync git, baseline)  11480    9951    9947   10123    9878    9951    9878
this PR (async parallel)      8418    8378    7838    8271    8240    8271    7838

Median delta:  -1680 ms  (-16.9 %)
Min delta:     -2040 ms  (-20.7 %)
\`\`\`

Run-to-run noise is ~1.5 s peak-to-peak on this Windows box (Defender / SearchIndexer file I/O); the 1.68 s median improvement is well outside that.

Improvement exceeds the 673 ms `spawnSync` self-time attribution alone — most likely because the sync calls also stalled GC and microtask flushes on the main thread, so async-ifying them recovers more than just their direct cost.

Linux/macOS savings will be smaller in absolute terms (cheaper git startup) but the parallelization win on multi-repo workloads (`--entryPointStrategy=packages`) should be comparable or larger (cross-repo concurrency stacks).

## Test plan

- [x] Existing converter snapshot tests still pass (1051 passing / 9 pending / 0 failing on `mocha.fast` — `source.url` values byte-identical to pre-PR).
- [x] ESLint clean.
- [x] dprint check clean.
- [x] No new IDE diagnostics.
- [ ] CI: full mocha suite green across Node versions.

## Out of scope (follow-up PRs)

- Watch mode (`convertAndWatch`) and `--entryPointStrategy=packages` (`_convertPackages`) bypass `Application.convert()`; next PR wires the new step into both.
- Removing the now-unused sync `git()` / `gitIsInstalled` / `tryCreateRepository` / `getRepository` helpers — also a follow-up after watch/packages mode is wired.

## Open question for a maintainer

Today `source.url` is populated synchronously inside `EVENT_RESOLVE_BEGIN`, so later converter events see it set. After this PR it's populated *after* the converter has fired all events — only `Application.convert()`'s returned-promise consumers see it set. Is the "URLs visible to in-converter events" guarantee a stable contract, or is "URLs populated by the time the `Application.convert()` promise resolves" sufficient? Tests in `testRendererUtils.ts` and `converter.test.ts` were updated to await the new step; if any other helper reaches into the project mid-convert it'll need the same.

## Rollback

Revert this PR. The async infrastructure in #3104 stays in place but unused; the sync API would re-take ownership of URL resolution.

## Stack context

Part of the typedoc speedup stack tracked by #3103.

- Group A (stacked): #3104 → **this PR** → upcoming watch/packages wiring → upcoming sync-removal cleanup.
- Group B (independent): #3105 #3106 #3107 #3108.

Blocks #3103

### DIFF
--- a/src/lib/application.ts
+++ b/src/lib/application.ts
@@ -457,6 +457,13 @@ export class Application extends AbstractComponent<
         this.logger.verbose(
             `Finished conversion in ${Date.now() - startConversion}ms`,
         );
+
+        const startSourceUrls = Date.now();
+        await this.converter.resolveDeferredSourceUrls();
+        this.logger.verbose(
+            `Resolving source URLs took ${Date.now() - startSourceUrls}ms`,
+        );
+
         return project;
     }
 
--- a/src/lib/converter/converter.ts
+++ b/src/lib/converter/converter.ts
@@ -250,6 +250,8 @@ export class Converter extends AbstractComponent<Application, ConverterEvents> {
     /** @internal @hidden */
     includePlugin: IncludePlugin;
 
+    private sourcePlugin!: SourcePlugin;
+
     constructor(owner: Application) {
         super(owner);
 
@@ -303,7 +305,7 @@ export class Converter extends AbstractComponent<Application, ConverterEvents> {
         new InheritDocPlugin(this);
         new LinkResolverPlugin(this);
         new PackagePlugin(this);
-        new SourcePlugin(this);
+        this.sourcePlugin = new SourcePlugin(this);
         new TypePlugin(this);
         this.includePlugin = new IncludePlugin(this);
         new MergeModuleWithPlugin(this);
@@ -348,6 +350,17 @@ export class Converter extends AbstractComponent<Application, ConverterEvents> {
         clearCommentCache();
 
         return project;
+    }
+
+    /**
+     * Resolve `source.url` for every {@link SourceReference} that was
+     * collected by {@link SourcePlugin} during conversion. The work is
+     * async because the underlying `git` lookups are async; the converter's
+     * event bus is synchronous, so this step has to run explicitly after
+     * {@link convert} returns. Safe to call multiple times.
+     */
+    async resolveDeferredSourceUrls(): Promise<void> {
+        await this.sourcePlugin.resolveDeferredUrls();
     }
 
     /** @internal */
--- a/src/lib/converter/plugins/SourcePlugin.ts
+++ b/src/lib/converter/plugins/SourcePlugin.ts
@@ -7,7 +7,8 @@ import { getCommonDirectory, normalizePath, Option } from "../../utils/index.js"
 import { isNamedNode } from "../utils/nodes.js";
 import { relative } from "path";
 import { SourceReference } from "../../models/index.js";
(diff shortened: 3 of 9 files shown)


Gerrit0:
> Open question for a maintainer

I know I've used it in snapshot tests for a plugin, and I could see other plugins wanting to use this during their resolve step, so while it isn't explicitly part of the public API, I'd like to have it set early if feasible.

With this in mind, I did a bit of experimentation with using this as a basis to move it earlier, resulting in #3114. On my box, getting repository information generally takes ~20ms, so neither of these PRs really make a difference, could you take a look and confirm that the performance gains from this are still present?

Gerrit0:
GitHub apparently got confused by a rebase, this was merged with #3114, thank you!

StoneCypher:
i'm sorry.  i was going to do these today.  it's a holiday weekend here.

i did not mean to leave these in your lap.

PR #3112, opened 2026-05-29
Add styles for inline & block <math> elements

kcastellino:
This commit adds styles for `<math>` elements. Inline element styles match inline `<code>` elements, except for font, while block element styles match `<pre><code>` elements, except for font, line wrap, and word break styles.

### DIFF
--- a/example/src/documents/markdown.md
+++ b/example/src/documents/markdown.md
@@ -88,3 +88,18 @@ To use an alert, include a blockquote in any markdown content which starts with 
 
 > [!CAUTION]
 > Advises about risks or negative outcomes of certain actions.
+
+## MathML
+
+Markdown supports inline HTML, including `<math>` elements.
+MathML can be used inline, for example
+<math><msup><mrow><mi>a</mi><mo>&InvisibleTimes;</mo><mi>x</mi></mrow><mn>2</mn></msup> <mo>+</mo> <mi>b</mi><mo>&InvisibleTimes;</mo><mi>x</mi><mo>+</mo> <mi>c</mi><mo> = </mo><mn>0</mn></math>.
+Block elements can also be used:
+
+<math display="block"><mtable columnalign="right left">
+<mtr><mtd><mtext>Area of a circle:</mtext></mtd><mtd><mi>a</mi><mo> = </mo><msup><mrow><mi>π</mi><mo>&InvisibleTimes;</mo><mi>r</mi></mrow><mn>2</mn></mtd></mtr>
+<mtr><mtd><mtext>Area of a triangle:</mtext></mtd><mtd><mi>a</mi><mo> = </mo><mfrac><mn>1</mn><mn>2</mn></mfrac><mo>&InvisibleTimes;</mo><mi>w</mi><mo>&InvisibleTimes;</mo><mi>h</mi></mtd></mtr>
+</mtable></math>
+
+If you'd prefer to write equations using LaTeX expressions, you can load a markdown-it plugin using
+the [`markdownItLoader` configuration option.](https://typedoc.org/documents/Options.Output.html#markdownitloader)
--- a/src/lib/utils/highlighter.tsx
+++ b/src/lib/utils/highlighter.tsx
@@ -103,7 +103,7 @@ class ShikiHighlighter {
         for (i = 0; i < this.schemes.size; i++) {
             style.push(`.hl-${i} { color: var(--hl-${i}); }`);
         }
-        style.push("pre, code { background: var(--code-background); }", "");
+        style.push("pre, code, math { background: var(--code-background); }", "");
 
         return style.join("\n");
     }
--- a/src/lib/utils/highlighter.tsx
+++ b/src/lib/utils/highlighter.tsx
@@ -103,7 +103,7 @@ class ShikiHighlighter {
         for (i = 0; i < this.schemes.size; i++) {
             style.push(`.hl-${i} { color: var(--hl-${i}); }`);
         }
-        style.push("pre, code, math { background: var(--code-background); }", "");
+        style.push("pre, code, math[display='block'] { background: var(--code-background); }", "");
 
         return style.join("\n");
     }
--- a/static/style.css
+++ b/static/style.css
@@ -531,6 +531,21 @@
         scroll-margin-block: calc(var(--dim-header-height) + 0.5rem);
     }
 
+    math {
+        padding: 0.2em;
+        margin: 0;
+        font-size: 0.875rem;
+        border-radius: 0.8em;
+    }
+
+    math[display="block"] {
+        position: relative;
+        padding: 10px;
+        border: 1px solid var(--color-accent);
+        margin-bottom: 8px;
+        font-size: 100%;
+    }
+
     code,
     pre {
         font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
--- a/static/style.css
+++ b/static/style.css
@@ -531,19 +531,12 @@
         scroll-margin-block: calc(var(--dim-header-height) + 0.5rem);
     }
 
-    math {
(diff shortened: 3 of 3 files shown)


Gerrit0 (changes requested):
Please add some math to the `example` folder's "Markdown Showcase" page so that this CSS is exercised somewhere which can be visually inspected easily in the future.

kcastellino:
I've added a small example, although it's all inline MathML.

- Although it's not in the example, fractions in an inline `<math>` element will break out of the top & bottom of the background box. Should I adjust/remove the background to try and compensate?
- I've added a note down the bottom about using markdown-it plugins to render LaTeX syntax. Do you think I should mention any particular plugins? If I was picking one, I'd probably go with [@vscode/markdown-it-katex](https://www.npmjs.com/package/@vscode/markdown-it-katex).

Gerrit0:
> Should I adjust/remove the background to try and compensate?

I think removing the background is likely an appropriate step, looking at how both VSCode and Sphinx does this, they don't have any special background on math blocks.

I wonder if the appropriate styling for a math plugin like this ought to be done by a typedoc plugin which sets the appropriate options in the markdownItLoader function automatically for the user... having some reasonable default makes sense for TypeDoc, but I wouldn't be surprised if someone using mathml heavily would also want styles that it doesn't make sense for this project to maintain since math syntax isn't built in (hmm, should it be? I don't like having extra dependencies, though this would be fairly small)

kcastellino:
The latest commit removes the styles on inline `<math>` elements while keeping them on `<math display="block">` elements (so basically mirroring the styles on `<pre>`). This should be a fairly neutral set of default styles.

For syntax transformation, it's pretty straightforward to do as a plugin to start with - I could probably work on putting one together.

As for the broken builds, not quite sure what is going on there.

Dayday10:
Why Core TypeDoc Should Stay LeanAvoids Dependency Bloat: Adding math libraries increases the package size for everyone.Reduces Maintenance: Math markdown plugins require separate updates and bug fixes.Keeps Focus: TypeDoc works best when it focuses strictly on API documentation.The Best Approach: A Dedicated TypeDoc PluginA separate TypeDoc plugin is the ideal way to handle this. The plugin can automatically hook into the markdownItLoader function. It can inject the required markdown-it math plugins and apply default CSS styles out of the box.This gives users a "one-click" setup while allowing power users to override the styles for heavy MathML use.

Dayday10:
You can configure MathJax in TypeDoc by using a custom markdown engine setup.To do this, you will pass a custom markdownItLoader function into your TypeDoc configuration. This allows you to load the markdown-it-mathjax3 plugin and automatically bundle the required MathJax styles.PrerequisitesFirst, install the necessary markdown-it math plugin in your project:bashnpm install markdown-it-mathjax3
Use code with caution.Configuration Code ExampleAdd the following code to your TypeDoc configuration file (such as typedoc.config.mjs or typedoc.js). This script configures the markdown parser and injects the necessary MathJax CSS styling directly into the generated documentation.javascriptimport { Application } from "typedoc";
import mathjaxPlugin from "markdown-it-mathjax3";

/** @param {Application} app */
export function load(app) {
  // 1. Hook into markdown-it to parse math syntax
  app.options.setValue("markdownItLoader", (customMarkdownIt) => {
    return customMarkdownIt.use(mathjaxPlugin);
  });

  // 2. Inject default MathJax CSS styles so formulas look correct
  app.renderer.addComponent(
    "mathjax-styles",
    new class {
      onRenderEnd(page) {
        // Simple default CSS styling for MathJax containers
        const mathStyles = `
          <style>
            .mjx-container {
              display: inline-block;
              margin: 0.5em 0;
              overflow-x: auto;
              overflow-y: hidden;
              vertical-align: middle;
            }
            mjx-container[display="true"] {
              display: block;
              text-align: center;
              margin: 1em 0;
            }
          </style>
        `;
        
        // Insert the styles into the HTML <head>
        if (page.contents) {
          page.contents = page.contents.replace("</head>", `${mathStyles}</head>`);
        }
      }
    }()
  );
}

Dayday10:
AI Mode
All
Images
Videos
News
Maps
Shopping
Books
Flights
Finance
It makes a lot of sense to want a seamless setup, but keeping math syntax out of the core TypeDoc project is usually the best approach.
Why Core TypeDoc Should Stay Lean
Avoids Dependency Bloat: Adding math libraries increases the package size for everyone.
Reduces Maintenance: Math markdown plugins require separate updates and bug fixes.
Keeps Focus: TypeDoc works best when it focuses strictly on API documentation.
The Best Approach: A Dedicated TypeDoc Plugin
A separate TypeDoc plugin is the ideal way to handle this. The plugin can automatically hook into the markdownItLoader function. It can inject the required markdown-it math plugins and apply default CSS styles out of the box. 
This gives users a "one-click" setup while allowing power users to override the styles for heavy MathML use.
To explore this further, tell me:
Are you building a plugin or configuring your own docs?
Which math engine do you prefer (KaTeX or MathJax)?
Do you want to see a code example of a markdown loader hook?
Knowing your goals will help me provide the most useful next steps.





You can configure MathJax in TypeDoc by using a custom markdown engine setup.
To do this, you will pass a custom markdownItLoader function into your TypeDoc configuration. This allows you to load the markdown-it-mathjax3 plugin and automatically bundle the required MathJax styles.
Prerequisites
First, install the necessary markdown-it math plugin in your project: 
bash
npm install markdown-it-mathjax3
Use code with caution.
Configuration Code Example
Add the following code to your TypeDoc configuration file (such as typedoc.config.mjs or typedoc.js). This script configures the markdown parser and injects the necessary MathJax CSS styling directly into the generated documentation.
javascript
import { Application } from "typedoc";
import mathjaxPlugin from "markdown-it-mathjax3";

/** @param {Application} app */
export function load(app) {
  // 1. Hook into markdown-it to parse math syntax
  app.options.setValue("markdownItLoader", (customMarkdownIt) => {
    return customMarkdownIt.use(mathjaxPlugin);
  });

  // 2. Inject default MathJax CSS styles so formulas look correct
  app.renderer.addComponent(
    "mathjax-styles",
    new class {
      onRenderEnd(page) {
        // Simple default CSS styling for MathJax containers
        const mathStyles = `
          <style>
            .mjx-container {
              display: inline-block;
              margin: 0.5em 0;
              overflow-x: auto;
              overflow-y: hidden;
              vertical-align: middle;
            }
            mjx-container[display="true"] {
              display: block;
              text-align: center;
              margin: 1em 0;
            }
          </style>
        `;
        
        // Insert the styles into the HTML <head>
        if (page.contents) {
          page.contents = page.contents.replace("</head>", `${mathStyles}</head>`);
        }
      }
    }()
  );
}
Use code with caution.
How to Use Math Syntax
Once configured, you can write math in your TypeScript comment blocks using standard MathJax delimiters:
Use single dollar signs $...$ for inline math.
Use double dollar signs $$...$$ for block math. 
typescript
/**
 * Calculates the area of a circle.
 * 
 * The mathematical formula is $A = \pi r^2$.
 *
 * $$A = \pi r^2$$
 */
export function calculateArea(radius: number): number {
  return Math.PI * Math.pow(radius, 2);
}

kcastellino:
Mate, what are you on about? Anyway, just to clarify, there isn't a documented API to add plugins to markdown-it other than passing a function to the `markdownItLoader` option, right?

Gerrit0:
This looks reasonable to me, thank you! The broken CI is from a newer pnpm version that slightly changed the rules for postinstall scripts I think, I need to get my fix merged.

The spam is... well... spam. That account somewhat randomly responds to issues and PRs here with stuff that's obviously copied from some AI, and other times is just incoherent.

The markdownItLoader option is the way to hook into markdown-it for typedoc, if a plugin wanted it, I'd be open to adding another hook (probably some event) which could be used instead

Issue #3113, opened 2026-05-29
Link tags in line comments are not resolved via TypeScript


marijnh:
## Search terms

@link line comment useTsLinkResolution

## Expected Behavior

If I use line comments for my doc comments, I'd expect TypeScript-driven resolution of `@link` targets to still work.

## Actual Behavior

Only the old, built-in resolution is used in this case.

In fact, the machinery needed to do this seems to be entirely absent from `lineLexer.ts`. Is this intentional, or an oversight? (If you want, I can look into adding it. In that case, I'd like to know why these lexers exist as completely separate code, rather than being unified, since they seem very similar.)

## Steps to reproduce the bug

Set `commentStyle` to `"line"`, and use a `@link` that requires TypeScript to resolve (I ran across this trying to reference an import from a sibling package included in the build). TypeDoc will never call `getSymbolAtLocation` for it, and thus fail to resolve it.

## Environment

 - TypeDoc version: 0.28.19
 - TypeScript version: 6.0.3
 - Node.js version: 22.20.0
 - OS: Debian Linux

Gerrit0:
This is unfortunately intentional. It is like this because typescript does not parse line comments for tags, so we can't use TS link resolution here.

marijnh:
Oh right, I see, you're getting the node you pass to `getSymbolAtLocation` from the TypeScript parse.

If I were to find some kind of horrible kludge to beat this information out of the TS API anyway (I haven't yet, but I'd try), would you be interested, or do you feel that it wouldn't be worth the trouble?

marijnh:
It seems possible to do this, if a little awkwardly, though `resolveName`. I have a sketch of a solution at https://github.com/TypeStrong/typedoc/pull/3115, would appreciate feedback.

Gerrit0:
That's far less code than I expected it to be! Unfortunately, we now have 3 different resolution algorithms, which is far from ideal.

Before committing to this route I'd definitely want to try to remove the `ts.JSDoc` path and convert it to use `resolveName`. If that can be made to produce the same results (or slightly different but acceptable results due to the meaning handling) then I think I'd be happy to replace the `ts.JSDoc` path with `resolveName`

marijnh:
> want to try to remove the ts.JSDoc path

That should be very easy (and maybe even make the change a net negative in lines of code).

[...]

PR #3115, opened 2026-05-30
WIP Resolve local links via TypeChecker

marijnh:
This shows a potential way to improve resolution of `@link` tags via the TypeScript typechecker (see https://github.com/TypeStrong/typedoc/issues/3113). It's intended as a sketch to get feedback on at this point, I don't expect it merged as-is.

The most obvious issue with this approach is that it introduces another resolution algorithm, because the existing one in this tree works with TypeDoc reflection objects, which aren't available at the time where we have the node and typechecker around to do this, and the one the TypeScript compiler isn't available to us in a usable way. If that's a non-starter, let me know, and I'll try to just do this in a local plugin instead.

Another issue is whether comments/index.ts is the right place for this code. I can put it into a separate file if you prefer.

Please let me know what you think and whether you see any problems with the code.

### DIFF
--- a/src/lib/converter/comments/index.ts
+++ b/src/lib/converter/comments/index.ts
@@ -12,15 +12,7 @@ import {
 import { lexLineComments } from "./lineLexer.js";
 import { parseComment } from "./parser.js";
 import type { FileRegistry } from "../../models/FileRegistry.js";
-import {
-    assertNever,
-    i18n,
-    Logger,
-    type MeaningKeyword,
-    parseDeclarationReference,
-    setUnion,
-    type SymbolReference,
-} from "#utils";
+import { assertNever, i18n, Logger, parseDeclarationReference, setUnion } from "#utils";
 import { resolveAliasedSymbol } from "../utils/symbols.js";
 import type { Context } from "../context.js";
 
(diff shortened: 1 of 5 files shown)


marijnh:
[Edit: Never mind, figured it out—`StableKeyMap` does the magic. I was just looking at the type, not the initializer.]

I'm trying to debug an issue with this where some (but only some) of the `ReflectionSymbolId` instances I'm creating in this code later, in `resolveLinkTag`, fail to resolve. Or rather, I'm unsure why others _do_ resolve, since as far as I can see those are just objects created in `createSymbolIdImpl`, at which point the only reference to them exists in the `@link` part, and I don't see how they are supposed to land in `ProjectReflection.symbolToReflectionIdMap`. There _is_ a matching symbol ID in that map when the lookup fails, but it's a different object.

So how are the symbol ID objects created in the comment lexer (the JSDoc `@link` lexer does it the same way as my added code) supposed to end up registered in `symbolToReflectionIdMap`?

marijnh:
Okay, so the actual problem is that the `transientId` of the reflection symbol IDs doesn't match. The one my resolver creates has `NaN`, whereas the actual one produced and registered has a number there. I have not been able to figure out why.

marijnh:
Rebased the patches on the current code, and added https://github.com/TypeStrong/typedoc/pull/3115/changes/5fa966cf5a2479d1d40bab07d836c4da8208fdf5, which shows what only resolving the path's root might look like. I think that is a promising direction—it avoids the resolution algorithm duplication, and could probably also replace the JSDoc-based resolution (though I don't know precisely enough what that provides to be certain). It does add another property to `InlineTagDisplayPart` (and I'm not entirely sure if there's any serialization/deserialization stuff I missed), but otherwise it's super simple and seems to work great.

marijnh:
@Gerrit0 you may not be receiving notification for this, but I'd like you to take a look. Sorry for the ping.

src/lib/converter/comments/index.ts
```diff
@@ -101,6 +105,34 @@ function getCommentIgnoringCacheNoDiscoveryId(
     return comment;
 }
 
+function resolveLocalLinks(comment: Comment, context: CommentContextOptionalChecker) {
+    const { checker, node } = context;
+    if (!checker || !node) return;
+    for (const elt of comment.summary) {
```
Gerrit0, line 111: We also need to loop over the parts in `comment.blockTags` or we'll miss `@link` tags in `@param` comments

src/lib/converter/comments/index.ts
```diff
@@ -101,6 +105,34 @@ function getCommentIgnoringCacheNoDiscoveryId(
...
+                pos++;
+            }
+            const parsed = parseDeclarationReference(elt.text, pos, elt.text.length);
+            if (parsed && parsed[0].resolutionStart === "local") {
+                const ref = parsed[0].symbolReference;
+                if (ref && ref.path) {
+                    const symbol = checker.resolveName(
+                        ref.path[0].path,
+                        node,
+                        ts.SymbolFlags.Value | ts.SymbolFlags.Type | ts.SymbolFlags.Namespace,
```
Gerrit0, line 124: `ts.SymbolFlags.All`?

src/lib/converter/comments/index.ts
```diff
@@ -101,6 +105,34 @@ function getCommentIgnoringCacheNoDiscoveryId(
...
+                        true,
```
Gerrit0, line 125: Is there a reason to exclude globals here? This introduces a delta between this behavior and the behavior for `ts.JSDoc` in the case I do something like `{@link Map.size}` (assuming `Map` is in the documentation of course)

src/lib/converter/comments/index.ts
Gerrit0, line 222: I think we need to add `node: sf` to context here, this link resolution doesn't work on `//` comments for the module today

Gerrit0 (commented):
This is very clever! There are a few issues still, but I agree this seems like a promising path forwards.

I did a bit of experimenting with replacing the `ts.JSDoc` implementation, and it almost entirely just works! The remaining delta I think is because TS has weird rules about link resolution (a comment on a namespace which links to `Foo` checks for `Foo` in the namespace first, but `resolveName` with the node we pass in today doesn't do that) https://github.com/TypeStrong/typedoc/tree/resolve-via-typechecker-no-jsdoc has what I got to, though I'm out of time for this week, will have to take another look next weekend.

src/lib/converter/comments/index.ts
marijnh: That somehow returned different results, and is what sent me down the rabbit hole of believing there were multiple symbols for a name in the first place. Maybe I did something stupid, but just swapping to that broke my tests.

src/lib/converter/comments/index.ts
marijnh: My thinking was that this is the place where we resolve _local_ links, and the regular resolution can take care of globals. But probably that's not right, since links like `{@link RegExp}` would count as local links in the sense of the resolver, because they don't contain a `!`, yet refer to a global binding.

marijnh:
Hi Gerrit, thanks for looking into this! Unfortunately, because I ended up moving on to a different approach for my own use case (directly querying the TS data structures, rather than TypeDoc reflections), I won't be continuing work on this patch. If you feel this approach would a be an improvement, you're of course welcome to fix it up to your standards. If not, also feel free to drop it.

Gerrit0:
Thanks for the heads up!

PR #3118, opened 2026-06-14
Resolve via typechecker no jsdoc

Gerrit0:
Extension of #3115 experimenting with removing JSDoc resolution entirely


Gerrit0:
I got this into a working spot, and then remembered https://github.com/Gerrit0/typedoc-plugin-missing-exports/issues/33...

Bah. If I end up merging this, this will either end up breaking the includeDocCommentReferences option of typedoc-plugin-missing-exports, or I will have to duplicate some code to have the symbol traversal happen early so that when creating symbol IDs it targets the actual symbol that's being linked to.

PR #3119, opened 2026-06-20
chore: use HTTPS repository metadata

9904099:
## Summary
- replace the npm repository metadata URL with the public HTTPS GitHub URL
- keep the existing repository target unchanged

## Verification
- `node -e "const p=require('./package.json'); if (p.repository.url !== 'https://github.com/TypeStrong/TypeDoc.git') throw new Error(p.repository.url); console.log(p.repository.url)"`
- `git diff --check`
- `npx -y pnpm@10.25.0 install --frozen-lockfile --ignore-scripts`
- `npm pack --dry-run --ignore-scripts --json .`

I also tried `npx -y pnpm@10.25.0 run build:tsc`, but it fails before this metadata change is involved because `@typestrong/fs-fixture-builder` is unavailable after the install was run with scripts disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "license": "Apache-2.0",
     "repository": {
         "type": "git",
-        "url": "git://github.com/TypeStrong/TypeDoc.git"
+        "url": "https://github.com/TypeStrong/TypeDoc.git"
     },
     "bugs": {
         "url": "https://github.com/TypeStrong/TypeDoc/issues"


9904099:
The current CI failure happens during `pnpm install`, before build/test/lint run and before this metadata-only change is involved.

The failing step is pnpm preparing the existing git-hosted `@typestrong/fs-fixture-builder` dependency:

```text
ERR_PNPM_GIT_DEP_PREPARE_NOT_ALLOWED
@typestrong/fs-fixture-builder ... needs to execute build scripts but is not in the onlyBuiltDependencies allowlist
```

I also reproduced the related local failure path after removing `node_modules`: the dependency prepare path reaches its git package setup rather than this repository metadata change. I have not bundled a dependency/CI configuration change into this PR because the intended change here is limited to the published repository URL.

Gerrit0:
I've now received half a dozen of these across repos I own - my question is still - why? What's the point of doing this change? Everything seems to work as expected with the existing version.

9904099:
Thanks for calling that out. I should have explained the motivation better in the PR description.

I don't think the current value is breaking TypeDoc for normal users. This is not meant as a functional bug fix, and I agree that `npm install typedoc` and the usual GitHub flows work fine today.

The small issue I was trying to clean up is with the repository URL published in npm package metadata. That field is consumed by a lot of tooling which is not always running in a normal developer environment: package mirrors, dependency dashboards, license/SBOM generators, security scanners, and corporate CI jobs which inspect package metadata and sometimes try to resolve or clone the repository URL. In those environments, `git://` is often blocked because it is unauthenticated/plain Git protocol, while HTTPS is usually the only allowed outbound path. When the metadata already uses HTTPS, those tools do not need to special-case or rewrite the repository URL.

So the practical benefit is not that TypeDoc itself starts working differently. It is that downstream tooling which reads TypeDoc's npm metadata gets a more broadly usable repository URL. This also matches the existing `bugs.url`, which already points at GitHub over HTTPS.

That said, I completely understand if this is too minor to be worth maintainer attention, especially if you have received several similar PRs. If you would rather avoid metadata-only churn like this, I am happy to close this PR, and I won't open more of these against repos you maintain.

Issue #2684, opened 2024-08-21
The sorting strategy "alphabetical" should use localeCompare()


kozi:
The sorting strategy "alphabetical" should use localeCompare()

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

https://github.com/TypeStrong/typedoc/blob/ea0ea91d144a34295569113d3208beb1ee2efa03/src/lib/utils/sort.ts#L65
https://github.com/TypeStrong/typedoc/blob/ea0ea91d144a34295569113d3208beb1ee2efa03/src/lib/utils/sort.ts#L74

```javascript
a.name.localeCompare(b.name)
```

Gerrit0:
PR welcome

PR #3120, opened 2026-06-29
fix: sort group and category titles with localeCompare

mahirhir:
`sortGroupCallback` and `sortCatCallback` order groups/categories by a fixed weight first, then fall back to comparing the titles when two entries share the same weight. That fallback used `a.title > b.title ? 1 : -1`, which compares by UTF-16 code point. Code point order places every uppercase letter before every lowercase one and ignores locale collation, so a group named `components` sorted after `Helpers` and `Utilities`, and accented titles sorted after every unaccented one.

The `alphabetical` reflection sort already moved to `localeCompare` in #2684, and `sortGroupCallback` carries a note that it should be kept in sync with the category sort. This change makes both title fallbacks use `localeCompare` so custom `@group` / `@category` titles order the same way reflection names do.

Only equal-weight titles are affected, which in practice means user-defined `@group` / `@category` names that are not listed in `groupOrder` / `categoryOrder`; built-in kind groups keep their existing weighted order. I added a behavior test covering both sorts. Its assertion relies on case differences (`components` before `Helpers` before `Utilities`), which holds regardless of the host locale, and the existing converter specs are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ title: Changelog
 - Introduced `generateOutputsBegin` and `generateOutputsEnd` events on `Application` for plugin use.
 - Group/category section headings (`<h2>`) in the default theme now include an `id` attribute so they can be linked to via fragment identifiers (e.g. `modules.html#runtime-guards`), #3029.
 
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


Gerrit0:
Thank you!

Issue #3098, opened 2026-04-24
TypeScript 7 (TS-Go) support


Gerrit0:
This is the tracking issue for support for TypeScript's port to Go.

Updating to TypeScript 7 is unfortunately not an easy change for TypeDoc as:

1. The TypeScript 7 API is a complete rewrite from the TypeScript 6 and prior APIs.
2. TypeDoc heavily depends on some internal features in TypeScript 6 which are not present in the TypeScript 7 API.

The plan today is to perform a feature freeze once TypeScript 7.0 has released. At that time, additional feature requests will be put on hold until after the necessary changes to support TypeScript 7 have been made.

**There is no timeline for how long this will take.** I work on TypeDoc during my free time during the weekends, usually an hour or two per week. If your company has a pressing need for TypeDoc to support TypeScript 7 and would like to sponsor faster development, please reach out.

samchon:
@Gerrit0 May I challenge this issue with AI coding?

I've fully migrated [`typia`](https://github.com/samchon/typia) to `tsgo`, and also built [`ttsc`](https://github.com/samchon/ttsc) with `Codex` and a handful of prompt snippets.

Honestly, without tools like `Claude Code` or `Codex`, I think I would've given up on the `typia` migration partway through - the surface area was just too large to chew through manually on weekends.

For people like you and me who can only spare a few hours a week, I think AI coding is realistically the only way to keep up with breaking changes of this scale.

## Migration Plan

Given the size of the rewrite, I don't think this can be incremental. It would have to land as a single migration PR. To make that PR trustworthy, I'd expand the test suite significantly beforehand, so the migration is validated against a much tighter safety net than what exists today.

I'd also battle-test the migrated version against my own projects first - `typia`, `nestia`, and a few others - which together exercise a fairly wide range of TypeScript's type system. Only after it survives those would I open the PR.

## Long-Term: Playground Mode

One more thing, if you're open to it: I'd like to eventually contribute a playground mode to `TypeDoc`, as a separate PR sometime after the `tsgo` migration settles. No timeline, just flagging the direction.

[`typia.io/playground`](https://typia.io/playground) currently bundles `tsgo` + `ttsc` + `typia` into a single WebAssembly build, so users can try the whole toolchain in-browser. Something similar for `TypeDoc` would, I think, help library authors a lot - previewing generated docs live while writing `JSDoc`, without running a full build, would shorten the iteration loop quite a bit.

[...]

PR #3124, opened 2026-07-17
Fix typedoc-ts export path for typedoc/browser

### LINKED ISSUE #3123: Export `node-utils` from package.json
Before you lock down TypeDoc for the TSGO7 upgrade...

I have a small request for a custom deployment of TypeDoc where I have some advanced use cases that benefit from access to the runtime node utilities. I need to access [discoverAllReferenceTypes](<https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/reflections.ts>). For programmatic deployments it may be useful to simply expose `node-utils` similar to how the `browser-utils` has an export of `typedoc/browser`. This can be accomplished simply by adding the proper sub-path export condition to `package.json`

```json
{
   "exports": {
      "./node": {
         "typedoc-ts": "./src/lib/node-utils.ts",
         "types": "./dist/types/node-utils.d.ts",
         "default": "./dist/node-utils.js"
      }
   }
}
```

This allows:
```js
import { discoverAllReferenceTypes } from 'typedoc/node';
```

At the end of the day... I can simply modify the `package.json` after TypeDoc installation into my specific project to add the above, but I think this is generally useful for more advanced programmatic use cases in general. 

----

Also just a quick thing I noticed when looking at this enhancement request.. It appears the browser export condition for `typedoc-ts` is not pointing to the source. IE `"typedoc-ts": "./src/lib/browser-utils/index.ts"` likely should be `"typedoc-ts": "./src/lib/browser-utils.ts",` as there is no `./src/lib/browser-utils` folder in the latest repo update that I see.

lazerg:
The `typedoc-ts` condition on the `./browser` export points at `./src/lib/browser-utils/index.ts`, which doesn't exist. The actual file is `./src/lib/browser-utils.ts`, so resolving `typedoc/browser` under that condition currently fails. This corrects the path.

Fixes #3123

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
             "default": "./dist/models.js"
         },
         "./browser": {
-            "typedoc-ts": "./src/lib/browser-utils/index.ts",
+            "typedoc-ts": "./src/lib/browser-utils.ts",
             "types": "./dist/types/browser-utils.d.ts",
             "default": "./dist/browser-utils.js"
         },


lazerg:
Hi @Gerrit0, this PR has been open for a while with no review yet. Could you take a look when you have time? Thanks!

Gerrit0:
> At this point, I'm spending all the time I have for TypeDoc working on this, so unless there's a critical bug, I'm probably going to defer looking at PRs for a bit! - #3098

I'll get to it eventually

